PR #229, opened 2020-02-06
Ddpb 3181: CSV upload from Sirius

aiesoftware:
## Purpose
Support Lay CSV upload from multiple sources

```
As a deputy
I want lay CSVs to be imported from CasRec and Sirius
So that I can register for the digital reporting service
```
Fixes [DDPB-3181](https://opgtransform.atlassian.net/browse/DDPB-3181)

## Approach
There are two things we need to take care of: deleting the relevant items in the `casrec` table according to the source of the CSV, and uploading new entries into the `casrec` table with a new `source` field, also according to the source of the CSV.

Identifying the source is straightforward: if there isn't one in the CSV, or there is an invalid one (i.e not "casrec" or "sirius"), assume "casrec", otherwise use the source value from the CSV. 

Use the determined source to hit a new endpoint `casrec/delete-by-source/{source}` which will remove all casrec entries by the given source. Now we are ready to upload new entries into casrec.

Uploading casrec rows from a new Sirius source is a case of creating a new Assembler that has knowledge of the Sirius CSV headers, and assembles a `LayDeputyshipDto` from the data. The DTO is a normalised domain object that the `LayDeputyshipUploader` takes in. So long as the Assembler creates a valid DTO and passes it to the uploader, then the upload will work as usual.

## Learning
As it stands, each upload begins with an API call to a /truncate endpoint. At first, instead of creating a /delete-by-source endpoint, I did it directly from within the API inside the upload process. This would reduce the extra API call across the network, and also allow us to add the deletion into the database transaction that the upload uses, so any failure would not execute the delete. However, the separate API call is required as it’s a one time event for the entire upload, whereas the upload code within the API runs once for every batch that is sent, effectively meaning we would be deleting by source many times per upload of the entire file (once for each batch). At least this clarifies why we have the separate call out.

## Checklist
- [ ] I have updated documentation (Confluence/GitHub wiki/tech debt doc) where relevant
- [ ] I have added tests to prove my work, and they follow our [best practices](https://github.com/ministryofjustice/opg-digi-deps-client/wiki/Testing-best-practices)
- [ ] The product team have tested these changes

### Frontend

### DIFF
--- a/api/src/AppBundle/Entity/CasRec.php
+++ b/api/src/AppBundle/Entity/CasRec.php
@@ -200,7 +200,9 @@ class CasRec
         $this->deputyPostCode = self::normaliseSurname($row['Dep Postcode']);
         $this->typeOfReport = self::normaliseCorrefAndTypeOfRep($row['Typeofrep']);
         $this->corref = self::normaliseCorrefAndTypeOfRep($row['Corref']);
-        $this->source = isset($row['source']) ? strtolower($row['source']) : self::CASREC_SOURCE;
+
+        $source = isset($row['Source']) ? $row['Source'] : self::CASREC_SOURCE;
+        $this->setSource($source);
 
         $this->otherColumns = serialize($row);
         $this->createdAt = new \DateTime();
@@ -392,6 +394,11 @@ class CasRec
      */
     public function setSource($source)
     {
+        $source = strtolower($source);
+        if (!in_array($source, self::validSources())) {
+            throw new \InvalidArgumentException(sprintf('Attempting to set invalid source: %s given', $source));
+        }
+
         $this->source = $source;
         return $this;
     }
--- a/api/src/AppBundle/Entity/CasRec.php
+++ b/api/src/AppBundle/Entity/CasRec.php
@@ -9,7 +9,7 @@ use Symfony\Component\Validator\Constraints as Assert;
 
 /**
  * @ORM\Table(name="casrec", indexes={@ORM\Index(name="updated_at_idx", columns={"updated_at"})})
- * @ORM\Entity()
+ * @ORM\Entity(repositoryClass="AppBundle\Entity\Repository\CasRecRepository")
  */
 class CasRec
 {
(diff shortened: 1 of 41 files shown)


api/src/AppBundle/Entity/CasRec.php
```diff
@@ -366,4 +377,33 @@ public function getUpdatedAt()
...
+        return $this->source;
+    }
+
+    /**
+     * @param mixed $source
+     * @return CasRec
+     */
+    public function setSource($source)
+    {
+        $this->source = $source;
```
gregtyler: Might be worth adding a check here of `in_array($source, self::validSources())`?

api/tests/AppBundle/Controller/CasRecControllerTest.php
```diff
@@ -54,37 +54,43 @@ public function setUp(): void
...
-            'Surname' => 'I should get deleted',
-            'Deputy No' => 'Deputy No',
-            'Dep Surname' => 'Dep Surname',
-            'Dep Postcode' => 'SW1',
-            'Typeofrep'=>'OPG102',
-            'Corref'=>'L2'
+        $url = '/casrec/delete-by-source/unknownsource';
+
+        $this->assertJsonRequest('DELETE', $url, [
+            'mustSucceed' => false,
```
gregtyler: I think this should be:
```suggestion
            'mustFail' => true,
```
If `mustSucceed` is false, it won't check anything

behat/tests/features/registration/self-registration.feature
```diff
@@ -0,0 +1,27 @@
...
+      | Parker        | peter@parker.com | SW10 1AA       | Davies        | T5001034   |
+    And I am logged in as "peter@parker.com" with password "Abcd1234"
+    Then I should be on "/lay"
+    And I should see "Davies"
+
+  Scenario: Register a deputy from a sirius source
+    Given an admin user uploads the "behat-lay-sirius.csv" file into the Lay CSV uploader
+    When these deputies register to deputise the following court orders:
+      | deputySurname | deputyEmail    | deputyPostCode | clientSurname | caseNumber |
+      | Wayne        | bruce@wayne.com | SW1 2SA        | Hendry        | T5001041   |
```
gregtyler: Davies... Hendry... Will we see Parrott next?

aiesoftware: Well spotted (:

api/src/AppBundle/Entity/CasRec.php
aiesoftware: Added check, and ensured the constructor sets it via this method, instead of directly on to the `$this->source` property.

acsauk-mojd:
Nicely done - I found this much easier to follow from the run through you did 👍

behat/tests/features/registration/self-registration.feature
aiesoftware: Although I noticed the error in my ways with Davies instead of Davis, but thought it really wasn't worth correcting 😂

PR #231, opened 2020-02-10
DDPB-3031 - Add account level cloud9 instance

acsauk-mojd:
## Purpose
Adding a single shared account level cloud9 means we can reduce the number of resources on our account (and associated costs) while locking down the security for RDS <--> cloud9.

Fixes DDPB-3031

## Approach
This was a bit of a pain to get set up as we need to apply the security group automatically generated when creating a cloud9 environment to each environments RDS instance. We didn't have a mechanism to get the id of the account level cloud9 instance from `shared` to `environment` and terraform doesn't provide a data source for cloud9 instances. Instead, I've output the id of the cloud9 instance created during the `shared` which is then available in the tfstate file. We then access this state file remotely in S3 using a `terraform_remote_state` data source to access and assign the security group (phew).

One point to note, I had to assign myself as the owner of the cloud9 env in order to make it visible int he console. I'll manually share this with the rest of the team as there is (currently) no way to automate this in code via Terraform.

## Learning
I wasn't even aware accessing tfstate remotely was possible until Andrew filled me in in web ops community. Its something to keep in mind for any other occasion we need to pass data around. 

Also, I came up against a few issues when trying to initially create the cloud9 instance using Terraform by not applying an owner or an owner other than myself. This led to multiples of the same environment being created which were then unable to be destroyed as they had a dependency chain on SGs and network interfaces. In the end, I resorted to destroying my entire PR environment and rebuilding - don't make the same mistake I did and try and go down the rabbit hole of deleting the dependencies int he correct order 🙃 

## Checklist

### DIFF
--- a/shared/cloud9.tf
+++ b/shared/cloud9.tf
@@ -3,13 +3,6 @@ resource "aws_cloud9_environment_ec2" "shared" {
   name                        = "shared-env"
   automatic_stop_time_minutes = 20
   description                 = "Shared Cloud9 instance to be used by all devs"
-  subnet_id                   = aws_subnet.private[0].id
+  subnet_id                   = aws_subnet.public[0].id
+  owner_arn                   = "arn:aws:iam::${var.accounts[terraform.workspace].account_id}:assumed-role/operator/alex.saunders"
 }
-
-
-//name - (Required) The name of the environment.
-//instance_type - (Required) The type of instance to connect to the environment, e.g. t2.micro.
-//automatic_stop_time_minutes - (Optional) The number of minutes until the running instance is shut down after the environment has last been used.
-//description - (Optional) The description of the environment.
-//owner_arn - (Optional) The ARN of the environment owner. This can be ARN of any AWS IAM principal. Defaults to the environment's creator.
-//subnet_id - (Optional) The ID of the subnet in Amazon VPC that AWS Cloud9 will use to communicate with the Amazon EC2 instance.
--- a/shared/cloud9.tf
+++ b/shared/cloud9.tf
@@ -1,6 +1,6 @@
 resource "aws_cloud9_environment_ec2" "shared" {
   instance_type               = "t2.micro"
-  name                        = "shared-env"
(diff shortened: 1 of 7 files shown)


shared/cloud9.tf
```diff
@@ -0,0 +1,8 @@
+resource "aws_cloud9_environment_ec2" "shared" {
+  instance_type               = "t2.micro"
+  name                        = "team-cloud9-env"
+  automatic_stop_time_minutes = 20
+  description                 = "Shared Cloud9 instance to be used by all devs"
+  subnet_id                   = aws_subnet.public[0].id
```
gregtyler:
```suggestion
  subnet_id                   = aws_subnet.private[0].id
```
It looks like our old Cloud9 environments were set up on the private subnet/VPC.

acsauk-mojd: I originally had this as private VPC and private subnet but this meant the environment wasn't able to successfully be created. When I looked into it further, the github instructions all point to private VPC and public subnet
Screenshot 2020-02-11 at 18 00 26

Screenshot 2020-02-11 at 18 02 47

Although, I'd missed the differentiation between development and production values. I'll update the terraform to use the correct values for production but from what I understand, development is OK as it is.

acsauk-mojd: Actually scratch that, it should be OK for production as well:

Production VPCs/Subnets
Screenshot 2020-02-11 at 18 05 50

jamesrwarren: Yes this looks correct then if we want it to emulate the old cloud9 setups which we do. Nice.

gregtyler: I got confused because the private/public subnets are both attached to the "private" VPC. I'd thought that made them private in Terraform terms. Looking in a bit more depth though, we only use the one VPC in Terraform. The one without a name looks like a default AWS one or something 👍

acsauk-mojd: Great stuff, cheers guys 👍

PR #234, opened 2020-02-12
ddpb-3224 add replication group without dns switch

jamesrwarren:
## Purpose
Initial part in 3 stage process to release DDPB-3192. 

Fixes DDPB-3192

## Approach
Create the initial replication groups alongside the original cluster to test a switchover.

Full description in main PR: https://github.com/ministryofjustice/opg-digideps/pull/233

## Learning
See main DDPB-3192 and https://github.com/ministryofjustice/opg-digideps/pull/233

## Checklist

### Frontend

### DIFF
--- a/environment/admin_elasticache.tf
+++ b/environment/admin_elasticache.tf
@@ -1,5 +1,6 @@
 resource "aws_elasticache_replication_group" "admin" {
-  count                         = local.account.is_production == 0 ? 1 : 0
+  // Include this once first two subtasks of DDBP-3192 are done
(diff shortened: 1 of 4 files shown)


jamesrwarren:
> I'm happy with the new infra, but I would like to ensure that the resources used by production and not-production are as close as possible.
> 
> So I'd like to bin the `aws_elasticache_cluster`s and use `aws_elasticache_replication_group` for every environment (with `number_cache_clusters` varying for production, I guess?)

I thought I had tried this but actually you're right, this should be possible.

PR #235, opened 2020-02-12
DDPB 3160 Ping Sirius gateway for health check

aiesoftware:
## Purpose
Adds a Sirius ping to our health check page.

Fixes DDPB-3160

## Approach
We want to report an error without causing the availability page to return a 500. To prevent this, we have to report `isHealthy` as `true`, despite any errors. To support this, I added a new `customMessage` property to the service, which if set, will override the default 'OK' text that is printed when `isHealthy = true`. This way we can render the failure response instead of 'OK'.

**Note**
The current status returned is `403 Forbidden` because the environment cannot currently authenticate with the endpoint. However, this isn't a defect of this ticket - the availability page is working correctly in that it _is_ reporting the current state of affairs.

This does need addressing elsewhere though @jamesrwarren 

## Learning
We've used an AWS `Parameter` for the first time here, to store the API gateway base URI. It's very much the same as a `Secret`, but incurs no extra cost. A notable difference is the absence of secret rotation, but this is not relevant for a URI so is a good candidate for a parameter.

## Checklist

### Frontend


client/src/AppBundle/Service/Availability/SiriusApiAvailability.php
```diff
@@ -0,0 +1,28 @@
...
+use Symfony\Component\DependencyInjection\ContainerInterface;
+
+class SiriusApiAvailability extends ServiceAvailabilityAbstract
+{
+    public function __construct(ContainerInterface $container)
+    {
+        $this->isHealthy = true;
+
+        try {
+            $response = $container->get('AppBundle\Service\Client\Sirius\SiriusApiGatewayClient')->get('v1/healthcheck');
```
gregtyler: You should be able to get this with dependency injection:
```
public function __construct(ContainerInterface $container, SiriusApiGatewayClient $client)
...
$response = $client->get('v1/healthcheck');
```

environment/terraform.tfvars.json
```diff
@@ -16,7 +16,8 @@
       "mock_emails": false,
       "symfony_env": "prod",
       "db_subnet_group": "rds-private-subnets-prod-vpc",
-      "ec_subnet_group": "ec-pvt-subnets-prod-vpc"
+      "ec_subnet_group": "ec-pvt-subnets-prod-vpc",
+      "sirius_api_account": "649098267436"
```
gregtyler: I would've expected a separate account for the production API gateway. Is prod and preprod shared, or has it just not been set up yet @jamesrwarren?

client/src/AppBundle/Service/Availability/SiriusApiAvailability.php
aiesoftware: True. I begrudgingly left it as `$container` as it's in keeping with the other Availability services. They are each instantiated manually from within the `ManageController::availabilityAction`, which just passes `$this->container` into each Availability service's constructor. I _could_ inject the `SiriusApiGatewayClient` service into that controller action and pass it in from there, but the same could then be said for each other service. In short, sticking with what we have seemed less contentious.

gregtyler: That's fair enough! Makes sense to stay consistent 👍

environment/terraform.tfvars.json
jamesrwarren: Hasn't been set up as far as I know. Will ask Tom today as he was wokring on it and we need to push the deploy out to pre and prod.

gregtyler: From Slack: Other environments are still in the process of being set up so we'll review these account IDs once that work's complete.

PR #237, opened 2020-02-12
Prefer github source for slack module to fix path bug

acsauk-mojd:
## Purpose
Using GitHub as source for Slack terraform module to fix the path bug referenced [here](https://github.com/terraform-aws-modules/terraform-aws-notify-slack/issues/21#issuecomment-565092931).


gregtyler:
I threw some changes on here to fix an issue with nightly jobs

PR #240, opened 2020-02-13
DDPB-2930: Use Serverless Aurora for development environments

gregtyler:
## Purpose
We're paying a lot of money for unused databases in our non-production environments. They're used at most eight hours a day and often much less. We therefore want to run them in RDS's Serverless mode so they can automatically pause when not in use.

Outcome of DDPB-2930

## Approach
I've added a new `always_on` setting to each environment in Terraform. If it's set to true, the environment has a provisioned database that's always available. If it's set to false, a serverless one that can automatically pause is used instead.

I've had to turn off the availability healthcheck for not-always-on environments: checking the availability page every 10 seconds forces the database to stay awake, defeating the point of having a Serverless DB. Besides, we only really care about downtime in production.

I've added an ADR to explain the direction of travel here, but I'm starting with just the development environments because there's more of them, and it's less impactful. But over time, I'm proposing we roll out serverless to all environments apart from production and preproduction.

## Learning
I went around the houses a lot with this, trying different types of database and identifying how to safely migrate data when changing DB type (specifically Postgres instance -> Aurora provisioned). All of that was really useful learning, and I think gives a clear plan of how to progress further in making our DBs robust and updated.

I think the next logical step is a manual process to upgrade all our DBs to Postgres 10 (DDPB-3228). This will require downtime, but should just be a button press in the AWS console. We'll then be able to migrate to Aurora databases, either provisioned or serverless, saving costs and giving us more resilience.

## Checklist
  - Created an ADR and DDPB-3228
  - N/A, all the tests rely heavily on the database


environment/terraform.tfvars.json
```diff
@@ -68,7 +71,8 @@
       "symfony_env": "test",
       "db_subnet_group": "private",
       "ec_subnet_group": "private",
-      "state_source": "preproduction"
+      "state_source": "preproduction",
+      "always_on": true
```
aiesoftware: Is not feasible for `master` and `preproduction` to have `"always_on": false`?

gregtyler: Eventually it will be, I've got [another ticket](https://opgtransform.atlassian.net/browse/DDPB-3242) in for that.

We can't do it (easily) in this PR because it requires in-place replacement of the database. For preproduction/training we'll have to do a proper migration to avoid data loss (and have [upgraded to Postgres 10](https://opgtransform.atlassian.net/browse/DDPB-3228)).

Also, Serverless requires Postgres 10 but production currently uses Postgres 9. Keeping master on Postgres 9 for now ensures the tests are run against the correct DB version at some point in the pipeline.

aiesoftware:
Great work!

environment/api_rds.tf
```diff
@@ -40,10 +41,92 @@ resource "aws_db_instance" "api" {
...
+      "Name" = "api.${local.environment}.${local.account.account_id}.${data.aws_route53_zone.public.name}"
+    },
+  )
+
+  lifecycle {
+    ignore_changes = [engine_version, master_password]
+  }
+}
+
+# resource "aws_rds_cluster_instance" "api" {
```
jamesrwarren: Is this commented out bit supposed to be there?

gregtyler: Kind of. We'll need it when we switch our always on databases to provisioned Aurora and I didn't want to lose it.

PR #243, opened 2020-02-17
ddpb-3192 remove old elasticache cluster

jamesrwarren:
## Purpose
Create only 1 additional read node per ECS cluster for Front, Admin and API in production only (due to cost) in a different availability zone.
Fixes DDPB-3192

## Approach
My initial approach was to try and add non clustered read replicas managed by terraform through this resource: aws_elasticache_replication_group and adding the replicas as stated here https://www.terraform.io/docs/providers/aws/r/elasticache_replication_group.html. However after trying every combination and a lot of google searching, I wasn't able to get just two nodes with terraform managing the redis nodes so I had to let the aws_elasticache_replication_group manage it directly. For testing I used cloud9 with it's security group added to the redis sg allow list to use the redis-cli to set some values on redis, then fail over the (checking events that it failed over correctly) then doing a get and checking the values still appeared on the new primary. This worked fine. The other tests are based on the standard behat tests running through correctly in clustered mode.

For the release I had to split this into 3 stages. This is the final stage that will remove the superfluous code and the old elasticache cluster as it has been replaced by a distribution zone.

## Learning
What looks like it ought to work in the documentation that is supposed to give fine grained access for managing individual replicas in terraform doesn't work as advertised. It is possible to add 2 number_cache_clusters and 2 managed replicas for example but the documentation suggests that when they are named the same, it would create the 2 number_cache_clusters as the replicas. This isn't the case and it complains about them having the same name (also impossible to set number_cache_clusters to less than 2) so have had to go for non managed. Have learned that you can test redis manually by allowing the cloud 9 security group access on the redis security group temporarily.

## Checklist

### Frontend

### DIFF
--- a/environment/admin_elasticache.tf
+++ b/environment/admin_elasticache.tf
@@ -2,7 +2,6 @@ resource "aws_elasticache_replication_group" "admin" {
   automatic_failover_enabled    = local.account.elasticache_count == 1 ? false : true
   engine                        = "redis"
   engine_version                = "5.0.0"
-  availability_zones            = null_resource.elasticache_az_list.*.triggers.az
   replication_group_id          = "admin-rep-group-${local.environment}"
   replication_group_description = "Replication Group for Admin"
   node_type                     = "cache.t2.small"
--- a/environment/admin_elasticache.tf
+++ b/environment/admin_elasticache.tf
@@ -1,33 +1,16 @@
 resource "aws_elasticache_replication_group" "admin" {
-  // Include this once first two subtasks of DDBP-3192 are done
-  //  count                         = local.account.is_production == 1 ? 1 : 0
-  automatic_failover_enabled    = true
+  automatic_failover_enabled    = local.account.elasticache_count == 1 ? false : true
   engine                        = "redis"
   engine_version                = "5.0.0"
-  availability_zones            = ["eu-west-1a", "eu-west-1b"]
+  availability_zones            = null_resource.elasticache_az_list.*.triggers.az
   replication_group_id          = "admin-rep-group-${local.environment}"
   replication_group_description = "Replication Group for Admin"
   node_type                     = "cache.t2.small"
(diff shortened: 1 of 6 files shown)


environment/terraform.tfvars.json
```diff
@@ -34,7 +35,8 @@
       "symfony_env": "prod",
       "db_subnet_group": "private",
       "ec_subnet_group": "private",
-      "state_source": "preproduction"
+      "state_source": "preproduction",
+      "elasticache_count": 1
```
gregtyler: It would be nice to also have two nodes in preproduction so we've got a production-like environment. Would that cost a huge amount financially?

environment/admin_elasticache.tf
```diff
@@ -1,33 +1,16 @@
 resource "aws_elasticache_replication_group" "admin" {
-  // Include this once first two subtasks of DDBP-3192 are done
-  //  count                         = local.account.is_production == 1 ? 1 : 0
-  automatic_failover_enabled    = true
+  automatic_failover_enabled    = local.account.elasticache_count == 1 ? false : true
   engine                        = "redis"
   engine_version                = "5.0.0"
-  availability_zones            = ["eu-west-1a", "eu-west-1b"]
+  availability_zones            = null_resource.elasticache_az_list.*.triggers.az
```
gregtyler: I think we can just leave this blank and AWS will identify # suitable AZs in eu-west1 ([search "PreferredCacheClusterAZs" here](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-elasticache-replicationgroup.html)). Doesn't make a difference to what's actually built, but would save us a bit of configuration.

environment/terraform.tfvars.json
jamesrwarren: Ive made it 2. If we decide it's unneeded we can just change this value again.

PR #245, opened 2020-02-20
Behat spike

acsauk-mojd:
## Purpose
Some updates to existing ways of doing things in behat tests, the removal of a bunch of old unused contexts and proposed stubs for future feature tests.

Fixes DDPB-3029.

## Approach
This was more painful than I expected to make this meagre progress 😞 I'd originally set off trying to fix what is the most used /tested UI feature of logging in (200+ times used!) but I've not managed to find a working solution given the structure of our app. 

I've left the code in (`FixtureController::authenticateUserWithPassword()`) to show my attempt at creating a user token and then persisting this in the TokenStorage to simulate a logged-in user but I think the combination of not being able to reliably access the right session and having the app code inaccessible from the behat container meant this approach didn't work.

I did have some success at enabling the creation of Users by expanding the work Alex had put in place for report management. It's not sped the tests up by much (15 seconds tops in my estimations) and this could have been more but the way the function it was supposed to be replacing (`UserTrait::iCreateTheUserWithEmailAndPostcode()`) has been used relied heavily on the email flow, and in some places is still required to ensure we send emails, so I didn't want to do a replace all.

I also took the opportunity to look at the features we're testing through the feature tests and as I identified them created new feature files ready to be populated. I'm not sure how everyone feels about them but we can see how it goes and drop/expand as required.

## Checklist


api/Dockerfile
```diff
@@ -52,7 +52,8 @@ RUN openssl req -newkey rsa:4096 -x509 -nodes -keyout /etc/nginx/certs/app.key -
 
 RUN mkdir -p var/cache \
   && mkdir -p var/logs \
-  && chown -R www-data var
+  && chown -R www-data var \
+  && chmod -R 777 var
```
gregtyler: Is this some hangover code from another change? I don't think it should be necessary

client/src/AppBundle/Controller/Admin/FixtureController.php
```diff
@@ -78,4 +80,57 @@ public function completeReportSectionsAction(Request $request, $reportId, Kernel
...
+     * @Route("/createUser", methods={"GET"})
+     * @Security("has_role('ROLE_ADMIN', 'ROLE_AD')")
+     */
+    public function createUser(Request $request, KernelInterface $kernel)
+    {
+        if ($kernel->getEnvironment() === 'prod') {
+            throw $this->createNotFoundException();
+        }
+
+        var_dump($request->query->all());
```
aiesoftware: Can we remove this?

client/src/AppBundle/Controller/Admin/FixtureController.php
```diff
@@ -78,4 +80,57 @@ public function completeReportSectionsAction(Request $request, $reportId, Kernel
...
+
+        $email = $request->query->get('email');
+        $password = $request->query->get('password');
+
+        $payload = <<<JSON
+{
+    "email": $email,
+    "password": $password
+}
+JSON;
```
aiesoftware: Should we use json_encode to be consistent with the `createUser` method?

aiesoftware:
Good work, looks like a real slog! Good to see a lot of trait methods removed. Just two minor observations.

api/Dockerfile
acsauk-mojd: This came about after trying to set the user token in the var/cache folder. I had to manually change the permissions for it to work but as I'm removing this feature, for now, I'll drop this change also.

client/src/AppBundle/Controller/Admin/FixtureController.php
acsauk-mojd: Knew I'd leave one in - I'll remove.

client/src/AppBundle/Controller/Admin/FixtureController.php
acsauk-mojd: Good point yeah, I'll switch to using that.

PR #246, opened 2020-02-21
DDPB-3229: Generate pact files for DigiDeps

gregtyler:
## Purpose
To test our contract with the API gateway, we should generate pact files to test against and upload them to the Pact broker.

This preliminary ticket is to generate an initial example Pact file.

Fixes DDPB-3229

## Approach
I installed [pact-php](https://github.com/pact-foundation/pact-php/) and the `pact-cli` Docker image. In tests, we use Pact to generate a mock interaction and verify that the interaction works as prescribed when testing the service. After the tests have finished, we generate a Pact JSON file which can be uploaded to the broker.

To ensure we could test against a mock, and to extract the final Pact JSON file, I switched the client unit tests to run directly on CircleCI rather than in AWS.

## Learning
Came across a few limitations with Pact, particularly around multipart requests (it's designed much more for JSON), which led to creating a fairly ugly regex test for the request body.

There are a few things I'm unsure of, or that we could do better, but I think it's worth engaging with those as we develop the software rather than getting ahead of ourselves. This is more a proof-of-concept.

## Checklist
  - These are exemplar tests: I don't want to get into the weeds of implementation details before we've built the API or consumer code


.circleci/config.yml
```diff
@@ -346,6 +344,22 @@ jobs:
...
+          name: Run tests
+          command: |
+            docker-compose up -d pact-mock
+            docker-compose run --no-deps --rm frontend bin/phpunit -c tests/phpunit
+            docker-compose exec pact-mock cat /tmp/pacts/complete_the_deputy_report-opg_data.json > pact.json
+            docker-compose stop pact-mock
+      - store_artifacts:
+          path: pact.json
+          destination: Pact file
+
```
acsauk: 🎉

client/src/AppBundle/Service/DocumentSyncService.php
```diff
@@ -0,0 +1,110 @@
...
+        if ($report instanceof Ndr) {
+            $type = 'NDR';
+        } else if (in_array($report->getType(), [Report::TYPE_HEALTH_WELFARE, Report::TYPE_COMBINED_HIGH_ASSETS, Report::TYPE_COMBINED_LOW_ASSETS])) {
+            $type = 'HW';
+        } else {
+            $type = 'PF';
+        }
+
+        if ($report instanceof Ndr) {
+            $endDate = $report->getStartDate()->format('Y-m-d');
```
acsauk: Might be my ignorance with NDRs but how does the start date work out to be the end date 🤔

client/tests/phpunit/Pact/Listener/PactTestListener.php
```diff
@@ -0,0 +1,89 @@
...
+     */
+    public function startTestSuite(TestSuite $suite): void
+    {
+        $httpService = new MockServerHttpService(new GuzzleClient(), $this->mockServerConfig);
+        $httpService->deleteAllInteractions();
+    }
+
+    public function addError(Test $test, \Throwable $t, float $time): void
+    {
+        $this->failed = true;
```
acsauk: Not sure of the internal working of pact or outputs, but is there any way to provide a bit more context around why the test error-ed? Or will this be part of the failing test?

client/tests/phpunit/Pact/Listener/PactTestListener.php
```diff
@@ -0,0 +1,89 @@
...
+    }
+
+    public function addFailure(Test $test, AssertionFailedError $e, float $time): void
+    {
+        $this->failed = true;
```
acsauk: Same as above

client/tests/phpunit/Pact/Listener/PactTestListener.php
gregtyler: This is just a PHPUnit listener that's called if the test errors or fails. The normal PHPUnit output/exit code will happen as usual, but this allows us to stop generating a broken pact file (in `endTestSuite` below).

If the test fails because of something to do with Pact, details will be output like other tests: either with an error to stderr or with a "Expected value 1234 does not equal 4567"

client/src/AppBundle/Service/DocumentSyncService.php
gregtyler: NDRs are issued outside of annual reporting, so don't have a reporting period. In practice, they're assigned a start date to match the first report's start date. For the sake of providing a required value, I'm setting the end date to the same date.

Court date: 12th May 2017
NDR: 12th May 2017–12th May 2017
Report 1: 12th May 2017–11th May 2018
Report 2: 12th May 2018–11th May 2019

I'll draw this to Stacey's attention though. Maybe we don't want to require the end date in the API (or not sync NDRs).

.circleci/config.yml
```diff
@@ -345,6 +341,37 @@ jobs:
...
+            docker tag ${AWS_REGISTRY}/digideps/client:${VERSION} client:latest
+            docker save -o docker-images.tar client:latest
+      - persist_to_workspace:
+          root: .
+          paths:
+          - ./docker-images.tar
+
+  client-unit-test:
+    docker:
+      - image: circleci/python:2.7
```
acsauk-mojd: Is there a reason we can't use a later version of Python? Or is 2.7 still in support?

PR #249, opened 2020-02-24
Ddpb 3103

aiesoftware:
## Purpose
To allow super admin users to manually discharge a deputy from a client.

We currently have an automated discharge in place for the scenario of a client moving from one professional organisation to another, but it does not capture clients moving from lay to non lay (and vice versa). This feature will allow us to discharge across the board.

Fixes DDPB-3103

## Approach
To "discharge" a named deputy from a client, is actually to delete the client from the CTDR application. The method of discharging up until now, be it through the automation or manually updating the database, has been to soft delete the client by setting a `deleted_at` field. This is leading to a build up of duplicate clients in our system, seeing as a new client with the same case number is recreated. It is also not advised by the business that we hold on to any data that we no longer require. 

Changelog:
- Removed the `deleted_at` field from the `Client` entity/database table
- Updated exisiting auto discharge functionality to hard delete instead of soft delete
- Created a new admin page that allows super admin users to "discharge" (delete) a given client
- Created a new `DELETE` endpoint in the API to support the delete action
- Removed the `Discharged` column from the admin client list page
- Updated the `FixtureController->courtOrder` endpoint to support adding the same deputy to several court order fixtures in Behat without creating a new Organisation each time
- Tidied up newer feature based Behat tests into a `v2` directory to aid with the ongoing migration/audit of our tests

## Checklist

### Frontend


behat/tests/features/prof/03-report/09-deputy-discharged.feature
```diff
@@ -1,214 +0,0 @@
-Feature: Prof deputy is discharged
```
gregtyler: Can this file be deleted rather than emptied?

Very happy to see old tests go 💪

behat/tests/features/prof/05-organisation/04-organisation-updates-acl.feature
```diff
@@ -59,32 +61,13 @@ Feature: Organisation deputyship updates
       | GB                                    | deputy-details |
       | behat-prof1@example.com1              | deputy-details |
 
+  # Client has new deputy and new org - delete client and expect knew one created
```
gregtyler: Sorry
```suggestion
  # Client has new deputy and new org - delete client and expect new one created
```

client/src/AppBundle/Controller/Admin/Client/ClientController.php
```diff
@@ -27,6 +29,53 @@ public function detailsAction($id)
...
+     * @Route("/{id}/discharge", name="admin_client_discharge", requirements={"id":"\d+"})
+     * @Security("has_role('ROLE_SUPER_ADMIN')")
+     * @param Request $request
+     * @param $id
+     *
+     * @Template("AppBundle:Admin/Client/Client:discharge.html.twig")
+     *
+     * @return array
+     */
+    public function dischargeAction(Request $request, $id)
```
gregtyler: Don't think request is required here
```suggestion
    public function dischargeAction($id)
```

client/src/AppBundle/Controller/Admin/Client/ClientController.php
```diff
@@ -27,6 +29,53 @@ public function detailsAction($id)
...
+    }
+
+    /**
+     * @Route("/{id}/discharge-confirm", name="admin_client_discharge_confirm", requirements={"id":"\d+"})
+     * @Security("has_role('ROLE_SUPER_ADMIN')")
+     * @param Request $request
+     * @param $id
+     * @return \Symfony\Component\HttpFoundation\RedirectResponse
+     */
+    public function dischargeConfirmAction(Request $request, $id)
```
gregtyler:
```suggestion
    public function dischargeConfirmAction($id)
```

client/src/AppBundle/Resources/translations/admin-clients.en.yml
```diff
@@ -23,6 +23,12 @@ clientDetails:
     address: Named deputy address
     contact: Named deputy contact details
 
+clientDischarge:
+  htmlTitle: Administration - Clients
+  pageTitle: Discharge deputy from client
+  warning: %clientFirstName% %clientLastName% will be permanently detached from their named deputy.
```
gregtyler: Do we need to warn that associated unsubmitted data (e.g. reports, documents) will be deleted too?

behat/tests/features/prof/03-report/09-deputy-discharged.feature
aiesoftware: I thought I had deleted it. Resolved now

behat/tests/features/prof/05-organisation/04-organisation-updates-acl.feature
aiesoftware: Done 👍

client/src/AppBundle/Controller/Admin/Client/ClientController.php
aiesoftware: Done 👍

client/src/AppBundle/Controller/Admin/Client/ClientController.php
aiesoftware: Done 👍

client/src/AppBundle/Resources/translations/admin-clients.en.yml
aiesoftware: Done 👍

PR #250, opened 2020-02-24
DDPB-3237: Add "synchronised" status and time to document model

gregtyler:
## Purpose
When we integrate with the API, we will automatically send documents to Sirius so that users don’t have to. However, users need to be aware of any failures or delay in the integration. Therefore, we will log which documents/submissions have been synchronised, which are queued, and which have failed.

Fixes DDPB-3237

## Approach
I've added the new fields to the Document model, and also a new `uuid` field for ReportSubmission, which we'll need for making subsequent API requests to Sirius.

I also wanted to do PHPStan properly in this ticket, so I upgraded it in both API and client, and added phpstan-prophecy on the client (with our current tests it isn't needed on API). The rest of the work was tidying up based on PHPStan's feedback. Most significantly, I retyped a bunch of variables in unit tests.

## Learning
The latest version of PHPStan is insistent on return types, either from actual type-hinting or from PHPDoc comments. That should make us specify functions more accurately in-place, and reduce the confusion of cross-file dependencies.

## Checklist
  - N/A
  - N/A
  - N/A

### DIFF
--- a/client/tests/phpunit/Service/DocumentServiceTest.php
+++ b/client/tests/phpunit/Service/DocumentServiceTest.php
@@ -14,6 +14,7 @@ use AppBundle\Service\File\Storage\S3Storage;
 use Doctrine\Common\Collections\ArrayCollection;
 use Exception;
 use PHPUnit\Framework\TestCase;
+use Prophecy\Argument;
 use Prophecy\Prophecy\ObjectProphecy;
 use Psr\Log\LoggerInterface;
 use Twig\Environment;
@@ -141,7 +142,7 @@ class DocumentServiceTest extends TestCase
             ->willThrow(Exception::class);
 
         $this->restClient
-            ->apiCall('DELETE', 'document/1', null, 'array', [], false)
+            ->delete(Argument::cetera())
             ->shouldNotBeCalled();
 
         $this->expectException(Exception::class);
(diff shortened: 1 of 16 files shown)


client/tests/phpunit/Service/DocumentServiceTest.php
```diff
@@ -320,17 +325,25 @@ public function testRetrieveDocumentsFromS3ByReportSubmissionsMissingDocs()
...
         $storage = self::prophesize(S3Storage::class);
+
+        /** @var ObjectProphecy|LoggerInterface $logger */
         $logger = self::prophesize(LoggerInterface::class);
+
+        /** @var ObjectProphecy|RestClient $restClient */
         $restClient = self::prophesize(RestClient::class);
+
+        /** @var ObjectProphecy|Environment $twig */
         $twig = self::prophesize(Environment::class);
```
acsauk-mojd: This test setup is repeated a few times above. Could we do this once in the `setUp()` block instead?

gregtyler: Yep, good point. I've converted the class properties to use prophecy and rewritten the tests to use them rather than setting up their own mocks/subject

aiesoftware:
Good stuff. Who generates and sets the uuid?

gregtyler:
> Good stuff. Who generates and sets the uuid?

Sirius will, as a way of telling us where to send supporting docs. So the requests will look something like:

```
> POST /clients/1234567t/reports (data and report PDF)
< 201 (submission uuid)
> POST /reports/{uuid}/supportingdocuments (transactions.csv)
> POST /reports/{uuid}/supportingdocuments (Statement-08-19.pdf)
> POST /reports/{uuid}/supportingdocuments (Jerry.jpg)
```

PR #251, opened 2020-02-24
DDPB-3102: Enable removing OPG Admin user accounts 

acsauk-mojd:
## Purpose
Enables super admins to delete admins and other super admins.

Fixes DDPB-3102

**EDIT** - Have just seen the huge file change, we should wait until #244 is merged before actioning this.

## Approach
The logic for displaying the delete button was getting messy in the view so I've extracted this out to the User model. This then led to me seeing how bloated the User model has become so I'm on the fence on if I should leave the logic in the model or create some kind of service for it 🤔once the Behat spike is merged I'll create a follow-up PR for this to add the delete Lay deputy feature using the new helper functions.

I've consciously left the Behat test as just the happy path for now as we discussed trying to drive more of the edge cases down into unit tests which I covered in `UserTest::canDeleteUser()`.

## Learning
The built-in Behat context helpers rarely appear in my IDE 😢 

## Checklist


api/src/AppBundle/Controller/UserController.php
```diff
@@ -251,17 +254,20 @@ public function getUserTeamNames(Request $request, $email)
      * //TODO move to UserService
      *
      * @Route("/{id}", methods={"DELETE"})
-     * @Security("has_role('ROLE_ADMIN')")
+     * @Security("has_role('ROLE_ADMIN') or has_role('ROLE_SUPER_ADMIN')")
```
aiesoftware: I think roles are configured hierarchically, so just `has_role('ROLE_ADMIN')` will implicitly include `ROLE_SUPER_ADMIN`.

api/src/AppBundle/Controller/UserController.php
```diff
@@ -251,17 +254,20 @@ public function getUserTeamNames(Request $request, $email)
...
      */
     public function delete($id)
     {
         /** @var User $user */
         $user = $this->userRepository->find($id);
 
-        if ($user->getRoleName() !== User::ROLE_LAY_DEPUTY) {
-            throw $this->createAccessDeniedException('Cannot delete users with role ' . $user->getRoleName());
+        if (!in_array($user->getRoleName(),self::DELETABLE_ROLES)) {
+            throw $this->createAccessDeniedException('Cannot delete ' . $user->getRoleName() . ' users');
```
aiesoftware: I'm a little on the fence with this, but from what I can tell, the restriction as to which roles can delete which users is encapsulated in the API consumer (client), instead of/along with the API itself. If we were to unit test this endpoint in isolation, or hit it via postman, the requirement to restrict deletion would not be met. I don't know if that's an issue, so any thoughts would be good.

behat/tests/behat.yml
```diff
@@ -73,6 +73,12 @@ default:
             contexts:
                 - DigidepsBehat\Registration\RegistrationFeatureContext
 
+        userManagement:
```
aiesoftware: Others have been in format `user-management`, would be good to make them consistent.

behat/tests/behat.yml
```diff
@@ -73,6 +73,12 @@ default:
...
+            description: Coverage of user management features
+            paths: [ "%paths.base%/features/userManagement" ]
+            contexts:
+                - DigidepsBehat\FeatureContext
```
aiesoftware: For the "new" structure that is domain/feature oriented, we have created a dir and namespace for each domain, e.g. `DigidepsBehat\UserManagement\UserManagementFeatureContext`, and in there is a FeatureContext file and a `UserManagementTrait` file. Would be good to adopt that here :)

api/src/AppBundle/Controller/UserController.php
acsauk-mojd: Awesome - I'll drop it.

behat/tests/behat.yml
acsauk-mojd: Yep makes sense -  I've added that now 👍

api/src/AppBundle/Controller/UserController.php
acsauk-mojd: Yeah that's a good point. We protect against unauthorised roles accessing the endpoint from the client but theres no protection for future changes to the code at the API level. It feels a bit off duplicating the logic in API and client but they are useful to have on both sides...what do you think?

gregtyler: I would opt to duplicate the logic. It's a PITA but means the rules are at the lowest level. I can imagine a future where we have an additional client app and it starts deleting users it shouldn't.

behat/tests/bootstrap/UserTrait.php
```diff
@@ -16,6 +18,35 @@ trait UserTrait
...
+        foreach ($table as $inputs) {
+            $this->assertValidAdminRole($inputs['adminType']);
+
+            $adminType = $inputs['adminType'];
+            $firstName = $inputs['firstName'];
+            $lastName = $inputs['lastName'];
+            $email = $inputs['email'];
+            $activated = $inputs['activated'];
+
+            $query = "adminType=$adminType&firstName=$firstName&lastName=$lastName&email=$email&activated=$activated";
```
gregtyler: I believe this could just be:
```php
$query = http_build_query($inputs);
```
Though that doesn't sanitise the query keys, so test-writers could add anything they like to the query string.

client/src/AppBundle/Controller/Admin/FixtureController.php
```diff
@@ -80,6 +80,53 @@ public function completeReportSectionsAction(Request $request, $reportId, Kernel
...
+     * @Route("/getUserIDByEmail/{email}", methods={"GET"})
+     * @Security("has_role('ROLE_SUPER_ADMIN') or has_role('ROLE_ADMIN') or has_role('ROLE_AD')")
+     */
+    public function getUserIDByEmail(KernelInterface $kernel, string $email)
+    {
+        if ($kernel->getEnvironment() === 'prod') {
+            throw $this->createNotFoundException();
+        }
+
+        /** @var \GuzzleHttp\Psr7\Response $response */
```
gregtyler:
```suggestion
        /** @var array $response */
```

client/src/AppBundle/Entity/User.php
```diff
@@ -1202,4 +1208,26 @@ public function belongsToActiveOrganisation(): bool
...
+        if ($userToBeDeleted->isAdminOrSuperAdmin()) {
+            return $this->getRoleName() === 'ROLE_SUPER_ADMIN' && $userToBeDeleted->getId() !== $this->getId() ? true : false;
+        }
+
+        if ($userToBeDeleted->getRoleName() === 'ROLE_LAY_DEPUTY' && count($userToBeDeleted->getClients()) <= 1 && $userToBeDeleted->hasReports() === false) {
+            return true;
+        }
+
+        return false;
+    }
```
gregtyler: I think this should be part of `UserVoter->determineDeletePermission`, rather than a helper on the model.

In twig, you can then just do `is_granted('delete', user)`.

api/src/AppBundle/Controller/UserController.php
aiesoftware: I tend to agree. Where practical, I prefer to write methods that protect themselves by making no assumptions about the context in which it was called, so if we can have the buck stop here that would be good.

client/src/AppBundle/Controller/Admin/IndexController.php
```diff
@@ -225,14 +225,17 @@ public function editNdrAction(Request $request, $id)
...
-        if (!$this->isGranted('ROLE_ADMIN')) {
-            throw new DisplayableException('Only Admin can delete users');
-        }
-
         /** @var EntityDir\User $loggedInUser */
         $loggedInUser = $this->getUser();
+
+        if (!$loggedInUser->isAdminOrSuperAdmin()) {
+            $message = $userToDelete->isAdminOrSuperAdmin() ? 'Only Super Admins can delete Admins or Super Admins' : 'Only Admins or Super Admins can delete users';
+            throw new DisplayableException($message);
```
gregtyler: Should probably be (to throw a generic 403 response):
```suggestion
            throw new AccessDeniedHttpException($message);
```
or (to throw a custom error message):
```suggestion
            return $this->renderError($message, Response::HTTP_FORBIDDEN)
```

client/src/AppBundle/Controller/Admin/FixtureController.php
acsauk-mojd: 🤦‍♂ hasty autocomplete strikes again

client/src/AppBundle/Entity/User.php
acsauk-mojd: Nice - I had a little voice at the back of my head while I was writing the tests that this feels like a voter now but ignored it as it was the end of the day 😂

client/src/AppBundle/Controller/Admin/IndexController.php
acsauk-mojd: Wasn't aware - I'll be on to trading standards about this error type.

behat/tests/bootstrap/UserTrait.php
acsauk-mojd: Much nicer - I've expanded the admin type assertion to check for the required keys too 👌

client/src/AppBundle/Controller/Admin/IndexController.php
```diff
@@ -231,14 +237,13 @@ public function deleteConfirmAction($id)
         /** @var EntityDir\User $loggedInUser */
         $loggedInUser = $this->getUser();
 
-        if (!$loggedInUser->isAdminOrSuperAdmin()) {
-            $message = $userToDelete->isAdminOrSuperAdmin() ? 'Only Super Admins can delete Admins or Super Admins' : 'Only Admins or Super Admins can delete users';
-            throw new DisplayableException($message);
-        }
+        $token = new UsernamePasswordToken($loggedInUser, 'credentials', 'memory');
+        $canDelete = $this->userVoter->vote($token, $userToDelete, [UserVoter::DELETE_USER]);
```
gregtyler: Another helper: this one will automatically generate a 403 if the grant isn't met.
```suggestion
        $this->denyAccessUnlessGranted('delete-user', $userToDelete, 'Unable to delete this user');
```

api/src/AppBundle/Controller/UserController.php
```diff
@@ -264,27 +279,33 @@ public function getUserTeamNames(Request $request, $email)
...
+        $deletee = $this->userRepository->find($id);
 
-        if (!in_array($user->getRoleName(),self::DELETABLE_ROLES)) {
-            throw $this->createAccessDeniedException('Cannot delete ' . $user->getRoleName() . ' users');
-        }
+        /** @var TokenInterface $user */
+        $token = $this->securityHelper->getToken();
 
-        $clients = $user->getClients();
+        $canDelete = $this->userVoter->vote($token, $deletee, [UserVoter::DELETE_USER]);
```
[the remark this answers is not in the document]
acsauk-mojd: So I've kept this in here as I think the benefit of seeing why the user can't delete the user outweighs the additional dep...what do you think?

client/src/AppBundle/Controller/Admin/IndexController.php
acsauk-mojd: Cool, makes sense here 👍

api/src/AppBundle/Controller/UserController.php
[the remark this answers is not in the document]
gregtyler: I'm not too fussed. It means we don't support multiple voters, but I'm not aware of why we would want to any time soon.

PR #255, opened 2020-02-26
DDPB-3202: Notify - Client and Deputy details change notification emails

acsauk-mojd:
## Purpose
Updating the client and deputy change details emails to use Notify.

Fixes DDPB-3202.

## Approach
Following along with the existing Notify setup this was quite straight forward. The only thing which was initially confusing was working out what the existing updateAddress function was being provided for `$response`.

## Learning
Some behat tests failed when we attempted to send params that were null/blank (which makes sense). Just something to keep in mind when/if we mock out the NotifyClient to match that behaviour.

## Checklist

### Frontend

### DIFF
--- a/api/src/AppBundle/v2/Fixture/Controller/FixtureController.php
+++ b/api/src/AppBundle/v2/Fixture/Controller/FixtureController.php
@@ -295,4 +295,52 @@ class FixtureController
 
         return $this->buildSuccessResponse($fromRequest, 'User created', Response::HTTP_OK);
     }
+
+    /**
+     * @Route("/createClientAttachDeputy", methods={"POST"})
+     * @Security("has_role('ROLE_ADMIN', 'ROLE_AD')")
+     */
+    public function createClientAndAttachToDeputy(Request $request)
+    {
+        $fromRequest = json_decode($request->getContent(), true);
+
+        $client = $this->clientFactory->create([
+            "firstName" => $fromRequest['firstName'],
+            "lastName" => $fromRequest['lastName'],
+            "phone" => $fromRequest['phone'],
+            "address" => $fromRequest['address'],
+            "address2" => $fromRequest['address2'],
+            "county" => $fromRequest['county'],
+            "postCode" => $fromRequest['postCode'],
+        ]);
+
+        /** @var User $deputy */
+        $deputy = $this->em->getRepository(User::class)->findOneBy(['email' => $fromRequest['deputyEmail']]);
+
+        if ($deputy === null) {
+            return $this->buildNotFoundResponse(sprintf("Could not find user with email address '%s'", $fromRequest['deputyEmail']));
+        }
+
+        $deputy->addClient($client);
+
+        $this->em->persist($deputy);
+        $this->em->flush();
+
+        return $this->buildSuccessResponse($fromRequest, 'User created', Response::HTTP_OK);
+    }
+
+    /**
+     * @Route("/getUserIDByEmail/{email}", methods={"GET"})
+     * @Security("has_role('ROLE_SUPER_ADMIN') or has_role('ROLE_ADMIN') or has_role('ROLE_AD')")
+     */
+    public function getUserIDByEmail(string $email)
+    {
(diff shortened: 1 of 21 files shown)


api/src/AppBundle/v2/Fixture/Controller/FixtureController.php
```diff
@@ -197,4 +198,53 @@ public function createUser(Request $request)
...
+        if ($deputy === null) {
+            return $this->buildNotFoundResponse(sprintf("Could not find user with email address '%s'", $fromRequest['deputyEmail']));
+        }
+
+        $deputy->addClient($client);
+
+        $this->em->persist($deputy);
+        $this->em->flush();
+
+        return $this->buildSuccessResponse($fromRequest, 'User created', Response::HTTP_OK);
```
gregtyler: FWIW, `Response::HTTP_OK` is the default value so not strictly necessary
```suggestion
        return $this->buildSuccessResponse($fromRequest, 'User created', Response::HTTP_OK);
```

behat/tests/bootstrap/UserManagement/UserManagementTrait.php
```diff
@@ -0,0 +1,44 @@
...
+
+        /** @var Session $session */
+        $session = $this->getSession();
+
+        if ($session->getStatusCode() !== Response::HTTP_OK) {
+            throw new \Exception($session->getPage()->getContent());
+        }
+
+        $id = $session->getPage()->getContent();
+        $this->visitAdminPath("/admin/edit-user?filter=$id");
```
gregtyler: For future discussion: I'm not clear how far we want to take these fixture help functions if the functionality is achievable through the UI.

For example, this could also be achieved by opening `/admin`, searching for the email address and clicking `behat-link-user-{condensedEmail}`.

behat/tests/features/UserManagement/editingUserDetails.feature
```diff
@@ -0,0 +1,115 @@
+@acs
+Feature: Editing Deputy and Client details
```
gregtyler:
```suggestion
Feature: Editing Deputy and Client details
```

behat/tests/features/UserManagement/editingUserDetails.feature
```diff
@@ -0,0 +1,115 @@
...
+    Given the following users exist:
+      | ndr      | deputyType | firstName | lastName    | email                 | postCode | activated |
+      | disabled | LAY        | Hena      | Mercia      | hena.mercia@test.com  | HA4      | true      |
+
+    Given the following clients exist and are attached to deputies:
+      | firstName | lastName | phone       | address     | address2  | county  | postCode | caseNumber | deputyEmail            |
+      | Jory      | Dunkeld  | 01215552222 | 1 Fake Road | Fakeville | Faketon | B4 6HQ   | JD123456   |  hena.mercia@test.com  |
+
+  @deputy
+  Scenario: Editing deputy details
```
gregtyler:
```suggestion
  Scenario: Editing deputy details
```

behat/tests/bootstrap/UserManagement/UserManagementTrait.php
acsauk-mojd: True - it would be interesting to see if there's any difference in speed between API calls and page renders. I always tend to prefer going straight to the source if possible rather than driving a page but good to chat through pros and cons.

behat/tests/features/UserManagement/editingUserDetails.feature
acsauk-mojd: Gah - knew I'd leave one in. 

Incidentally, I tried to set up a generic debug suite in `behat.yml` but it was a no-go due to the way we are setting up the separate fixture contexts and requiring traits. As we would need all feature contexts in the debug suite it throws an error when trying to require traits more than once.

api/src/AppBundle/v2/Fixture/Controller/FixtureController.php
acsauk-mojd: Yup! I'll drop.

PR #256, opened 2020-02-27
DDPB-3204: Use Notify for activation and password reset emails

gregtyler:
## Purpose
Convert emails to use GOV.UK Notify:
- Password forgotten
- User activation
- Invited user activation

Also deprecate as not in use:
- Password changed

Fixes DDPB-3204

## Approach
The underlying change here was really easy, but it lead down a path of having to update lots of tests because they rely on "clicking the link from the activation email".

Initially, I switched the emails across to the Notify format with unit tests. I also added a new "invitation" email, for when organisation admins invite a new team member. These have a different wording to "normal" activation emails.

Then I updated tests. Where previously they would pick the activation link from the email, they now grab it from a fixture URL. This allows us to still run e2e tests programatically, but without having to rely on emails.

I took more care in some tests than others: in older tests that we want to deprecate (like teams), I just replaced the "get link from email" with "get link from fixture". I rewrote more modern tests to just generally make more sense (like creating a user that's already activated rather than creating then activating).

## Learning
This is the first time I've spent looks at tests aggressively since we started adopting the new style, and it's helped me see the value and practicalities of it. If we can continue to convert and rewrite tests as makes sense, I think we'll be able to ditch a lot of the old suites quite soon.

## Checklist
  - N/A


behat/tests/bootstrap/Common/AuthenticationTrait.php
```diff
@@ -41,6 +41,26 @@ public function iAmAtAdminLogin()
         $this->visitAdminPath('/login');
     }
 
+    /**
+     * @When /^I open the (admin |)(activation|password reset) page for "(.+)"$/
+     */
+    public function openActivationOrPasswordResetPage($admin, $pageType, $email)
+    {
+        $url = sprintf('/admin/fixtures/user-registration-token?email=%s', $email);
+        $this->iAmLoggedInToAdminAsWithPassword('admin@publicguardian.gov.uk', 'Abcd1234');
```
acsauk-mojd: Don't know why I didn't think about putting the login step in the helper functions that need it 🤦‍♂

behat/tests/features/OrganisationManagement/edit-members.feature
```diff
@@ -0,0 +1,150 @@
...
+      | Rey Gusciora       | guscioraonline@gmail.example | true      |
+    And the following users exist:
+      | ndr      | deputyType | firstName | lastName | email                        | postCode | activated |
+      | disabled | PROF       | Lavern    | Degroot  | l.degroot@malvern.example    | SW1H 9AJ | true      |
+      | disabled | PROF       | Landon    | Swayzer  | l.swayzer@outsource.example  | SW1H 9AJ | true      |
+      | disabled | PROF       | Rey       | Gusciora | guscioraonline@gmail.example | SW1H 9AJ | true      |
+    And the following users are in the organisations:
+      | userEmail                    | orgName            |
+      | l.degroot@malvern.example    | Malvern Solicitors |
+      | guscioraonline@gmail.example | Rey Gusciora       |
```
acsauk-mojd: This reads really well now 👍

client/src/AppBundle/Service/Mailer/MailFactory.php
```diff
@@ -90,30 +93,51 @@ private function generateAbsoluteLink($area, $routeName, array $params = [])
...
-            'recipientRole' => self::getRecipientRole($user)
+            'registerLink' => $this->generateAbsoluteLink($area, 'register'),
         ];
 
-        $email = new ModelDir\Email();
+        $email = (new ModelDir\Email())
+            ->setFromEmailNotifyID(self::NOTIFY_FROM_EMAIL_ID)
+            ->setFromName($this->translate('activation.fromName'))
+            ->setToEmail($user->getEmail())
+            ->setToName($user->getFullName())
```
acsauk-mojd: Did we decide we don't need to set this now?

client/src/AppBundle/Service/Mailer/NotifyClientMock.php
```diff
@@ -0,0 +1,22 @@
...
+
+    public function sendEmail($emailAddress, $templateId, array $personalisation = array(), $reference = '', $emailReplyToId = NULL)
+    {
+        if ($emailAddress === 'break@publicguardian.gov.uk') {
+            throw new NotifyException('Intentional mock exception');
+        } else {
+            return [];
+        }
+    }
+}
```
acsauk-mojd: 🙌

behat/tests/features/OrganisationManagement/edit-members.feature
gregtyler: It's so nice not having to remember what state `behat-user-prof1` is in 🙏

client/src/AppBundle/Service/Mailer/MailFactory.php
gregtyler: Good point, fixed

behat/tests/features/OrganisationManagement/edit-members.feature
aiesoftware: This may not be entirely relevant here, but for future, there is a `Given the following court orders exist:` trait method, which when given a `PA` or `PROF` deputy type, will create them an org and add them to it. If org exists, it will add them to existing.

client/src/AppBundle/Service/Mailer/MailFactory.php
```diff
@@ -90,30 +93,49 @@ private function generateAbsoluteLink($area, $routeName, array $params = [])
...
+        ];
+
+        $email = (new ModelDir\Email())
+            ->setFromEmailNotifyID(self::NOTIFY_FROM_EMAIL_ID)
             ->setFromName($this->translate('activation.fromName'))
             ->setToEmail($user->getEmail())
-            ->setToName($user->getFullName())
-            ->setSubject($this->translate('activation.subject'))
-            ->setBodyHtml($this->templating->render('AppBundle:Email:user-activate.html.twig', $viewParams))
-            ->setBodyText($this->templating->render('AppBundle:Email:user-activate.text.twig', $viewParams));
```
aiesoftware: Can we delete these files now?

gregtyler: Yep, good point. I'll clean that up.

PR #257, opened 2020-02-29
DDPB:3203: Migrate NDR and Report submission confirmation emails to N…

aiesoftware:
…otify templates

## Purpose
Sends NDR and Report submission emails via Notify.

Fixes DDPB-3203

## Approach
Straight swap from sending via Twig templates to sending via Notify templates.

Removed the email that sends the NDR PDF to the user - this won't be supported in Notify, and the PDF is accessible to the user from within DD.

Added support to Court Order fixture endpoint to create submittable NDR's.

## Checklist

### Frontend


api/src/AppBundle/v2/Fixture/Controller/FixtureController.php
```diff
@@ -66,7 +68,13 @@ public function createCourtOrderAction(Request $request)
 
         $client = $this->createClient($fromRequest);
         $deputy = $this->createDeputy($fromRequest);
-        $this->createReport($fromRequest, $client);
+
+        if (strtolower($fromRequest['reportType']) == 'ndr') {
```
gregtyler: Fussy, but I have a twitch about this from JavaScript:
```suggestion
        if (strtolower($fromRequest['reportType']) === 'ndr') {
```

PR #259, opened 2020-03-04
DDPB-3246: Handling no corref from sirius

aiesoftware:
## Purpose
To accept the CSV from Sirius, which does not contain a corref field.

Fixes DDPB-3246

## Approach
Whilst we don't get a corref in the CSV, we still need to set one internally as it is used to determine the report type throughout the app. Therefore, I am hardcoding the corref internally to a value that I know will evaluate towards the correct report type. Specifically, `OPG102 => L2` and `OPG103 => L3`. Any value other than `OPG102|OPG103` will be ignored, temporarily at least.

## Learning
Accepting a Sirius CSV has been an iterative process whilst liaising with Sirius team, which has made changes like this nicely isolated and easily tested. The structure of our code also makes any tweaks to ingested data contained and maintainable. 

## Checklist

### Frontend


behat/tests/features/v2/registration/lay-deputy-upload.feature
```diff
@@ -10,10 +10,10 @@ Feature: Lay deputy upload into registration lookup table
     Then I should see "25 users in the database"
     # Upload 15 users from a sirius source
```
gregtyler:
```suggestion
    # Upload 10 users from a sirius source
```

PR #260, opened 2020-03-04
DDPB-3264: Write unit tests for activation/reset emails

gregtyler:
## Purpose
In the process of switching to Notify, we’ve stopped checking email sending in our end-to-end tests. Instead, we should write unit tests for the relevant controllers to ensure emails are being sent as expected.

This is particularly essential for our activation and password reset emails, since they are critical for users to access DigiDeps. Unlike other emails, there is no work around and they permanently break the user journey.

We are prioritising writing these tests to give a clear plan of  how we can increase coverage to other controllers/emails.

Fixes DDPB-3264

## Approach
I went around most ways of writing unit tests and mocking things in Symfony, and settled on a method that I think works quite smoothly and is ripe for extending.

It uses Symfony's `createClient` function to create a driver for the tests, but:
- Uses a new `unittest` environment which has CSRF and Redis sessions switched off (to make testing easier)
- Exposes the client's `Container` so that services can easily be overwritten.

This allows you to easily prophesize a service and mock calls to it:
```php
$this->injectProphecyService(MailSender::class, function($mailSender) {
    $mailSender->send(Argument::cetera())->shouldNotBeCalled();
});

$crawler = $this->client->request('GET', "/org/settings/organisation/14/add-user");
```

I also added some helper functions: `injectProphecyService` (used above) quickly replaces a service with a prophet which you can hook into. `mockLoggedInUser` mocks token storage and rest client services so that you're treated as a logged in user with the supplied role(s).

## Learning
I hit a lot of stumbling blocks on this, which was frustrating but a lot of useful learning. A few take-aways:
- If you use `'debug' => true` in the client, the tests take several times longer
- By default, the client reloads the kernel between each request (deregistering our mocks). This is disabled with `$client->disableReboot()`
- You need to manually follow redirects (we could turn this on by default, but it would disrupt tests that don't want it)
- Some services need to registered twice: once with their class name and once with the shortcut (e.g. `RestClient` and `rest_client`)
- We have to remember to mock things we don't need (e.g. if you don't mock `Predis`, it'll interact with the real Redis service)
- Because Twig isn't mocked by default, you often need to provide more data than the action actually needs. You can mitigate this by mocking Twig or by using dummies (e.g. `new User()` rather than `self::prophesize(User::class)`
  - On the flip side, not mocking Twig means that we can test the templates/translations which can be clearer

## Checklist
  - N/A
  - Ohhhh boy have I
  - N/A

### DIFF
--- a/client/tests/phpunit/Controller/AbstractControllerTestCase.php
+++ b/client/tests/phpunit/Controller/AbstractControllerTestCase.php
@@ -2,10 +2,12 @@
 
 namespace AppBundle\Controller;
 
+use AppBundle\Entity\User;
 use Symfony\Bundle\FrameworkBundle\Client;
-use Symfony\Bundle\FrameworkBundle\Controller\Controller;
 use Symfony\Bundle\FrameworkBundle\Test\WebTestCase;
-use Symfony\Component\DependencyInjection\Container;
+use Symfony\Component\Security\Core\Authentication\Token\Storage\TokenStorage;
+use Symfony\Component\Security\Core\Authentication\Token\TokenInterface;
+use Symfony\Component\Security\Core\Role\Role;
 
 abstract class AbstractControllerTestCase extends WebTestCase
 {
@@ -15,5 +17,26 @@ abstract class AbstractControllerTestCase extends WebTestCase
     public function setUp(): void
     {
         $this->client = static::createClient(['environment' => 'unittest', 'debug' => false]);
+    }
+
+    protected function mockLoggedInUser(array $roleNames, User $user = null): void
+    {
+        $container = $this->client->getContainer();
+
+        $roles = array_map(function () {
+            return new Role('ROLE_ADMIN');
+        }, $roleNames);
+
+        $token = self::prophesize(TokenInterface::class);
+        $token->getUser()->willReturn(is_null($user) ? new User() : $user);
+        $token->serialize()->willReturn('');
+        $token->isAuthenticated()->willReturn(true);
+        $token->getRoles()->willReturn($roles);
+
+        $tokenStorage = self::prophesize(TokenStorage::class);
+        $tokenStorage->getToken()->willReturn($token);
+        $tokenStorage->setToken(null)->willReturn();
+
+        $container->set('security.token_storage', $tokenStorage->reveal());
     }
 }
--- a/client/tests/phpunit/Controller/AbstractControllerTestCase.php
+++ b/client/tests/phpunit/Controller/AbstractControllerTestCase.php
@@ -3,10 +3,12 @@
 namespace AppBundle\Controller;
 
 use AppBundle\Entity\User;
+use AppBundle\Service\DeputyProvider;
+use Prophecy\Argument;
 use Symfony\Bundle\FrameworkBundle\Client;
 use Symfony\Bundle\FrameworkBundle\Test\WebTestCase;
 use Symfony\Component\Security\Core\Authentication\Token\Storage\TokenStorage;
-use Symfony\Component\Security\Core\Authentication\Token\TokenInterface;
+use Symfony\Component\Security\Core\Authentication\Token\UsernamePasswordToken;
 use Symfony\Component\Security\Core\Role\Role;
 
 abstract class AbstractControllerTestCase extends WebTestCase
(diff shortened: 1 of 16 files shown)


client/tests/phpunit/Controller/AbstractControllerTestCase.php
```diff
@@ -2,77 +2,80 @@
...
+        // Mock token storage to return our fake token
+        $this->injectProphecyService(TokenStorage::class, function($tokenStorage) use ($token) {
+            $tokenStorage->getToken()->willReturn($token);
+            $tokenStorage->setToken(Argument::cetera())->willReturn();
+        }, ['security.token_storage']);
 
-        return $this->frameworkBundleClient->getResponse();
+        // Respond to calls to hydrate user details from API
+        $this->restClient->setLoggedUserId(1)->willReturn($this->restClient->reveal());
+        $this->restClient->get('user/1', Argument::cetera())->willReturn($user);
```
acsauk-mojd: Now I'm over my aghastness at setting stuff on the container during test runs, this is really cool! Its going to make testing a lot easier on the whole and opens up TDD to a whole new controller level 🎉

client/tests/phpunit/Controller/OrganisationControllerTest.php
```diff
@@ -0,0 +1,138 @@
...
+        $this->restClient->put('v2/organisation/14/user/21', '')->shouldBeCalled()->willReturn($user);
+
+        $this->injectProphecyService(MailSender::class, function($mailSender) use ($emailAddress) {
+            $mailSender->send(Argument::that(function ($email) use ($emailAddress) {
+                return $email instanceof Email
+                    && $email->getToEmail() === $emailAddress
+                    && $email->getTemplate() === MailFactory::INVITATION_TEMPLATE_ID
+                    && strpos($email->getParameters()['link'], "user/activate/invitation-token") !== false;
+            }), Argument::cetera())->shouldBeCalled()->willReturn();
+        });
```
acsauk-mojd: I guess this is the drawback of the method - its not super easy to see whats going on here. But I'll take it over not being able to test controllers!

gregtyler: One option is to break it down into separate `Argument` calls rather than use a callback. I'm not convinced that's more readable though, because the language of Prophecy is a bit confusing.

```suggestion
        $this->injectProphecyService(MailSender::class, function($mailSender) use ($emailAddress) {
            $mailSender
                ->send(Argument::allOf(
                    Argument::type(Email::class),
                    Argument::which('getToEmail', $emailAddress),
                    Argument::which('getTemplate', MailFactory::INVITATION_TEMPLATE_ID),
                    Argument::that(function ($email) {
                        return strpos($email->getParameters()['link'], "user/activate/invitation-token") !== false;
                    })
                ), Argument::any())
                ->shouldBeCalled()
                ->willReturn();
        });
```

That said, maybe I could just lay it out a bit clearer:
```suggestion
        $mailSender = $this->injectProphecyService(MailSender::class);
        $mailSender
            ->send(Argument::that(function ($email) use ($emailAddress) {
                return $email instanceof Email
                    && $email->getToEmail() === $emailAddress
                    && $email->getTemplate() === MailFactory::INVITATION_TEMPLATE_ID
                    && strpos($email->getParameters()['link'], "user/activate/invitation-token") !== false;
            }), Argument::any())
            ->shouldBeCalled()
            ->willReturn();
```

client/tests/phpunit/Controller/OrganisationControllerTest.php
```diff
@@ -36,14 +36,16 @@ public function testAddAction(): void
...
+            ->send(Argument::that(function ($email) use ($emailAddress) {
                 return $email instanceof Email
                     && $email->getToEmail() === $emailAddress
                     && $email->getTemplate() === MailFactory::INVITATION_TEMPLATE_ID
                     && strpos($email->getParameters()['link'], "user/activate/invitation-token") !== false;
-            }), Argument::cetera())->shouldBeCalled()->willReturn();
-        });
+            }))
+            ->shouldBeCalled()
+            ->willReturn();
```
acsauk-mojd: Much more readable 👍

aiesoftware:
👏

PR #267, opened 2020-03-09
DDPB-3258: Implement feature flag on document sync service

gregtyler:
## Purpose
The daemon created in #266 should be controllable by a feature flag, so that we can pause all documents from being sent to Sirius (for example, in the event of downtime or a problem).

Fixes DDPB-3258

## Approach
I added a feature flag in Parameter Store and built a service to retrieve it, in a format that we can scale for more feature flags if we want to go further down this route. I also updated the command so that it checks this feature flag before trying to synchronise documents.

Most of my time on this was spent getting environment variables set up on all the right containers, and sorting out permissions. This has resulted in a bunch of new terraform config to set up a VPC endpoint, security groups and permissions. Again though, this is all reusable if we want to use feature flags in other services.

## Learning
I encountered [a bug in localstack](https://github.com/localstack/localstack/issues/1184) which makes this hard to emulate locally (though it works fine in production, in a bizarre twist of the norm). This means that although we create the parameter in the localstack startup script, it's not available through the AWS client on the frontend container.

I resolved this by creating [a PHP script](https://github.com/ministryofjustice/opg-digideps/blob/8bcad1e4b75687fa0deaafb37c7dbd5d28602547/client/scripts/localstack-init.php) to initialise the SSM part of localstack. This is run automatically as part of `docker-compose up`, and can be run manually also. I don't like this solution, but I believe it's necessary until the underlying problem with localstack is fixed.

## Checklist
  - Updated the PUML/traffic flow diagram
  - N/A


aiesoftware:
Great job!

PR #271, opened 2020-03-12
DDPB-3058: Improve service health and stabilization

gregtyler:
## Purpose
During a recent deployment the stabilise task timed out from what looked like a slow start-up time on the api container which resulted in the health check timeout of 30 seconds being exceeded. After inspecting the containers there were no obvious faults and a rerun resulted in a successful stabilise. 

We should look at ways to avoid this in the future.

Fixes DDPB-3058

## Approach
I've made two major changes here. Firstly I sped up the starting of containers by warming up their cache as part of the build process. In doing this I deleted a command we don't need any more because it was throwing errors in the warmup process.

I've also added healthchecks to our containers where suitable. This gives ECS a way of knowing whether they're healthy or not, and waiting for 90 seconds (configurable) for them to get into shape. ECS will continually monitor this to know whether to restart a container, rather than relying on the ELB healthcheck or just waiting for an exit code.

## Learning
I'm not sure healthchecks are still 100% at their best, but this is a big step forward. It should also unlock autoscaling, since new containers will be able to spin up in a reasonable time now.

Shout-out to DDPB-2136 where I first learnt about ECS healthchecks.

## Checklist
  - N/A
  - N/A
  - N/A


environment/admin_service.tf
```diff
@@ -49,6 +49,15 @@ locals {
       "protocol": "tcp"
     }],
     "volumesFrom": [],
+    "healthCheck": {
+      "command": [
+        "CMD-SHELL",
+        "curl -f -k https://localhost:443/manage/elb || exit 1"
+      ],
+      "interval": 30,
```
aiesoftware: Is this supposed to be 90, or is `interval` referring to something different that is mentioned in the PR description?

acsauk-mojd: I'm reading this as being three 30 second attempts...but true, couldn't this be one 90 second attempt?

gregtyler: Yep, it's three 30 seconds attempts, which is the default configuration from AWS (I had to specify it here because Terraform throws a wobbly otherwise).

The benefit of having more regular attempts is that it can stabilise quicker (in practice, it should be ready in <30 seconds). I was tempted to do a check every five seconds, but it's not really necessary and I don't know what's happening on the AWS side that might not be happy about that.

PR #273, opened 2020-03-16
DDPB-3259 - Upload queued documents to Sirius from daemon command

acsauk-mojd:
## Purpose
This ticket adds the plumbing into the existing sync document command to actually send the document to API Gateway and persist the returned UUID against he report submission and sync status against the Document.

Fixes DDPB-3259.

## Approach
There are a few callouts that happen as part of this command:

`S3` - to get the document contents
`Sirius API Gateway` - To send the Report Document and retrieve the UUID and send supporting documents once the report Document has been successfully sent.
`API` - to persist the result of the sync against the Document

I've tried to keep the logic for making these calls encapsulated in the Entity they relate to but the function is still a bit larger than I would have liked. I considered extracting some of the error handling out into their own classes but I'm not sure if it would increase the readability enough to warrant it.

## Learning

## Checklist

### Frontend


client/app/config/services.yml
```diff
@@ -45,10 +45,31 @@ services:
...
 
     Aws\Ssm\SsmClient:
         arguments: ["%ssm_client_params%"]
 
     AppBundle\Service\FeatureFlagService:
         arguments: ['@Aws\Ssm\SsmClient', '%env(FEATURE_FLAG_PREFIX)%']
+
+    AppBundle\Serializer\SiriusDocumentUploadSerializer:
+        autowire: true
+        autoconfigure: true
```
acsauk-mojd: Still not sure why autoconfig and autowire are being ignored in defaults.yml...

client/src/AppBundle/Service/DocumentSyncService.php
```diff
@@ -24,87 +29,223 @@ class DocumentSyncService
     /** @var S3Storage */
     private $storage;
 
-    public function __construct(S3Storage $storage)
+    /**
+     * @var SiriusApiGatewayClient
+     */
+    private $siriusApiGateWayClient;
```
gregtyler: A minor thing, but the casing of this isn't quite right
```suggestion
    private $siriusApiGatewayClient;
```

client/src/AppBundle/Service/DocumentSyncService.php
```diff
@@ -24,87 +29,223 @@ class DocumentSyncService
...
-        } else if (in_array($report->getType(), [Report::TYPE_HEALTH_WELFARE, Report::TYPE_COMBINED_HIGH_ASSETS, Report::TYPE_COMBINED_LOW_ASSETS])) {
-            $type = 'HW';
+        if (!$document->supportingDocumentCanBeSynced()) {
+            return $this->handleDocumentStatusUpdate($document, Document::SYNC_STATUS_QUEUED);
+        }
+
+        $content = $this->retrieveDocumentContentFromS3($document);
+
+        if ($content instanceof Throwable) {
+            return null;
```
gregtyler: This should use `handleDocumentStatusUpdate` to set status/error message like the report PDF does. Might be worth including that setting in `retrieveDocumentContentFromS3`, `handleSiriusSync` rather than handling it in the sync functions?

Alternatively, something like this (which I prefer because I'm not a fan of returning errors from functions):
```
try {
  $content = $this->retrieveDocumentContentFromS3($document);
  $this->handleSiriusSync($document, $content);
  return $this->handleDocumentStatusUpdate($document, Document::SYNC_STATUS_SUCCESS);
} catch (Throwable $e) {
  return $this->handleDocumentStatusUpdate($document, Document::SYNC_STATUS_PERMANENT_ERROR, $e->getMessage());
}
```

client/src/AppBundle/Entity/Report/Document.php
```diff
@@ -343,6 +353,26 @@ public function isAdminDocument()
...
+    {
+        if ($this->isReportPdf()) {
+            return $this->getReportSubmission();
+        }
+
+        foreach ($this->getReport()->getReportSubmissions() as $submission) {
+            if ($submission->hasReportPdf()) {
+                return $submission;
+            }
+        }
```
gregtyler: I'm not sure this will handle resubmissions (after OPG has returned the report to the deputy) properly. It looks like it will always return the _first_ submission with a PDF when we probably want it to return the _previous_.

So we might need to do something more like:
```suggestion
        $previousSubmissions = array_filter($this->getReport()->getReportSubmissions(), function($submission) { $submission->createdOn <= $this->getReportSubmission()->createdOn; });
        foreach (array_reverse($previousSubmissions) as $submission) {
            if ($submission->hasReportPdf()) {
                return $submission;
            }
        }
```

client/tests/phpunit/Pact/SiriusApiGatewayClientTest.php
```diff
@@ -0,0 +1,244 @@
...
+    public function setUp(): void
+    {
+        $this->caseRef = '1234567T';
+        $this->reportPdfUuid = '33ea0382-cfc9-4776-9036-667eeb68fa4b';
+        $this->expectedSupportingDocumentUuid = '9c0cb55e-718d-4ffb-9599-f3164e12dbdb';
+        $this->client = new GuzzleClient();
+        $this->signer = self::prophesize(RequestSigner::class);
+        $this->baseUrl = getenv('PACT_MOCK_SERVER_HOST');
+        $this->serializer = (self::bootKernel(['debug' => false]))->getContainer()->get('serializer');
+        $this->sut = new SiriusApiGatewayClient($this->client, $this->signer->reveal(), 'http://' . $this->baseUrl, $this->serializer);
```
gregtyler: Can simplify this a bit since there's no need to set them as properties (as they're never reused):
```suggestion
        $client = new GuzzleClient();
        $this->signer = self::prophesize(RequestSigner::class);
        $baseUrl = getenv('PACT_MOCK_SERVER_HOST');
        $serializer = (self::bootKernel(['debug' => false]))->getContainer()->get('serializer');
        $this->sut = new SiriusApiGatewayClient($client, $this->signer->reveal(), 'http://' . $baseUrl, $serializer);
```

client/tests/phpunit/Pact/SiriusApiGatewayClientTest.php
```diff
@@ -0,0 +1,244 @@
...
+                        'date_submitted' => $matcher->dateTimeWithMillisISO8601(),
+                        'type' => $matcher->regex('PF', 'PF|HW|NDR'),
+                        'submission_id' => $matcher->integer(9876)
+                    ]
+                ],
+            ]);
+
+        // Create a configuration that reflects the server that was started. You can create a custom MockServerConfigInterface if needed.
+        $config  = new MockServerEnvConfig();
+        $builder = new InteractionBuilder($config);
```
gregtyler: Might be worth moving this into the constructor as `$this->builder` to avoid duplication?

client/tests/phpunit/Pact/SiriusApiGatewayClientTest.php
```diff
@@ -0,0 +1,244 @@
...
+
+        return $builder;
+    }
+
+    public function uploadProvider()
+    {
+        return [
+
+        ];
+    }
```
gregtyler: Looks like this isn't necessary?
```suggestion
```

client/tests/phpunit/Service/Client/Sirius/SiriusApiGatewayClientTest.php
```diff
@@ -0,0 +1,165 @@
...
+use DigidepsTests\Helpers\SiriusHelpers;
+use GuzzleHttp\Client;
+use GuzzleHttp\Psr7\Request;
+use GuzzleHttp\Psr7\Response;
+use Prophecy\Argument;
+use Prophecy\Prophecy\ObjectProphecy;
+use Symfony\Bundle\FrameworkBundle\Test\KernelTestCase;
+use Symfony\Component\Serializer\Serializer;
+
+class SiriusApiGatewayClientTest extends KernelTestCase
```
gregtyler: Is this file still needed now we have the other `SiriusApiGatewayClientTest`?

client/tests/phpunit/phpunit.xml
```diff
@@ -10,7 +10,7 @@
          convertNoticesToExceptions="true"
          convertWarningsToExceptions="true"
         stopOnError="true"
-        stopOnFailure="true"
+        stopOnFailure="false"
```
gregtyler: Presumably this was just for testing?
```suggestion
        stopOnFailure="true"
```

client/src/AppBundle/Service/Client/Sirius/SiriusDocumentUpload.php
```diff
@@ -0,0 +1,52 @@
+<?php declare(strict_types=1);
+
+
+namespace AppBundle\Service\Client\Sirius;
```
gregtyler: I think this should be in `AppBundle\Model\Sirius` since it's a model, not a service (same with the others)

client/tests/phpunit/Helpers/FileHelpers.php
```diff
@@ -0,0 +1,28 @@
...
+     * @param string $fileLocation
+     * @param string $originalName
+     * @param string $mimeType
+     * @return UploadedFile
+     */
+    static public function generateUploadedFile(string $fileLocation, string $originalName, string $mimeType)
+    {
+        //@TODO drop the /../ file path when projectDir works as expected in Symfony 4+
+        $projectDir = (self::bootKernel(['debug' => false]))->getProjectDir();
+        $location = sprintf('%s/../%s', $projectDir, $fileLocation);
```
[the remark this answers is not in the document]
acsauk-mojd: I went down the route of trying to mock out the chain of calls down `UploadedFile` and its parent `File` but part of the `File` constructor checks that a real file exists at a path so I couldn't think of how to get around mocking that...open to suggestions if you've managed it in the past 🤷‍♂

client/src/AppBundle/Service/DocumentSyncService.php
acsauk-mojd: I remember the day I was working on the error handling I was having an existential crisis about returning or throwing errors. I suspect I ended up on this solution as a temporary one and never came back to sorting it properly. I like your suggestion though so will go with that 👍

client/src/AppBundle/Entity/Report/Document.php
acsauk-mojd: Ah I hadn't considered that - I've updated to use a filtered list now.

client/tests/phpunit/Pact/SiriusApiGatewayClientTest.php
acsauk-mojd: Yup!

client/tests/phpunit/Service/Client/Sirius/SiriusApiGatewayClientTest.php
acsauk-mojd: Hmmmm I guess we are doubling up a little here. I'll leave it with the get request test in for now. We'll need to update that to Pact too. I'll add a ticket for it.

client/tests/phpunit/phpunit.xml
acsauk-mojd: Ah yeah, I like to see all the broken tests on a change rather than stopping at the first. I'll revert.

client/src/AppBundle/Service/Client/Sirius/SiriusDocumentUpload.php
acsauk-mojd: Agree! Moving over now.

client/tests/phpunit/Service/Client/Sirius/SiriusApiGatewayClientTest.php
gregtyler: Fair enough 👍

client/tests/phpunit/Entity/Report/DocumentTest.php
```diff
@@ -0,0 +1,73 @@
...
+        $newReportPdfDocument->setReportSubmission($newReportPdfSubmission);
+        $supportingDocumentSubmittedWithReport->setReportSubmission($oldReportPdfSubmission);
+        $supportingDocumentSubmittedAfterReport->setReportSubmission($supportingDocumentSubmission);
+
+        $report->setReportSubmissions([$newReportPdfSubmission, $oldReportPdfSubmission, $supportingDocumentSubmission]);
+
+        self::assertEquals($oldReportPdfSubmission, $oldReportPdfDocument->getPreviousReportPdfSubmission());
+        self::assertEquals($oldReportPdfSubmission, $newReportPdfDocument->getPreviousReportPdfSubmission());
+        self::assertEquals($oldReportPdfSubmission, $supportingDocumentSubmittedWithReport->getPreviousReportPdfSubmission());
+        self::assertEquals($oldReportPdfSubmission, $supportingDocumentSubmittedAfterReport->getPreviousReportPdfSubmission());
```
gregtyler: Should this be `$newReportPdfSubmission`?

client/src/AppBundle/Service/DocumentSyncService.php
gregtyler: Looks great 👍

client/src/AppBundle/Service/DocumentSyncService.php
gregtyler: Needs updating a few other places in that file, sorry didn't clock that 🤦‍♂

client/tests/phpunit/Helpers/FileHelpers.php
```diff
@@ -0,0 +1,28 @@
+<?php declare(strict_types=1);
+
+
+namespace DigidepsTests\Helpers;
+
+
+use Symfony\Bundle\FrameworkBundle\Test\KernelTestCase;
+use Symfony\Component\HttpFoundation\File\UploadedFile;
+
+class FileHelpers extends KernelTestCase
```
gregtyler: Don't think this file is needed any more?

client/src/AppBundle/Service/DocumentSyncService.php
acsauk-mojd: My bad! Updated now.

client/tests/phpunit/Entity/Report/DocumentTest.php
acsauk-mojd: This was more to show that when a new submission is made we still get the previous submission back.

acsauk-mojd:
Confirmed with integrations (https://mojdt.slack.com/archives/CBKSZGH0C/p1585070051112700) that they don't have the error handling code in place yet so I'm proposing we leave the Pact tests as they are with just the happy path.

PR #281, opened 2020-03-19
DDPB-3233: Update client search to prevent default search

aiesoftware:
## Purpose
Prevents client search page from searching and returning full list on page load. Pressing submit with empty criteria will return that full list if required.

Fixes DDPB-3233

## Approach
The implementation itself is a straightforward case of not firing the query if no params are sent in the request, i.e on initial page load.

This had a knock on effect to several behat tests which rely on the results returning in order to navigate to a client detail page. To bypass this cumbersome navigation, I have created an endpoint that retrieves client data by `caseNumber`: our tests always have access to a case number so we can hit this new URL directly, instead of navigating via the search. The new endpoint itself simply retrieves the ID of that case and performs a redirect to the exist id based URL.

## Learning

## Checklist

### Frontend

### DIFF
--- a/api/src/AppBundle/Entity/Repository/ClientRepository.php
+++ b/api/src/AppBundle/Entity/Repository/ClientRepository.php
@@ -122,7 +122,7 @@ class ClientRepository extends EntityRepository
 
     /**
      * @param $caseNumber
-     * @return |null
+     * @return array<mixed>|null
      */
     public function getArrayByCaseNumber($caseNumber)
     {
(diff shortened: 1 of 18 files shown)


api/src/AppBundle/Entity/Repository/ClientRepository.php
```diff
@@ -120,6 +120,22 @@ public function getArrayById($id)
         return count($result) === 0 ? null : $result[0];
     }
 
+    /**
+     * @param $caseNumber
+     * @return |null
```
gregtyler: Presumably:
```suggestion
     * @return array<mixed>|null
```

PR #285, opened 2020-03-23
DDPB-3161: Show synchronisation status in admin panel

gregtyler:
## Purpose
When we integrate with the API, we will automatically send documents to Sirius so that users don’t have to. However, users need to be aware of any failures or delay in the integration. Therefore, we will show on the documents page which documents/submissions have been synchronised, which are queued, and which have failed.

Fixes DDPB-3161

## Approach
Working back-to-front:

- Updated the `GET /report-submission` API endpoint to return a new subset of submissions called "pending". These are submissions which have _at least document_ whose sync status is not `null`. "New" submissions have all documents as `null`.
  - This makes the queries a bit more complex, but the performance impact is minimal because there are very few non-archived submissions
- Updated design of report submissions page to give more context/information to document status, per design
  - Tidied up some of the existing DOM/CSS/JavaScript
- Added new "detached-details" JavaScript module which allows you to have a [details](https://design-system.service.gov.uk/components/details/)-like component without using the markup that requires them to be side-by-side
- Created unit tests for the client controller, stealing code from DDPB-3204 which will cause merge conflicts somewhere down the line
- Updated Behat tests to match new markup
  - I haven't added Behat tests for the new tab because it's hard to recreate at this point without writing straight into the database

## Learning
I created a new `class_const` function in Twig which can be used to get a constant from a class. This let's you compare constant values without having to hard code them in Twig.

```
{% if document.syncStatus == class_const('AppBundle\\Entity\\Report\\Document', 'SYNC_STATUS_QUEUED') %}
```

## Checklist
  - N/A

### Frontend
- [x] There are no deprecated CSS classes noted in the profiler

### DIFF
--- a/client/src/AppBundle/Resources/views/Admin/ReportSubmission/_document-sync-table.html.twig
+++ b/client/src/AppBundle/Resources/views/Admin/ReportSubmission/_document-sync-table.html.twig
@@ -2,15 +2,15 @@
 
 {% macro syncStatus(status, document) %}
     {% if status == 'QUEUED' %}
-        <span style="color:darkgray">Queued</span>
+        <span class="opg-text--secondary">Queued</span>
     {% elseif status == 'IN_PROGRESS' %}
-        <span style="color:darkgray">In progress</span>
+        <span class="opg-text--secondary">In progress</span>
     {% elseif status == 'SUCCESS' %}
-        <strong style="color:green">Success</strong>
+        <strong class="opg-text--success">Success</strong>
     {% elseif status == 'PERMANENT_ERROR' %}
-        <strong style="color:red">Permanent fail</strong>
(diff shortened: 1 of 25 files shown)


client/src/AppBundle/Resources/views/Admin/ReportSubmission/_document-sync-table.html.twig
```diff
@@ -0,0 +1,56 @@
+{% trans_default_domain "admin-documents" %}
+
+{% macro syncStatus(status, document) %}
+    {% set DocumentClass = 'AppBundle\\Entity\\Report\\Document' %}
+
+    {% if status == class_const(DocumentClass, 'SYNC_STATUS_QUEUED') %}
```
acsauk-mojd: Nice - will be using this going forward.

PR #288, opened 2020-03-23
DDPB-3279: Remove the function for changing user email address and roles

gregtyler:
## Purpose
As the OPG we want to ensure security on out products and remove the potential risks of users sharing information they shouldn't with others who are not suitable to view client and other deputy data 

We've had some recent concerns (and a security incident) which highlight the dangers of allowing users to freely change their account email addresses. Amongst other things, a high-authority user account can easily be given to a low-authority user by setting it to an email address they control.

Similarly, it's too easy to move a user account between realms (e.g. prof/PA/lay/admin) despite there being no good reason to do so and the realms being completely separate.

Fixes DDPB-3279

## Approach
I removed the email and role fields from all places where it's available:

- "Edit user" in the admin panel
- "Edit my details" for all users
- "Edit team member" for team admins
- "Edit organisation member" for organisation admins

Both fields are still available when creating new users through these methods (where suitable). Also, team and organisation admins can still switch users between admin and team member roles since this is wholly internal.

I removed any tests which used these fields and wrote an ADR to record this decision.

## Checklist
  - Added an ADR
  - N/A, but I've removed some which are no longer relevant

### DIFF
--- a/docs/architecture/decisions/0004-user-emails-and-roles-are-immutable.md
+++ b/docs/architecture/decisions/0004-user-emails-and-roles-are-immutable.md
@@ -1,0 +1,25 @@
+# 4. User emails and roles are immutable
+
+Date: 2020-03-23
+
+## Status
+
+Accepted
+
+## Context
+
+Due to the freedom given to users and admins in changing a user's email address or role, we've had issues identifying why people have certain permissions in DigiDeps. As well as being confusing, we believe this has led to security issues where a high-authority user has had their email changed in an attempt to reuse the account, thereby granting unreasonable access to the owner of the new email address.
+
+There are rarely good reasons to change a user's email address or role. Because very little in DigiDeps belongs directly to a user, these situtations can be resolved by deleting the original account and creating a new one with the correct permissions.
+
+## Decision
+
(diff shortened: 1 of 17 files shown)


aiesoftware:
Nice work. Always nice to remove a bunch of tests no longer required.

docs/architecture/decisions/0004-user-emails-and-roles-are-immutable.md
```diff
@@ -0,0 +1,25 @@
...
+
+## Context
+
+Due to the freedom given to users and admins in changing a user's email address or role, we've had issues identifying why people have certain permissions in DigiDeps. As well as being confusing, we believe this has led to security issues where a high-authority user has had their email changed in an attempt to reuse the account, thereby granting unreasonable access to the owner of the new email address.
+
+There are rarely good reasons to change a user's email address or role. Because very little in DigiDeps belongs directly to a user, these situtations can be resolved by deleting the original account and creating a new one with the correct permissions.
+
+## Decision
+
+User email addresses and roles will henceforth be immutable: they are set when a new user is created and cannot subsequently be changed.
```
acsauk-mojd: Epic use of henceforth 👑

PR #289, opened 2020-03-24
DDPB-3281 - Delete Lay users - remove conditions

acsauk-mojd:
## Purpose
Removes restrictions around deleting deputies with clients and reports and removes delete permissions from admins.

Fixes DDPB-3281

## Approach
I was going to leave the fallback text in so super admin users got a message saying they can't delete themselves but I didn't think this added much value so I've removed it completely.

For other Users they just won't see the button now which, as we've actively removed this permission, seemed like the right thing to do.

## Checklist


acsauk-mojd:
I've been through the scenario of a Deputy having a client with a single report and multiple reports and in both cases, the User, Client and all Reports have been deleted successfully. Am I missing any other relationships linked to any of the entities?

gregtyler:
No I think that's it, just wanted to make sure we understood what would happen since I think the cascade triggers are a bit hidden :)

PR #292, opened 2020-03-26
DDPB-3296 - Fix internal authentication and API endpoint URI

acsauk-mojd:
## Purpose
Move from auth token to client secret authentication on Document and ReportSubmission UPDATE endpoints.

Fixes DDPB-3296

## Approach
I've removed auth token security from the Documents and ReportSubmission endpoints as the callouts will never be made from a logged-in user. Instead, we're using the client secret authentication method. I also missed adding https to the Sirius API Gateway endpoint which was highlighted during some debugging with the integrations team.

## Learning
This really highlighted to me the need for an integration test that ties all of the changes together made in #273. As I'd been mocking out the responses from the API while updating entities, the auth problems were masked.

## Checklist

### DIFF
--- a/api/app/config/security.yml
+++ b/api/app/config/security.yml
@@ -96,6 +96,14 @@ security:
            pattern: ^/document/queued
            security: false
            methods: [GET]
+        updateDocument:
+            pattern: ^/document/\d+
+            security: false
+            methods: [PUT]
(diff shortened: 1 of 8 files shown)


acsauk-mojd:
Closing and merging into DDPB-3350.

PR #298, opened 2020-04-01
DDPB-3291: Mark public services or inject them

gregtyler:
## Purpose
In Symfony v4, all services are treated as private by default and cannot be retrieved with `Container::get()`. That means that services will either need to be marked as public in configuration, or dependency injected.

Dependency injection is preferable, as it means that Symfony only instantiates the classes necessary. Any public services are loaded on every request (which slows down the application).

To get ahead of this, we want to implement more dependency injection and identify services which should be public.

Fixes DDPB-3291

## Approach
I searched for usages of `Container::get()` and replaced them with dependency injection. In a few places this wasn't possible so I explicitly made the service public (e.g. `RestClient` is loaded dynamically in far too many places to reasonably rewrite right in one PR).

I also tidied up some configuration files, and more or less completely rewrote ManageController because it was leaning really heavily on dynamic service loading.

I briefly tried to upgrade to Symfony v4 to check if the changes worked, but there are too many differences to reasonably test against. The application built successfully though, which suggests the configuration is fully compatible.

## Learning
- There's way more we can do to tidy up the config files; I think a lot of it is unnecessary now
- I've mostly done away with aliases and instead use class names (e.g. `->get(AuthService::class)` rather than `->get('auth_service')`) to make dependencies more explicit (in the few instances of public services)
- A couple of v4 things I noted to be aware of:
  - In unit tests, we can't just use `Container::set()` to replace a service with a mock (because the service will be private). To allow this, we just need to add `AppBundle\Service\SomeService: { public: true }` to each service that needs to be overwritten in `config_unittest`.
  - I saw some issues with the AppKernel and container tags. I hope the changes in DDPB-3290 will help with this, but undoubtedly we'll have some follow-up work at the point of upgrade.

## Checklist
  - N/A
  - I've tried to make minimal test changes to demonstrate this work hasn't actually changed anything


client/src/AppBundle/Controller/ManageController.php
```diff
@@ -11,6 +16,30 @@
  */
 class ManageController extends AbstractController
 {
+    /** @var array<ServiceAvailabilityAbstract> */
+    private $services = [];
+
+    public function __construct(
+        ContainerInterface $container,
```
aiesoftware, line 23: Will it be problematic injecting the container in v4/5? At the least, it's won't be best practice - perhaps we can inject the `env` parameter directly?

aiesoftware:
Great job. Seems like a huge part of the upgrade has been addressed here. Just one question to address

client/src/AppBundle/Controller/ManageController.php
gregtyler: In v4/5 we'll be able to inject `ParameterBagInterface` which lets us get application parameters in a clean dependency-injecty way. But unfortunately that's not supported in v3 so I thought using the container for now would give us the easiest upgrade path.

aiesoftware: Happy with that

PR #300, opened 2020-04-01
Ddpb 3044

aiesoftware:
## Purpose
Symfony 3.4 has a security option `logout_on_user_change` which, when set, will log out a user when the user has changed*. In 3.4 the option defaults to false, which is considered a security flaw and as such, Symfony 4.0 defaults it to true and converts it to a No-Op option, effectively meaning that setting it to false will have no effect. Symfony 5.0 removes the config option altogether.

\* The user is considered to have changed when any of the properties required by the UserInterface (like the username, password or salt) changes.

## Approach
To enable a smooth upgrade, we can set the flag to `true` for the time being, giving us an automatic redirect to login out of the box. This will effect the current user journey as they will now arrive at the login page after changing their password. To ease this change, we now have contextual text on the login page that will explain why they have landed on the page.

I targeted each entry point for setting/changing a password, and at the point of redirect, I set a `login-context` var in the session to either `password-create` or `password-create`. I then read this session var on login page to determine if we need to display a contextual message. The var is removed from session on login to prevent it from reappearing.

## Learning
During the upgrade to version 5, we will likely see an error for an unrecognised option in the config, so we can simply remove the option and the code will behave as expected: logging the user out on password change.

## Checklist

### Frontend


api/src/AppBundle/Entity/User.php
```diff
@@ -124,7 +124,7 @@ class User implements UserInterface
     /**
      * @var string
      * @ORM\Column(name="password", type="string", length=100, nullable=false)
-     * @JMS\Exclude
+     * @JMS\Groups({ "user-login"})
      */
     private $password;
```
aiesoftware, line 129: Without returning the password, the `AbstractToken` on the client side could never detect that it had changed, thus preventing the automatic redirect to login from working.

PR #301, opened 2020-04-02
DDPB-3306: Fix firefox page scroll bug

acsauk-mojd:
## Purpose
Fixes a bug in admin autosize boxes where the page scrolls to the top of the box when a user types in an expanded text box.

Fixes DDPB-3306

## Approach
Went a bit crazy trying to replicate this until I decided to switch browser to Firefox where the page scrolls to the top of the textarea when autosizing is enabled.

I'm not taking any credit for the code - I stole it from here - https://stackoverflow.com/questions/18262729/how-to-stop-window-jumping-when-typing-in-autoresizing-textarea/60868351#60868351

## Checklist

### Frontend

### DIFF
--- a/client/src/AppBundle/Resources/assets/javascripts/modules/textarea-autosize.js
+++ b/client/src/AppBundle/Resources/assets/javascripts/modules/textarea-autosize.js
@@ -7,8 +7,8 @@ module.exports = function (containerSelector) {
   const textArea = $(containerSelector).find("[class*='js-auto-size'] textarea")
 
   textArea.on('keyup input paste change', function (event) {
-    let $this = $(this)
-    let initialHeight = $this.height()
+    const $this = $(this)
+    const initialHeight = $this.height()
 
     $this
       .height(initialHeight - 20)
--- a/client/src/AppBundle/Resources/assets/javascripts/modules/textarea-autosize.js
+++ b/client/src/AppBundle/Resources/assets/javascripts/modules/textarea-autosize.js
@@ -4,14 +4,14 @@
 // Note that associated styles live under _.forms.scss
 
 module.exports = function (containerSelector) {
-  var textArea = $(containerSelector).find("[class*='js-auto-size'] textarea")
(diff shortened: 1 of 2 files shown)


docker-compose.yml
```diff
@@ -70,6 +70,7 @@ services:
             - ./client/src:/var/www/src:ro
             - ./client/tests:/var/www/tests
             - ./client/phpstan.neon:/var/www/phpstan.neon
+            - ./client/web:/var/www/web
```
gregtyler, line 73: This needs to be re-removed. It causes problems with a fresh build because you generate the assets in `docker-compose build` and then overwrite them with whatever's in your local folder in `docker-compose up`. Normally this results in there being two assets folders and the application getting confused.

Certainly something to fix in the future (perhaps after the folder restructure in v4).

client/src/AppBundle/Resources/assets/javascripts/modules/textarea-autosize.js
```diff
@@ -4,14 +4,14 @@
 // Note that associated styles live under _.forms.scss
 
 module.exports = function (containerSelector) {
-  var textArea = $(containerSelector).find("[class*='js-auto-size'] textarea")
+  const textArea = $(containerSelector).find("[class*='js-auto-size'] textarea")
 
   textArea.on('keyup input paste change', function (event) {
-    var $this = $(event.target)
+    let $this = $(this)
+    let initialHeight = $this.height()
```
gregtyler, line 11: Stealing from `npm run lint`:
```suggestion
    const $this = $(this)
    const initialHeight = $this.height()
```

docker-compose.yml
acsauk-mojd: Ahhh OK - was just getting annoyed by having to sync the folders up after a yarn build.

gregtyler: Yeah, it's a massive PITA when you're doing CSS/JS work.

PR #303, opened 2020-04-03
DDPB-3267: Create court order entities

aiesoftware:
## Purpose
Pave the way for creating and using CourtOrder entities by laying the initial foundations for creating CourtOrders and the relevant touch points in the application.

Fixes DDPB-3267

## Approach
Whilst there seems a large number of file changes, most of those are housekeeping changes that enables us to push this feature forward. 

The general approach is to be able to take some input of various formats (a `CasRec` row used on registration, a CSV row used on PA/prof upload, some POSTed data from an admin fixture, etc), and normalise it into a domain entity, from which we can easily build a CourtOrder. This allows us to change our inputs and still build CourtOrders in an extensible and isolated way.

For example, this PR contains 2 entry points for creating court orders, both of which have a different set of inputs to build them from, so we create 2 different `Assemblers` that both create the normalised DTO. Here is a simplified summary:

```
// Lay Registration
$dto = new (LayToCourtOrderDtoAssembler())->assemble($casRecRow);
$courtOrder = new (CourtOrderFactory())->create($dto);

// Org Upload
$dto = new (OrgCsvToCourtOrderDtoAssembler())->assemble($csvRow);
$courtOrder = new (CourtOrderFactory())->create($dto);
```
Creating CourtOrders from Doctrine and Admin fixtures will be a case of creating new Assemblers for each and reusing the Factory.

## Learning

## Checklist

### Frontend


api/src/AppBundle/v2/Controller/CourtOrderController.php
```diff
@@ -0,0 +1,83 @@
...
+    /**
+     * @Route("", methods={"POST"})
+     * @Security("has_role('ROLE_LAY_DEPUTY')")
+     *
+     * @param Request $request
+     * @return JsonResponse
+     * @throws ORMException
+     * @throws OptimisticLockException
+     */
+    public function createAction(Request $request): JsonResponse
```
gregtyler, line 69: Rather than using `$request`, can we not set the route to `@Route("{id}", methods={"POST"})` and have the function argument as `$id`?

Not a big deal, just might look a bit neater.

gregtyler: Looking at this closer, I see it's from POST data not GET

aiesoftware: 👍

PR #305, opened 2020-04-07
DDPB-3350: Update document sync payload

gregtyler:
## Purpose
Following the decision to change from multipart/related to application/json when sending documents to Sirius we need to update our DocumentSync code and Pact code to match the new swagger doc.

Fixes DDPB-3350

## Approach
_In progress_

## Learning

## Checklist


client/src/AppBundle/Command/DocumentSyncCommand.php
```diff
@@ -69,7 +69,38 @@ private function getQueuedDocuments(): array
...
+                    'documents',
+                    'document-synchronisation',
+                    'document-storage-reference',
+                    'document-report',
+                    'report',
+                    'report-client',
+                    'client-case-number',
+                    'document-report-submission',
+                    'report-submission',
+                    'documentSync',
```
gregtyler, line 82: With the new nested groups, I think there are some duplicates in here that can be removed?
```suggestion
                    'document-report-submission',
```

client/src/AppBundle/Service/Client/Sirius/SiriusApiGatewayClient.php
```diff
@@ -91,49 +79,37 @@ public function sendReportPdfDocument(SiriusDocumentUpload $upload, string $cont
...
             'POST',
-            $multipart,
-            'application/vnd.opg-data.v1+json',
-            'multipart/form-data'
+            $reportJson,
+            'application/vnd.opg-data.v1+json'
         );
 
-        return $this->httpClient->send($signedRequest);
+        return $this->httpClient->send($signedRequest, ['debug' => true]);
```
gregtyler, line 93: Probably don't still want debug mode enabled?
```suggestion
        return $this->httpClient->send($signedRequest);
```

client/src/AppBundle/Service/DocumentSyncService.php
```diff
@@ -193,9 +201,13 @@ private function handleDocumentStatusUpdate(Document $document, string $status,
         }
 
         try {
-            return $this->restClient->put(
+            return $this->restClient->apiCall(
+                'put',
                 sprintf('document/%s', $document->getId()),
-               json_encode(['data' => $data])
+               json_encode($data),
```
gregtyler, line 207: Tiny thing:
```suggestion
                json_encode($data),
```

client/src/AppBundle/Service/DocumentSyncService.php
```diff
@@ -204,25 +216,25 @@ private function handleDocumentStatusUpdate(Document $document, string $status,
...
-            return $exception;
-        }
+        return $this->restClient->apiCall(
+            'put',
+            sprintf('report-submission/%s/update-uuid', $reportSubmissionId),
+            json_encode(['uuid' => $uuid]),
+            'raw',
+            [],
+            false
+        );
```
gregtyler, line 226: Probably not for now, but would be nice to have a shortcut method for making clientsecret API calls without having to specify all the parameters like this.

client/src/AppBundle/Command/DocumentSyncCommand.php
acsauk-mojd: Just had a quick test and I agree with all of these apart from `document-report-submission` and `report-submission` as these bring back required properties on Document.

client/src/AppBundle/Service/DocumentSyncService.php
acsauk-mojd: Mmmm yeah - would need to jig the method signature around as it would only remove the need for the last two arguments. Will keep in mind next time we're looking at rest client 👍

gregtyler:
Great stuff, happy to approve but can't because I created the PR :( Take this instead:

image

PR #307, opened 2020-04-09
DDPB-2754: Apply consistent pattern to header hierarchy throughout the application

gregtyler:
## Purpose
We need to apply a consistent pattern to header hierarchy throughout the application so that we comply with WCAG regulations, and the design of the application is easy for a user follow. 

Page titles should be unique and succinctly describe what the page is (e.g. the question being asked). In report sections they should be supported by captions noting the section title. In admin pages they should be supported by captions noting the area of the admin panel.

We need to remove unnecessary duplication of headings, labels and legends, and avoid skipping heading levels.

Fixes DDPB-2754

## Approach
I changed various translation files to use proper heading hierarchy, and updated the application template to put this on the page better.

I updated several `form_` functions to support making the label/legend the page title ([per the Design System](https://design-system.service.gov.uk/patterns/question-pages/#label-as-a-page-heading)) with captions.

I made use of the new `form_` settings on single-question pages to avoid heading duplication, and tidied up other bits of Twig/HTML where suitable.

## Learning

## Checklist
  - N/A
  - N/A

### Frontend
  - Where possible
- [x] There are no deprecated CSS classes noted in the profiler
  - I cleaned out some where suitable, but didn't rebuild the templates more than necessary

### DIFF
--- a/client/src/AppBundle/Resources/views/Report/MoneyInShort/category.html.twig
+++ b/client/src/AppBundle/Resources/views/Report/MoneyInShort/category.html.twig
@@ -7,8 +7,8 @@
 {% set transOptions = {'%client%': report.client.firstname | e } %}
 
 {% block htmlTitle %}{{ ('categoryPage.moneyIn.htmlTitle') | trans }}{% endblock %}
-{% block pageTitle %}{{ ('categoryPage.moneyIn.pageTitle') | trans }}{% endblock %}
-
+{% block pageTitle %}{{ 'form.categoriesIn.label' | trans(transOptions) }}{% endblock %}
+{% block supportTitleTop %}{{ ('categoryPage.moneyIn.supportTitle') | trans }}{% endblock %}
 
 {% block linkBack %}
     {{ macros.linkBackStep(backLink, 'back' | trans({}, 'common')) }}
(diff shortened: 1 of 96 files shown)


client/src/AppBundle/Resources/views/Report/MoneyInShort/category.html.twig
```diff
@@ -7,8 +7,8 @@
 {% set transOptions = {'%client%': report.client.firstname | e } %}
 
 {% block htmlTitle %}{{ ('categoryPage.moneyIn.htmlTitle') | trans }}{% endblock %}
-{% block pageTitle %}{{ ('categoryPage.moneyIn.pageTitle') | trans }}{% endblock %}
-
+{% block pageTitle %}{{ 'form.categoriesIn.label' | trans(transOptions) }}{% endblock %}
```
aiesoftware, line 10: Just a quick check that your intention is to keep this `pageTitle` block, as it most others it has been removed? There are a couple others below too

gregtyler: Yeah, each page should have either:
- A `pageTitle` block or
- An empty `pageHeader` block (to hide the default) and a label/legend with `isPageHeading: true`

This ensures that there is exactly one h1 on the page. The second option is for pages which are just a single question (e.g. "Do you have any fees to report?"), with the question itself being the page heading (where we previously might have just had the word "Fees")

PR #308, opened 2020-04-15
DDPB-3268: Create court order deputies when ingesting data

gregtyler:
## Purpose
Create court order deputy and address information when ingesting data (i.e. uploading CSVs)

Fixes DDPB-3268

## Approach
- Created new entities, DTOs, assembler and factory
- Created `CourtOrderCreator` which upserts court order and deputy information
- Called `CourtOrderCreator` from `OrgService` upload
- Updated and added tests

## Checklist
  - N/A


api/src/AppBundle/Entity/CourtOrderAddress.php
```diff
@@ -0,0 +1,173 @@
...
+        return $this;
+    }
+
+    public function setDeputy(CourtOrderDeputy $deputy): CourtOrderAddress
+    {
+        $this->deputy = $deputy;
+
+        return $this;
+    }
+}
```
aiesoftware, line 173: Two thoughts about this entity:
* It's a CourtOrderAddress yet it has reference to a CourtOrderDeputy, not a CourtOrder. If the reference is correct, should it be a CourtOrderDeputyAddress, else should we update the reference to a $courtOrder instead of a $deputy?

* That aside, do we want any relation at all? Besides from that relation it's just a generic Address object, which could have a many-to-many relationship with deputies/organisations if we removed that mapping to an entity?

api/src/AppBundle/v2/Controller/CourtOrderController.php
```diff
@@ -66,15 +69,25 @@ public function __construct(
...
         $registrationData = $this->casRecRepository->findOneBy(['caseNumber' => $client->getCaseNumber()]);
 
         $courtOrderDto = $this->courtOrderAssembler->assemble($registrationData);
         $courtOrder = $this->courtOrderFactory->create($courtOrderDto, $client, $client->getCurrentReport());
 
+        $deputyDto = $deputyAssembler->assemble($registrationData, $user);
+        $deputy = $deputyFactory->create($deputyDto, $courtOrder);
+
+        $deputy->setUser($user);
+
```
aiesoftware, line 90: Should we call `$this-em->persist($deputy)`?

api/src/AppBundle/Service/CourtOrderCreator.php
```diff
@@ -0,0 +1,171 @@
...
+
+    public function upsertCourtOrderDeputy(CourtOrderDeputyDto $deputyDto, CourtOrder $courtOrder, Organisation $organisation): CourtOrderDeputy
+    {
+        if ($this->courtOrderHasLayDeputy($courtOrder)) {
+            throw new \RuntimeException('Case number already used by lay deputy');
+        }
+
+        if (count($courtOrder->getDeputies()) > 1) {
+            throw new \RuntimeException('Court order has multiple organisations attached');
+        }
```
aiesoftware, line 140: Can you briefly explain this scenario please? :)

api/src/AppBundle/v2/Controller/CourtOrderController.php
gregtyler: We don't need to, since the CourtOrder has `cascade={"persist", "remove"}` on the relationship. I was a bit iffy on this though since that's not clear in situ (as you've identified!). Happy to fire another persist in if you think it'd be clearer.

api/src/AppBundle/Service/CourtOrderCreator.php
gregtyler: In the future we'll want to support a court order having multiple deputies, so we can model Sirius properly. To that end, the database relationship is one-to-many. But at the moment we don't actually support that, because CasRec doesn't provide the right data.

This line should never be hit, because our code doesn't currently add two deputies to the same order. It _could_ be hit if there was some data corruption, dodgy code, or someone messed with the database directly. This catches that "something is wrong" state and stops us making assumptions about whether the deputy should be updated or replaced.

It's basically the same as the "Can't determine if deputy has moved with client to new org" error we have for messed-up NamedDeputy records.

api/src/AppBundle/v2/Controller/CourtOrderController.php
aiesoftware: I'm happy with the cascade operation 👍

api/src/AppBundle/Service/CourtOrderCreator.php
aiesoftware: Thanks

api/src/AppBundle/Entity/CourtOrderAddress.php
gregtyler: As discussed on Meet, will turn this into `CourtOrderDeputyAddress` (and the corresponding DTO). It'll remain a one-to-many with CourtOrderDeputy to avoid collisions and data creep.

PR #309, opened 2020-04-15
DDPB-2898: Add option to close (resubmit) an unsubmitted report

aiesoftware:
## Purpose
Give case managers to "close" (resubmit) a report that has been marked as incomplete, so that the deputy can progress to the next report. As it stands, deputies need to manually resubmit on behalf on the case manager which is an unnecessary inconvenience.

Fixes DDPB-2898

## Approach
Added a separate form to the report management page, as per the design, that only renders when a report has been marked as incomplete (unsubmitted).

## Checklist

### Frontend

### DIFF
--- a/client/src/AppBundle/Controller/Admin/Client/ReportController.php
+++ b/client/src/AppBundle/Controller/Admin/Client/ReportController.php
@@ -6,6 +6,8 @@ use AppBundle\Controller\AbstractController;
 use AppBundle\Entity\Report\Checklist;
 use AppBundle\Entity\Report\Report;
 use AppBundle\Exception\ReportNotSubmittedException;
+use AppBundle\Form\Admin\CloseReportConfirmType;
+use AppBundle\Form\Admin\CloseReportType;
 use AppBundle\Form\Admin\ManageActiveReportType;
 use AppBundle\Form\Admin\ReviewChecklistType;
 use AppBundle\Form\Admin\ReportChecklistType;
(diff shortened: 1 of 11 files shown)


client/src/AppBundle/Controller/Admin/Client/ReportController.php
```diff
@@ -438,6 +452,35 @@ private function populateReportFromSession(Report $report, array $sessionData):
...
+     * @param Request $request
+     * @param $id
+     * @return array|RedirectResponse
+     * @Template("AppBundle:Admin/Client/Report:manageCloseReportConfirm.html.twig")
+     *
+     * @throws \Exception
+     */
+    public function manageCloseReportConfirmAction(Request $request, $id)
+    {
+        $form = $this->createForm(CloseReportConfirmType::class);
```
gregtyler, line 468: For a form this simple, you might consider [building it on the fly](https://symfony.com/doc/current/forms.html#creating-forms-in-controllers) rather than having a dedicated class
```suggestion
        $form = $this->createFormBuilder()
            ->add('save', SubmitType::class)
            ->getForm();
```

aiesoftware: I considered this, but decided against the inconsistency. I quite like the predictability that we currently have: if there's a form, there's a class. Admittedly it's a tad bloated but there is always a trade off - and makes it easier to add translations and validation rules if we need to.

gregtyler: Fair enough, I see how it's helpful for translations

PR #310, opened 2020-04-18
DDPB-3356: Reinstates option to edit deputy email address for super a…

aiesoftware:
## Purpose
To allow super admins (only) to update the email of address of users, from the user edit page in admin.

Fixes DDPB-3356

## Approach
Added the `email` field to the `EditUserType` form class when the operating user has role `ROLE_SUPER_ADMIN`. I created a new `admin_edit_user` JMS group and validation group to separate it from the `admin_add_user` group.

## Checklist

### Frontend


client/src/AppBundle/Resources/views/Admin/Index/editUser.html.twig
```diff
@@ -22,6 +23,11 @@
 
     {{ form_input(form.firstname,'addUserForm.firstname') }}
     {{ form_input(form.lastname,'addUserForm.lastname') }}
+
+    {% if form.email is defined and app.user.roleName == class_const(UserClass, 'ROLE_SUPER_ADMIN') %}
```
gregtyler, line 27: Is the role check necessary? In other forms I've just used `if form.email is defined` and let the form handle permissions exclusively

behat/tests/features/UserManagement/editiingUserDetails.admin.feature
```diff
@@ -0,0 +1,75 @@
+Feature: Editing Deputy details as an admin user
```
gregtyler, line 1: The filename has an extra `i` in it, should be `editingUserDetails.admin.feature`

client/src/AppBundle/Resources/views/Admin/Index/editUser.html.twig
aiesoftware: No not especially, happy to remove it.

behat/tests/features/UserManagement/editiingUserDetails.admin.feature
aiesoftware: 👀

PR #312, opened 2020-04-21
DDPB-3163: Allow users to queue a document for upload

gregtyler:
## Purpose
We need to add a way to manually put a document in that state.

In early days of the integration, this will allow us to test that it works as anticipated with a small number of clients. Going forward, it will also allow us to retry failed documents if/when we fix the reason for failure.

Fixes DDPB-3163

## Approach
I added a new "Synchronise" button alongside the current report submission options. If clicked it will use a new API endpoint to synchronise documents of all selected report submissions.

It will only synchronise null, `TEMPORARY_ERROR` and `PERMANENT_ERROR` documents. The rest will be left as-is.

## Checklist
  - N/A
  - No Behat test because the setup doesn't lend itself well to Behat (need a submitted report, with a document which has failed) and it would only duplicate unit tests


aiesoftware:
Nice work

PR #320, opened 2020-04-24
DDPB-3272: Generate application version statically, deprecate semvertag

gregtyler:
## Purpose
We want to deprecate semvertag because it requires Python 2.7. Since we don't use semver properly, we'll replace with a hash-based versioning system rather than a direct replacement.

Fixes DDPB-3272

## Approach
I removed semvertag and instead generated the version by combining the current branch and Git hash (`$CIRCLE_BRANCH-$SHORT_HASH`). This then allows us to upgrade the CircleCI executor to Python 3.

I removed the `git tag` step, since it seems pointless to tag a Git hash with a name including the hash (e.g. `master-0b92fed` would be a tag of commit 0b92fed).

Everything else works the same as before: the new version string is used as the tag in ECR, set as the environment variable `OPG_DOCKER_TAG` and displayed in the application footer.

## Checklist
- [x] I have updated documentation (Confluence/ADR/tech debt doc) where relevan
  - N/A

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -322,7 +322,6 @@ jobs:
           command: |
             export VERSION=${TF_WORKSPACE}-${CIRCLE_SHA1:0:7}
 
-            echo "$VERSION"
             echo "export VERSION=${VERSION}" >> $BASH_ENV
             echo "$VERSION" >> ~/project/VERSION
       - persist_to_workspace:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -320,8 +320,12 @@ jobs:
       - run:
           name: Set version
           command: |
+            echo $CIRCLE_SHA1
             export SHORT_HASH=${CIRCLE_SHA1:0:7} >> $BASH_ENV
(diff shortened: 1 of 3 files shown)


.circleci/config.yml
```diff
@@ -321,27 +315,22 @@ jobs:
...
+          command: ~/project/.circleci/set_env.sh >> $BASH_ENV
       - run:
-          name: Bump version
-          command: semvertag bump patch << parameters.semver_stage >> >> ~/project/VERSION
+          name: Set version
+          command: |
+            export VERSION=${TF_WORKSPACE}-${CIRCLE_SHA1:0:7}
+
+            echo "export VERSION=${VERSION}" >> $BASH_ENV
+            echo "$VERSION" >> ~/project/VERSION
```
gregtyler, line 326: A little clarification on what this is doing:
- Line 325 adds the environment variable to `$BASH_ENV` so it can be used by other steps in this job
- Line 326 persists the version to a file so it can be loaded by other jobs (through `set_env.sh`)

PR #321, opened 2020-04-24
Ddpb 3367

aiesoftware:
## Purpose
Fix for bug that was causing the extra details box to render when the incorrect radio option is selected.

Fixes DDPB-3367

## Approach
Straight forward case of adding a missing options object to the form renderer, which was causing the options and their corresponding config object to be out of sync.

Also updated the conditional logic to use the Govuk Design System

## Learning
This is a relatively simple bug to reintroduce - when adding a new radio option it's easy to miss that an extra config option needs passing to the `form_checkbox_group` in Twig. Something to be mindful of.

## Checklist

### Frontend

### DIFF
--- a/client/src/AppBundle/Resources/views/Report/Report/declaration.html.twig
+++ b/client/src/AppBundle/Resources/views/Report/Report/declaration.html.twig
@@ -64,6 +64,10 @@
 
         {{ form_checkbox(form.agree, 'agree', { 'labelClass': 'required' }) }}
 
+        {% set conditional_details %}
+            {{ form_input(form.agreedBehalfDeputyExplanation, 'agreedBehalfDeputyExplanation' )}}
+        {% endset %}
+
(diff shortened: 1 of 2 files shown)


client/src/AppBundle/Resources/views/Report/Report/declaration.html.twig
```diff
@@ -72,7 +76,8 @@
                 'items': [
                     {},
                     {},
-                    {'dataTarget': 'agreed-behalf-deputy-explanation-section' }
```
gregtyler, line 75: Now this is removed, you can take out the `<div id="agreed-behalf-deputy-explanation-section">...</div>` block below. Symfony can be a bit weird if you render the same form field twice.

aiesoftware: Good point

PR #322, opened 2020-04-24
In 174rework

jamesrwarren:
## Purpose
IN-174 on the digideps side did not work as intended as had tested commands locally and completely missed that we were running it on the golang image and not the python one so it didn't have the aws-cli tools installed! It didn't fail previously as the command only gets called on a promotion to production.

Decided to make a go script to do it as it makes the circle code a bit neater and follows the pattern for this circle container. Tested it on circleci.  
Fixes IN-174

## Approach
Create a go script, check it actually works in Circle!!!

## Learning
Always test circle changes in circle! Wasn't sure which dependency manager was being used for the vendor folder so learned about that!
## Checklist

### Frontend


.circleci/set_env_pact.sh
```diff
@@ -6,25 +6,28 @@ PACT_BROKER_USER="admin"
 PACT_BROKER_URL="pact-broker.api.opg.service.justice.gov.uk"
 API_VERSION="v1"
 
-export SECRETSTRING=$(aws sts assume-role \
---role-arn "arn:aws:iam::${ACCOUNT}:role/get-pact-secret-production" \
---role-session-name AWSCLI-Session | \
-jq -r '.Credentials.SessionToken + " " + .Credentials.SecretAccessKey + " " + .Credentials.AccessKeyId' 2>/dev/null)
+if [ $(aws --version 2>&1 | grep -c "aws-cli") -ne 0 ]
```
acsauk-mojd, line 9: Is this just checking aws or aws-cli is installed?

.circleci/set_env_pact.sh
```diff
@@ -6,25 +6,28 @@ PACT_BROKER_USER="admin"
...
-export AWS_SESSION_TOKEN=$(echo "${SECRETSTRING}" | awk -F' ' '{print $1}' 2>/dev/null)
+  #local export so they only exist in this stage
+  export AWS_ACCESS_KEY_ID=$(echo "${SECRETSTRING}" | awk -F' ' '{print $3}' 2>/dev/null)
+  export AWS_SECRET_ACCESS_KEY=$(echo "${SECRETSTRING}" | awk -F' ' '{print $2}' 2>/dev/null)
+  export AWS_SESSION_TOKEN=$(echo "${SECRETSTRING}" | awk -F' ' '{print $1}' 2>/dev/null)
 
-export PACT_BROKER_PASS=$(aws secretsmanager get-secret-value \
---secret-id pactbroker_admin \
---region eu-west-1 | jq -r '.SecretString' 2>/dev/null)
+  export PACT_BROKER_PASS=$(aws secretsmanager get-secret-value \
```
acsauk-mojd, line 21: Slightly picky but is there a reason for using PASS rather than PASSWORD here? At first glance, I thought this env var was to do with a pact test passing or not.

ecs_helper/cmd/pact_tags/main.go
```diff
@@ -0,0 +1,69 @@
...
+	"github.com/aws/aws-sdk-go/service/secretsmanager"
+)
+
+func basicAuth(username, password string) string {
+	auth := username + ":" + password
+	return base64.StdEncoding.EncodeToString([]byte(auth))
+}
+
+func getPactPassword(roletoassume string, pactsecretkey string) string {
+	mysession := session.Must(session.NewSession())
```
acsauk-mojd, line 20: Generally in Go code, multi-word var names are in camel case to aid readability (https://golang.org/doc/effective_go.html#mixed-caps). I'm not sure if we can enforce this with `gofmt` somehow but worth looking into.

ecs_helper/cmd/pact_tags/main.go
```diff
@@ -0,0 +1,69 @@
...
+	creds := stscreds.NewCredentials(mysession, roletoassume)
+	cfg := aws.Config{Credentials: creds,Region: aws.String("eu-west-1")}
+	sess := session.Must(session.NewSession(&cfg))
+	sm := secretsmanager.New(sess)
+	secretkey := pactsecretkey
```
acsauk-mojd, line 25: Looks like this is just referencing `pactsecretkey` - could we pass it straight into `GetSecretValue()` instead rather than assigning?

ecs_helper/cmd/pact_tags/main.go
```diff
@@ -0,0 +1,69 @@
...
+	pact_password, err := sm.GetSecretValue(&secretsmanager.GetSecretValueInput{SecretId: &secretkey})
+	if err != nil {
+		panic(err.Error())
+	}
+	return *pact_password.SecretString
+}
+
+func main() {
+
+	base_url := os.Getenv("PACT_BROKER_BASE_URL")
```
acsauk-mojd, line 35: As above, it's best to stick to one type of naming convention for env vars to aid readability.

ecs_helper/cmd/pact_tags/main.go
```diff
@@ -0,0 +1,69 @@
...
+	roletoassume := fmt.Sprintf("arn:aws:iam::%s:role/get-pact-secret-production", account)
+	pactsecretkey := "pactbroker_admin"
+
+	pact_password := getPactPassword(roletoassume, pactsecretkey)
+
+	url := "https://"  + base_url + "/pacticipants/" + consumer + "/versions/" + consumer_version
+	req, err := http.NewRequest("GET", url, nil)
+	req.Header.Add("Authorization","Basic " + basicAuth("admin",pact_password))
+
+	response, err := http.DefaultClient.Do(req)
```
acsauk-mojd, line 54: Not sure how critical it is but we're not handling this error - could be tricky to see whats gone wrong when debugging.

.circleci/set_env_pact.sh
jamesrwarren: if you do aws --version, then you get back a string like 'aws-cli/1.17.12 Python/2.7.16 Darwin/17.7.0 botocore/1.14.12' when it's the pip installed aws-cli. :)

.circleci/set_env_pact.sh
jamesrwarren: I don't mind picky. Changing it :)

ecs_helper/cmd/pact_tags/main.go
jamesrwarren: Yeah sorry, moved it out to a function and obviously missed this.

ecs_helper/cmd/pact_tags/main.go
jamesrwarren: gofmt doesn't do it but doing it manually :)

ecs_helper/cmd/pact_tags/main.go
```diff
@@ -1,29 +1,28 @@
...
 	mysession := session.Must(session.NewSession())
-	creds := stscreds.NewCredentials(mysession, roletoassume)
-	cfg := aws.Config{Credentials: creds,Region: aws.String("eu-west-1")}
+	creds := stscreds.NewCredentials(mysession, roleToAssume)
+	cfg := aws.Config{Credentials: creds, Region: aws.String("eu-west-1")}
 	sess := session.Must(session.NewSession(&cfg))
 	sm := secretsmanager.New(sess)
-	secretkey := pactsecretkey
-	pact_password, err := sm.GetSecretValue(&secretsmanager.GetSecretValueInput{SecretId: &secretkey})
+	pact_password, err := sm.GetSecretValue(&secretsmanager.GetSecretValueInput{SecretId: &pactSecretKey})
```
acsauk-mojd, line 25: 🐫

jamesrwarren: Oh deary me!

PR #328, opened 2020-04-27
DDPB-3176 - Run API unit tests in CircleCI

acsauk-mojd:
## Purpose
This change removes the ECS task for running API unit tests and runs it directly in CircleCI. This is in order to support test measures as part of the pipeline and in a future ticket, we will have the code coverage delta acting as a another gate for merging code.

Fixes DDPB-3176.

## Approach
This was a painfully long process of repeated attempts to debug issues arising in Circle CI. Initially, it seemed like an out of memory issue as the Docker container was exiting with a 137 code which typically indicates OOM errors. However, after reducing all memory usage to a minimum and switching to a tiny image to run the tests the issues were still present and on inspecting the logs it turned out it was a permissions issue with the Symfony var/cache directory. 

I've tried a number of different owners on the folder ranging from root, 1001, 0 and taking the UID of the user running the docker command but nothing seemed to be appropriate so in the end I just gave the folder 777 access. I was initially hesitant to do this due to security concerns but as it's not in the base API Docker image, and this is just to run tests, I'm fairly relaxed now. If anyone has any concerns then let me know in PR comments 👍 

## Learning
So many blogs and deep dives on Docker and Symfony permissions issues:

https://stackoverflow.com/questions/34949083/symfony-docker-permission-problems-for-cache-files?noredirect=1&lq=1

https://medium.com/@einenlum/testing-your-app-with-docker-compose-on-circleci-ffd4cc8e2cfa

https://github.com/symfony/cli/issues/183

https://sagikazarmark.hu/blog/containerizing-a-symfony-application/

## Checklist


aiesoftware:
Great work!

.circleci/config.yml
```diff
@@ -440,3 +465,18 @@ jobs:
...
+      file_to_copy:
+        description: Path to the file where the cov file will be copied from host machine.
+        type: string
+      copy_to_dir:
+        description: Path to the directory where the cov file will be copied to on the host machine.
+        type: string
+    steps:
+      - run:
+          command: |
+            docker-compose -f docker-compose.yml --project-name api-unit-tests exec api chmod -R 777 "<< parameters.file_to_copy >>"
```
jamesrwarren, line 481: Could probably make this more restrictive but given it's just for the tests it should be fine.

PR #329, opened 2020-04-29
DDPB-3380 - Fix timeouts when getting large number of queued documents

acsauk-mojd:
## Purpose
Fixes timeout issues and enables queueing on additional documents (post report submission).

Closes DDPB-3380, DDPB-3389 and DDPB-3388.

## Approach
During UAT we witnessed time outs when there were large amounts of documents queued. On investigation, this happened when the GET request was made to get all queued documents when there were 100+ documents with a status of QUEUED or 50+ when the total number of documents in the database was 1000+.

I initially tried setting JMS groups to request only the used properties of Document, Report, Client and Report Submission but this only allowed a small number of documents to be requested before timing out and also led to out of memory errors on the client side. Instead, me and @jamesrwarren built up a raw SQL query to get the data we needed and then created model that represented this new structure. The existing synchronisation code then had to be updated to use this new model.

Due to the way we are querying the DB for documents and associated report submissions we can get back additional rows for the same document. To ensure we get back all associated report submissions within the set LIMIT, a DENSE_RANK() is added to the SELECT statement and the results are then ordered and limited to being below the limit so we don't get have situations where additional rows are missed out in a request due to being over the limit.

This ticket also enables queuing documents submitted after a report has been sent to OPG.

## Learning
Doctrine is super useful for writing complex queries, but this comes at a memory or time cost. I wouldn't normally write raw SQL due to readability issues but for this specific use case, it seemed like the best course of action.

I also learnt some techniques to reduce memory usage when debugging the client-side out of memory errors. These included referencing the variable used in foreach loops and reducing the number of variables set in code by passing in the results of functions calls directly to other function calls. Again, this isn't something I'd usually do to keep things readable but we were left with few choices.

## Checklist

### Frontend


api/src/AppBundle/Entity/Repository/DocumentRepository.php
```diff
@@ -22,4 +25,77 @@ public function retrieveSoftDeleted()
...
         return $records;
     }
+
+    public function getQueuedDocumentsAndSetToInProgress()
+    {
+        $limit = self::ROWS_LIMIT;
+        // Using DENSE_RANK here as we get multiple rows for the same document due to multiple report submissions. This
+        // ensures any limit applied will not miss out submissions by chance
+        $queuedDocumentsQuery = "
+SELECT case_number, document_id, document_report_submission_id, is_report_pdf, filename, storage_reference, report_start_date, report_end_date, report_submit_date, report_type, ndr_id, ndr_start_date, ndr_submit_date, report_submission_id, report_submission_uuid
```
jamesrwarren, line 35: SQL looks fine but for readability I dunno if we wouldn't want to put each of the fields on a new line? It might make it quite long but maybe easier for people going forward.

acsauk-mojd: Good point - I'll drop them down on to individual lines 👍

api/src/AppBundle/Entity/Repository/DocumentRepository.php
```diff
@@ -22,4 +25,92 @@ public function retrieveSoftDeleted()
 
         return $records;
     }
+
+    public function getQueuedDocumentsAndSetToInProgress()
+    {
+        $limit = self::ROWS_LIMIT;
+
+        // Using DENSE_RANK here as we get multiple rows for the same document due to multiple report submissions. This
+        // ensures any limit applied will not miss out submissions by chance
```
aiesoftware, line 34: My knowledge here is limited, so more a question for my own curiosity, but does grouping by `document.id` accomplish the same thing? i.e discarding the multiple rows per document?

client/src/AppBundle/Command/DocumentSyncCommand.php
```diff
@@ -48,13 +57,15 @@ protected function execute(InputInterface $input, OutputInterface $output): int
...
             }
 
-            $documents = $this->getQueuedDocuments();
+            /** @var QueuedDocumentData[] $documents */
+            $documents = $this->getQueuedDocumentsData();
+
             $output->writeln(count($documents) . ' documents to upload');
 
-            foreach ($documents as $document) {
+            foreach ($documents as &$document) {
```
aiesoftware, line 65: I don't think pass by reference is needed as it's an object.

client/src/AppBundle/Service/DocumentSyncService.php
```diff
@@ -54,86 +53,81 @@ public function __construct(
...
+     * @param QueuedDocumentData $documentData
+     * @return QueuedDocumentData|\Exception|mixed|Throwable|null
      */
-    public function syncDocument(Document $document)
+    public function syncDocument(QueuedDocumentData $documentData)
     {
-        $this->handleDocumentStatusUpdate($document, Document::SYNC_STATUS_IN_PROGRESS);
-        if ($document->isReportPdf() && mimetype_from_filename($document->getFileName()) == 'application/pdf') {
-            return $this->syncReportDocument($document);
+//        $this->handleDocumentStatusUpdate($document, Document::SYNC_STATUS_IN_PROGRESS);
```
aiesoftware, line 61: Shall we remove this line?

aiesoftware:
Great stuff! Couple of questions, but overall nice job.

client/src/AppBundle/Command/DocumentSyncCommand.php
acsauk-mojd: I was getting some out of memory errors when looping over the array of `QueuedDocumentData` so I've tried to minimise anything which I thought could be impacting this. From what I understand, if you don't use a reference in a foreach loop then a copy is made of the variable which over a large loop would start to add up.

Admittedly I also reduced the number of items in the array at the same time so it could just be this that helped but is there a downside to using reference here over a copy?

client/src/AppBundle/Service/DocumentSyncService.php
acsauk-mojd: Ah yeah good spot - I'll remove.

client/src/AppBundle/Command/DocumentSyncCommand.php
aiesoftware: I think a copy is made by default for arrays and scalar types, but not for objects - as far as I know objects are passed by reference by default. That said, the reference here is harmless, just at worst unnecessary.

api/src/AppBundle/Entity/Repository/DocumentRepository.php
acsauk-mojd: I think that would work in the sense we would have everything grouped together but in this case I'm relying on the additional rows to be able to nest each submission related to a report under the `report_submissions` key.

The alternative was to make a new query to populate these after we had all the queued documents...I'm not totally against this idea as I think it could improve the readability of the query and I'm assuming the difference in performance would be minimal. I'll have a quick try and see.

client/src/AppBundle/Command/DocumentSyncCommand.php
acsauk-mojd: OK I wasn't aware of that - I'll drop the reference if it's not doing anything 👍

api/src/AppBundle/Entity/Repository/DocumentRepository.php
aiesoftware: That's fine, thanks for clarifying. I suspected it was because you were needing to hold on to the other rows.

client/src/AppBundle/Command/DocumentSyncCommand.php
aiesoftware: Hopefully the tests will reveal any obvious failures

PR #332, opened 2020-05-06
DDPB-3372: Backfill all lay deputyships into court orders

aiesoftware:
## Purpose
To create a `court_order` for each existing Lay deputyship, including associated `court_order_deputy` and `court_order_address` relationships, and to attach existing `report`s and `odr`s to thier respective order.

I also noticed that the previous CourtOrder creation work did not account for creating them when a Deputy becomes NDR enabled and therefore may bypass the Report creation step of registration (which is currently where Lay CourtOrders are created).

Fixes DDPB-3372

## Approach
Firstly, I have updated code to also create a CourtOrder when admin sets a Deputy to NDR enabled and they do not currently have a CourtOrder.

On a high level, the brief is:
* create a court_order for every client who has an active lay deputy and who either has a report OR an NDR (or both)
* create a court_order_deputy for every active lay deputy who belongs to a client for which there is a court_order for (step 1).
* Attach each court_order_deputy, report, and odr to the appropriate court_order

My initial approach was the programmatical route using Doctrine, but with some 25000 qualifying deputyships, the process quickly timed out and became unmanageable. So I reverted to raw SQL queries. The query's are commented where necessary so ought to be fairly self documenting for the purpose of the review.

The main concern was that after the migration, we can answer the following questions:

* _Are there any lay deputies attached to a court order who's client does not belong to them?_

```
select cod.user_id
from court_order_deputy cod
join court_order co on co.id = cod.court_order_id
join dd_user u on u.id = cod.user_id
where u.role_name = 'ROLE_LAY_DEPUTY'
and u.active = true
and not exists 
	(select user_id from deputy_case where user_id = cod.user_id and client_id = co.client_id)
;
```

* _Are there any reports attached to a court order that does not belong to the client on that order?_

```
select count(*)
from report r
join court_order co on co.id = r.court_order_id 
where r.client_id != co.client_id;
```

* _Are there any odrs attached to a court order that does not belong to the client on that order?_
```
select count(*)
from odr o
join court_order co on co.id = o.court_order_id
where o.client_id != co.client_id;
```

* _Does a client exist in more than one court_order?_
```
-- expect result to be zero except for any bizarre edge cases
select co.client_id
from court_order co
group by co.client_id
having count(co.id) > 1
;
```

* _How many reports can we expect to have a link to a court order, vs how many have actually ended up with on?_
```
-- expect both results to be equal
select count(distinct(r.id))
from report r
inner join deputy_case dc on dc.client_id = r.client_id
inner join dd_user u on u.id = dc.user_id
and u.role_name = 'ROLE_LAY_DEPUTY'
and u.active = true
;

select count(distinct(r.id)) 
from report r
inner join deputy_case dc on dc.client_id = r.client_id
inner join dd_user u on u.id = dc.user_id
where r.court_order_id is not null
and u.role_name = 'ROLE_LAY_DEPUTY'
and u.active = true
;
```
* _How many odrs can we expect to have a link to a court order, vs how many have actually ended up with on?_

```
-- expect both results to be equal
select count(distinct(o.id))
from odr o
inner join deputy_case dc on dc.client_id = o.client_id
inner join dd_user u on u.id = dc.user_id
and u.role_name = 'ROLE_LAY_DEPUTY'
and u.active = true
;

select count(distinct(o.id))
from odr o
inner join deputy_case dc on dc.client_id = o.client_id
inner join dd_user u on u.id = dc.user_id
where o.court_order_id is not null
and u.role_name = 'ROLE_LAY_DEPUTY'
and u.active = true
;
```

* _Does a client with n deputies have n court_orderdeputy rows attached to them?_
```
-- expect both results to be equal
select dc.client_id
from client c
inner join deputy_case dc on dc.client_id = c.id
inner join dd_user u on u.id = dc.user_id
left join report r on r.client_id = c.id and r.id = (select MAX(id) from report where client_id = c.id)
left join odr o on o.client_id = c.id
where u.role_name = 'ROLE_LAY_DEPUTY'
and u.active = true
and (r.id is not null or o.id is not null)
group by dc.client_id
having count(dc.user_id) > 1
;

select co.client_id
from court_order co
inner join court_order_deputy cod on co.id = cod.court_order_id
group by co.client_id
having count(co.id)  > 1
;
```

## Learning
This has certainly pushed my SQL to it's limits, but it's a reminder that with better SQL exposure we can make better use of it and less reliance on the ORM's.

## Checklist

### Frontend


api/app/DoctrineMigrations/Version237.php
```diff
@@ -0,0 +1,91 @@
...
+
+        $this->addSql("
+insert into court_order (client_id, type, supervision_level, order_date, case_number)
+select 
+  distinct(c.id),
+  case
+    -- if client does not have a report (only has an NDR) then type at this point is NULL
+    when r.id is null then null
+    -- otherwise assign HW for any '4' based reports, and PFA for all others
+    when r.type = '102_4'
```
jamesrwarren, line 33: dunno if it's more clear or not but you can do r.type in ('102_4', 103_4', etc)

aiesoftware: Good point, cheers

api/app/DoctrineMigrations/Version237.php
```diff
@@ -0,0 +1,91 @@
...
+  case
+    -- can be nullable as some clients may no longer be in casrec table but we still want a court_order created
+    when ca.order_date is not null then ca.order_date
+    else null
+  end,  
+  c.case_number
+from client c
+-- left join as some clients may not have a report but an odr instead
+-- a client likely has > 1 report but we only want 1 report in order to read the r.type - using MAX(due_date) as selection criteria
+left join report r on r.client_id = c.id and r.due_date = (select MAX(due_date) from report where client_id = c.id)
```
jamesrwarren, line 54: This is cool. Another way of doing this without the sub join would be using ROW_NUMBER() and partition by client_id, order by due_date. Just a bit of sql fun to look in to for another time. I'm not sure which is quicker to be honest.

aiesoftware: Oh nice

api/app/DoctrineMigrations/Version237.php
```diff
@@ -0,0 +1,91 @@
...
+;         
+       ");
+    }
+
+    public function down(Schema $schema) : void
+    {
+        // this down() migration is auto-generated, please modify it to your needs
+        $this->abortIf($this->connection->getDatabasePlatform()->getName() !== 'postgresql', 'Migration can only be executed safely on \'postgresql\'.');
+        $this->addSql("
+delete from court_order where id in
```
jamesrwarren, line 80: Just another general interest thing. Don't think it makes it much faster (though generally joins are a bit faster than in) or more clear but just for interest whilst i get to talk about sql... ;), you can do deletes and updates without the in:

 DELETE tbl1 
      FROM tbl1 
INNER JOIN tbl2 
        ON tbl1.id=tbl2.id
     WHERE tbl1.foo  = 'bar' 
       AND tbl1.date = '2013-05-06';

api/app/DoctrineMigrations/Version240.php
```diff
@@ -0,0 +1,90 @@
...
+(
+	select MAX(co.id)
+	from court_order co
+	inner join court_order_deputy cod on cod.court_order_id = co.id
+	inner join dd_user u on u.id = cod.user_id
+	where o.court_order_id is null
+	and co.client_id = o.client_id
+	and u.role_name = 'ROLE_LAY_DEPUTY'
+	and u.active = true
+	group by co.client_id
```
jamesrwarren, line 51: Not sure this group by is doing anything. Not 100% sure here as can't remember how this would work but if it's just getting the max then it will just be getting the max co.id regardless. If the group by is being applied then you're getting the max for each client_id so this would bring back multiple rows and you're setting every row in odr to that co.id

aiesoftware: Happy to remove the group by - I have tested locally and it yields the same result as when it is not there, i.e the report.court_order_id points to the correct court order in both cases. I think I put it in as the select was bringing back more than 1 result, so my first attempt to reduce it to one was to group by, which didn't work, so I used MAX and presumably left the group by in. Nonetheless it's a great spot 👍

jamesrwarren: Yep cool, might as well remove it then. Doesn't make much difference but to avoid confusion in future :)

PR #337, opened 2020-05-13
DDPB-3379: Fix to allow org deputies to delete documents from reports

aiesoftware:
## Purpose
Fix to allow org deputies to delete documents from reports

Fixes DDPB-3379

## Approach
The current `DocumentVoter` in the client was only granting permissions on its actions if the Report belongs to a Client that is attached to the Deputy. In the case of an Org deputy whose court order never existed pre the organisation migration, the Client will never be associated to the Deputy, but instead they are associated by belonging to the same Organisation. 

I have updated the Voter to run a further a check on whether the Client belongs to the same Organisation as the Deputy.

I added unit tests for the entire Voter, and also tidied up the Voter code.

## Learning

## Checklist

### Frontend

### DIFF
--- a/client/tests/phpunit/Security/DocumentVoterTest.php
+++ b/client/tests/phpunit/Security/DocumentVoterTest.php
@@ -1,0 +1,232 @@
+<?php declare(strict_types=1);
+
+namespace AppBundle\Security;
+
+use AppBundle\Entity\Client;
+use AppBundle\Entity\Organisation;
+use AppBundle\Entity\Report\Document;
+use AppBundle\Entity\Report\Report;
+use AppBundle\Entity\User;
+use Doctrine\Common\Collections\ArrayCollection;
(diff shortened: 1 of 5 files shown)


client/tests/phpunit/Security/DocumentVoterTest.php
```diff
@@ -0,0 +1,232 @@
...
+    /**
+     * @test
+     */
+    public function voteOnAttributeAllowsLayDeputiesToAddDocumentsToTheirOwnReport(): void
+    {
+        $this->token->method('getUser')->willReturn($this->deputy);
+
+        $this
+            ->ensureReportBelongsToClient()
+            ->ensureClientBelongsToDeputy()
```
acsauk-mojd, line 69: Need to remember this style of setting up test data - I find it really easy to read compared to generator functions

PR #341, opened 2020-05-21
DDPB-3406 - Expand fixtures endpoints

acsauk-mojd:
## Purpose
During the investigation for DDPB-3406 I realised other than manipulating the CSV there was no way to easily create Casrec entries or registered co-deputies. I've expanded the court order fixture creation page to have the option of creating a co-deputy and if the user should be activated or not. I also added a new endpoint /fixtures/create-casrec that lets you add sole or co-deputy entries there ready to be used int he registration flow.

Fixes DDPB-3406.

I had some issues with returning entities when POSTing to the user creation endpoint which is why I've had to make the extra callout to get the created deputies. If there's a trick to this let me know @aiesoftware!

## Checklist

### Frontend

### DIFF
--- a/api/src/AppBundle/DataFixtures/UserFixtures.php
+++ b/api/src/AppBundle/DataFixtures/UserFixtures.php
@@ -198,7 +198,7 @@ class UserFixtures extends AbstractDataFixture
 
         // Create CasRec record for lay deputies
         if ($data['deputyType'] === 'LAY') {
-            $casRec = new CasRec([
+            $casRec = ([
                 'Case' => $data['id'],
                 'Surname' => $data['id'],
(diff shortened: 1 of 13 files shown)


aiesoftware:
Good work on the fixture. Does this also address/explain the underlying issue on the ticket which is to investigate why the invitation link is not showing for deputies?

acsauk-mojd:
I couldn't replicate the flag not being set during testing using fixtures and the csv upload. My hunch is it could be a case where we initially had a sole deputy and then a co-deputy was added in casrec at a later date but I'm not sure how we'd check that unless we have some old versions of the CSV?

PR #345, opened 2020-05-27
DDPB-3384: Improve readability of DocumentSync errors

acsauk-mojd:
## Purpose
We are currently storing the full json blob error responses from the integration API and displaying this to users in the submission admin panel. While there is some detail in the blob its not always easy to see what has gone wrong. This change translates the error types into easier to read versions.

Fixes DDPB-3384

## Approach
I've used the Open API spec as the source of truth for all the error types (https://github.com/ministryofjustice/opg-data/blob/master/docs/deputy-reporting-openapi-v1.yml). I've raised the scenario where a caseref doesn't exist within Sirius with the integrations team (https://mojdt.slack.com/archives/CBKSZGH0C/p1590577117014100?thread_ts=1590484609.003800&cid=CBKSZGH0C) as it doesn't look like we have a way to identify this right now, but this shouldn't stop us from moving ahead with this as we can update the translations at a later date.

## Learning
Mocking the Symfony serializer in tests is a pain and requires lots of set up calls. Instead I've used a real instance of the translator with two of the built in encoders/normalizers.

## Checklist

### Frontend

### DIFF
--- a/client/src/AppBundle/Service/SiriusApiErrorTranslator.php
+++ b/client/src/AppBundle/Service/SiriusApiErrorTranslator.php
@@ -21,7 +21,7 @@ class SiriusApiErrorTranslator
 
     public function translateApiError(string $errorJson)
     {
-        $apiError = $this->serializeError($errorJson);
+        $apiError = $this->deserializeError($errorJson);
 
         $translations = [
             'OPGDATA-API-FORBIDDEN' => 'Credentials used for integration lack correct permissions',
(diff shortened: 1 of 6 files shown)


client/src/AppBundle/Service/SiriusApiErrorTranslator.php
```diff
@@ -0,0 +1,65 @@
...
+        } else {
+            return sprintf('%s: %s', $apiError->getCode(), $translations[$apiError->getCode()]);
+        }
+    }
+
+    /**
+     * @param string $errorJson
+     * @return SiriusApiError
+     */
+    private function serializeError(string $errorJson): SiriusApiError
```
aiesoftware, line 60: Minor, but should this be `deserializeError`?

acsauk-mojd: Good point - I'll update.

PR #348, opened 2020-05-28
DDPB-3394: Added N/A to prof deputy cost checklist

acsauk-mojd:
Co-authored-by: Alex Saunders <alex.saunders@digital.justice.gov.uk>
Co-authored-by: Gugandeep Chani <gugandeep.chani@digital.justice.gov.uk>

## Purpose
Adds N/A options to question L9.1 and L9.2 checklist questions.

Fixes DDPB-3394

## Approach
Following the behat changes we've put in place I've not added validation tests in behat and have instead added a unit test around the form.

## Learning
Testing forms is easier than expected https://symfony.com/doc/current/form/unit_testing.html.

## Checklist

### Frontend


client/tests/phpunit/Form/Admin/ReportChecklistTypeTest.php
```diff
@@ -0,0 +1,101 @@
...
+    }
+
+    public function invalidFormValuesProvider()
+    {
+        return [
+            'String too long' => ['abcd','abcd','abcd', false],
+            'Incorrect type' => [1, 2, 3, false]
+        ];
+    }
+}
```
aiesoftware, line 101: I really like bringing form testing into unit tests, although I notice the link you provide mentions to _not_ unit test the form validation:

_Don't test the validation: it is applied by a listener that is not active in the test case and it relies on validation configuration. Instead, unit test your custom constraints directly or read how to add custom extensions in the last section of this page._

So essentially we should be using this to test the mapping between form fields and the Entity, and for the presence of any custom vars that we pass to it. I think the idea is that we only test our custom validation constraints, and trust the Symfony ones - although that does not test that we have configured the validation correctly which is rather important in my opinion. Perhaps we'll have to use behat for form validation after all?

client/tests/phpunit/Form/Admin/ReportChecklistTypeTest.php
```diff
@@ -0,0 +1,101 @@
...
+            'Yes' => ['yes', false],
+            'No' => ['no', false],
+            'Null' => [null, false]
+        ];
+    }
+
+    /**
+     * @dataProvider invalidFormValuesProvider
+     */
+    public function testSubmitInvalidData($costCertificates, $reasonableCosts, $overChargedCosts, bool $deputyDetails)
```
aiesoftware, line 61: I'm not sure that this test has asserted that a validation error occurs, or perhaps I'm missing it? It seems to have the same assertions as the test above.

client/tests/phpunit/Form/Admin/ReportChecklistTypeTest.php
acsauk-mojd: Yeah I did notice that but like you said it felt wrong to not test that side of the form...I guess if we wanted to ensure the validation on the entity was configured correctly then a simple direct entity test would prove that by passing in some valid and invalid values to the setters and trust that Symfony internals are doing the right thing.

client/tests/phpunit/Form/Admin/ReportChecklistTypeTest.php
acsauk-mojd: The assertion here was checking that the expected values for the properties passed invalid values was null (e.g. the form failed to set it). Not ideal and now that you've bought it up I think it would make more sense at the entity level.

client/tests/phpunit/Form/Admin/ReportChecklistTypeTest.php
aiesoftware: I don't think that will cover the specific constraints that we configure in the annotation of each property in the entity. To do that it looks like we would need to use the `Validation` component in isolation: https://stackoverflow.com/questions/7004601/symfony2-testing-entity-validation-constraints

client/tests/phpunit/Form/Admin/ReportChecklistTypeTest.php
aiesoftware: Ah I see. Perhaps we can leave this as a workaround until we decide a better way to test our constraint annotations 👍

client/tests/phpunit/Form/Admin/ReportChecklistTypeTest.php
acsauk-mojd: Nice find - I'll see how that goes applying to this ticket and if it works then lets use this approach going forward 👍

PR #358, opened 2020-06-08
add monitoring lambda for db queries

jamesrwarren:
## Purpose

We wanted alerts from our RDS databases based on SQL queries. Initially we want to know if there have been any documents sat at QUEUED status for more than an hour.

We would like something expandable that we can easily add more checks to.

It should also be secure on the private subnets.

Fixes DDPB-3430

## Approach

I decided for simplicity in expanding the solution (also to capture the edge case where documents are stuck at queued because the poller isn't running) to go for an external lambda on the private VPC subnets that would get called by a cloudwatch event on a schedule.

I originally wanted to use golang but as it needs to be compiled, it made it challenging to work with inside the lambda for development. As such I used python as I have some familiarity with it and it works well with our existing CI pipeline (our build step uses a python container).

We can then have a single lambda per environment that could be kicked off by a multitude of cloudwatch events, each dyamically mapped to a function call. This means we can add new metrics super easily.

## Learning
All the things that caught me out:
- You need to add service endpoint to secretsmanager to get creds using boto3 when inside a VPC that doesn't have NAT outbound!
- Documentation on filters had me going off piste. It's actually quite simple but got confused between the format that logs need to be in and boto3 cloudwatch metric creation..
- psycopg2 does not work in lambda! You need to use some version that has c libs pre compiled..

Every thing else wan't too bad, just fiddly...

## Checklist

### Frontend


lambda_functions/functions/monitoring/monitoring.py
```diff
@@ -0,0 +1,83 @@
...
+import json
+
+
+def lambda_handler(event, context):
+
+    checks = {
+        "queued_documents": queued_documents,
+    }
+
+    print(event)
```
aiesoftware, line 14: Do we want this `print`, and the `print` on line 64?

aiesoftware:
Looks like a good job! Curious about the library classes being committed into repo, is this a common practice? Does it make it difficult to update those libs?

lambda_functions/functions/monitoring/monitoring.py
jamesrwarren: Python can use print to log stuff regardless of log level. This first one was left in from debugging though so good spot! The other one is legit. I could use a logger set to info to do this also but this works so not sure if there's an advantage for this use case :)

aiesoftware: Ah okay 👍

jamesrwarren:
@aiesoftware Yes for the lib classes, it's only the psycopg2 one. I don't have to put it in. I mentioned in the learning that I added it because it's from a bespoke repo. It has been precompiled with some c classes that are apparently needed for it to run in a lambda. I get the other requirements in the circleci job but it didn't work with the standard pip one in lambda for this library. I'm happy to not include it in this repo and pull in from the other repo at build time. This just seemed a bit quicker.

PR #359, opened 2020-06-08
DDPB-3349: Audit logging of discharged clients

aiesoftware:
## Purpose
Provide a foundation for all of our audit logging.
Audit log when a client is discharged (soft deleted) 
Write logs to Cloudwatch

Fixes DDPB-3349

## Approach
The process involves using a custom log handler that knows how to write logs to our specified log group on our AWS account. I initially hoped to use an [existing package on Github](https://github.com/maxbanton/cwh), but it required Symfony v5 and Monolog v2, neither of which we currently have. The package is however is actually quite simple, so I created our own custom handler and based it on the package.

**Invoking the handler**
For every `$logger->notice()` call, our handler will pick up the call, and if that message contains `['type' => 'audit']`, it will write the message to the log group. All other calls to `$logger->notice()` (or any other level eg `->warning()`) will be ignored by our handler.

The handler will not create the log group as Terraform takes care of this per environment, but it will create the log stream if it doesn't exist. The log stream to use is determined by the value of `'event'` that we pass in as part of the log message.

To verify logging messages are formatted correctly during dev, I have written a local handler that will be used instead, which writes to the regular `stdout` which can be viewed in docker logs.

**Writing event objects**
Each call to `->notice(...)` takes an array that is json encoded and sent to the log. To give us a consistent array structure, I have borrowed from the `Redux` pattern of having a centralised class that is responsible for creating and returning the "event" arrays. For each new thing we want to log, we create a new `AuditEvents->thingThatHappened($data)` method, eg:

```
$logger->notice('', $events->userDeleted(['email' => 'foo@test.com']); 
```
This makes it easy to dispatch the same log event in areas where occurrences of the event may be scattered throughout the code.  

## Learning
The main learning was in understanding how monolog uses a hierarchical based system in delegating log messages off to it's handlers in order that they are added to the `Logger`. If the current handler chooses to handle the message (usually based on the log level of the message), it can either stop propagating through the rest of the handlers, or let it continue through the chain. This is set by a `$bubble` bool on each Handler. 

We need to set the log group as an environment variable and inject into our service, however the logging service is built at build time in docker when `clear-cache` is called. This was a problem as the env vars do not exist on the container at that time. I found that we can set a default fallback value to prevent the "undefined environment variable" errors. That value is never actually used at runtime as the actual env var exists by then.

## Checklist

### Frontend


client/src/AppBundle/Service/OrgService.php
```diff
@@ -186,10 +210,18 @@ protected function processChunks($chunks)
             /** @var array $upload */
             $upload = $this->restClient->post('org/bulk-add', $compressedChunk);
 
+            if (count($upload['added']['discharged_clients']) > 0) {
```
jamesrwarren, line 213: Ah I get what you were saying the other day. Makes sense :)

client/src/AppBundle/Controller/SettingsController.php
```diff
@@ -5,6 +5,7 @@
 use AppBundle\Entity as EntityDir;
 use AppBundle\Form as FormDir;
 use AppBundle\Service\Redirector;
+use Psr\Log\LoggerInterface;
```
acsauk-mojd, line 8: Doesn't look like we're using this anywhere in the class?

client/src/AppBundle/Service/Audit/AwsAuditLogHandler.php
```diff
@@ -0,0 +1,174 @@
...
+        $entry = $this->formatEntry($entry);
+
+        if (false === $this->initialized) {
+            $this->initialize();
+        }
+
+        // send items, retry once with a fresh sequence token
+        try {
+            $this->send($entry);
+        } catch (CloudWatchLogsException $e) {
```
acsauk-mojd, line 62: Is this to account for expired tokens?

client/tests/phpunit/Service/Audit/AuditEventsTest.php
```diff
@@ -0,0 +1,43 @@
...
+    public function clientDischarged(): void
+    {
+        $expected = [
+            'event' => AuditEvents::CLIENT_DISCHARGED,
+            'trigger' => 'ADMIN_BUTTON',
+            'case_number' => '19348522',
+            'type' => 'audit'
+        ];
+
+        $actual =  (new AuditEvents())->clientDischarged('ADMIN_BUTTON', '19348522');
```
acsauk-mojd, line 21: I know its not really what this is testing, but would the event type be different here? Looks like we have a couple of different event types for discharges - maybe `CLIENT_DISCHARGED `?

client/tests/phpunit/Service/Audit/AwsAuditLogHandlerTest.php
```diff
@@ -0,0 +1,304 @@
...
+        $record = [
+            'level' => Logger::NOTICE,
+            'message' => "Client Deleted",
+            'datetime' => new \DateTime('2018-09-02 13:42:23'),
+            'context' => ['type' => 'audit']
+        ];
+
+        $this
+            ->assertLogStreamWillNotBeCreated()
+            ->assertLogWillNotBePutOnAws();
```
acsauk-mojd, line 57: I say this every time but I really need to start writing custom assertions like this, it's a pleasure to read 👍

client/tests/phpunit/Service/Audit/AuditEventsTest.php
aiesoftware: Ah yes, technically it would be 'CSV_UPLOAD', which represents the trigger/source of the event. But for the purpose of the test it could as well be 'FOO', so long as what goes in also comes out.

client/tests/phpunit/Service/Audit/AwsAuditLogHandlerTest.php
aiesoftware: It overkill on small tests, but wherever a lot of mock instantiating is required, describing it with a function seems the way to go 👍

client/src/AppBundle/Service/Audit/AwsAuditLogHandler.php
aiesoftware: This try catch was borrowed from the lib mentioned in the description, but yes, from what I can tell it allows for another attempt with a fresh sequence token.

client/src/AppBundle/Controller/SettingsController.php
aiesoftware: Ah yes forgot to remove, I'll take it out

client/src/AppBundle/Service/Audit/AuditEvents.php
```diff
@@ -0,0 +1,41 @@
...
+     */
+    public function clientDischarged(string $trigger, string $caseNumber, string $dischargedBy = null): array
+    {
+        $event = [
+            'trigger' => $trigger,
+            'case_number' => $caseNumber
+        ];
+
+        if (null !== $dischargedBy) {
+            $event['discharged_by'] = $dischargedBy;
```
mattmachell, line 24: What scenarios would the discharged_by be null? Could leave a gap in audit trail if it's not set. 

Is this to cover an automated process? In which case maybe be explicit in capturing that.

aiesoftware: Yes it's to cover an automated process, which we are looking at removing in my current ticket, so now in reality we will always set this field. I expect that ticket to go ahead of this one into master, so this PR can go on hold until I update it to not include the automated process.

client/src/AppBundle/Service/Audit/AwsAuditLogHandler.php
```diff
@@ -0,0 +1,174 @@
...
+            'logGroupName' => $this->group,
+            'logStreamName' => $this->stream,
+            'logEvents' => $entry
+        ];
+
+        if (!empty($this->sequenceToken)) {
+            $data['sequenceToken'] = $this->sequenceToken;
+        }
+
+        $response = $this->client->putLogEvents($data);
```
mattmachell, line 169: How do we handle a failed response due to connectivity?

aiesoftware: The AWS SDK implements automatic retries with exponential backoffs for increasing delay periods between retries. This particular client instance is using the default number of retries which is 3.

mattmachell: Cool thought it might be under the hood, but wasn't sure.

PR #360, opened 2020-06-09
DDPB-3438: Fix getSyncedReportSubmission logic

acsauk-mojd:
## Purpose
~Currently, when getting a synced report submission linked to a QueuedDocumentData object we choose the first submission that has a UUID. This works for single submissions but as soon as there are multiple report submissions attached to a QueuedDocumentData we won't be getting the correct submission.~

We were sending the wrong UUID when additional submissions were made. This change ensures we get the UUID that is linked to the Document.

Fixes DDPB-3438

## Approach
~I've added a check to ensure the document id of the QueuedDocumentData matches one of the document IDs of a submissions Documents property. This way we know the submission relates to the document being synchronised and will be added to the correct folder in Sirius.~

After some chats with Jim on the query we came up with a simpler solution to this problem. Rather than getting back all of the submissions and arranging them next to the `QueuedDocumentData` we change the join on ReportSubmission to be on the submission id and document. This way we get back the UUID if it exists so we can use it knowing its is definitely linked to that document or we can tell if the reportPDF hasn't been synced if its null.

I've also added some behat tests to check for what happens after the sync command is run so we can have more confidence to make changes in the future. One point to note on this - as the document-sync schedule runs every five minutes on the environment the tests would occasionally fail. I made the choice of changing the rate to be once every 24 hours for non-prod envs to greatly reduce the risk of a sync taking place during a test run.

## Checklist


aiesoftware:
Good work on the Behat tests!

api/src/AppBundle/Entity/Repository/DocumentRepository.php
```diff
@@ -43,36 +64,46 @@ public function getQueuedDocumentsAndSetToInProgress(string $limit)
...
+WHERE d.synchronisation_status = 'QUEUED'
+) AS sub WHERE dn < $limit) as sub2
+WHERE sub2.rown = 1;";
 
         $conn = $this->getEntityManager()->getConnection();
         $stmt = $conn->prepare($queuedDocumentsQuery);
         $stmt->execute();
 
-        $queuedDocumentData = [];
+        $queuedDocumentDatas = [];
```
jamesrwarren, line 101: Data is a plural I think

acsauk-mojd: You're right - I think I was trying to show that there would be multiple `QueuedDocumentData` objects in the array as opposed to some data about a queued document. Does that make sense!?

jamesrwarren: Yeah I get what you're going for ;)

PR #367, opened 2020-06-15
DDPB-3431 add alert

jamesrwarren:
## Purpose
Add the alarm and metric for notifying us of documents that have been queued for more than an hour.

Fixes DDPB-3431

## Approach
Added the alarm and added some extra testing in containers so we can expand on it in the future. Checked alarm triggers properly

## Learning
Seems the log group name for the lambda has to be the same as the name of the lambda or it uses the wrong log. Caused a bit of head scratching as to why the metric points weren't populating.

## Checklist

### Frontend


environment/alarms.tf
```diff
@@ -48,6 +48,32 @@ resource "aws_cloudwatch_metric_alarm" "nginx_errors" {
...
 }
 
+resource "aws_cloudwatch_log_metric_filter" "queued_documents" {
+  name           = "MonitorQueuedDocuments.${local.environment}"
+  pattern        = "{ $.eventType = \"Queued_Documents\" }"
+  log_group_name = aws_cloudwatch_log_group.monitoring_lambda.name
+
+  metric_transformation {
+    name      = "QueuedGreaterThanHour.${local.environment}"
+    namespace = "Test"
```
aiesoftware, line 58: Change this from "Test" or is it a genuine name?

jamesrwarren: No! Good call!

PR #369, opened 2020-06-16
Ddpb 3458

jamesrwarren:
## Purpose
The real fix for our availability woes. It seems like the constructor on this class was getting called each time the check elb was getting called meaning any calls to the healthcheck where calling out to a load of other services. This causes quit e bit of overhead and makes this healthcheck reliant on other services

200s:
image

499s:
image

Fixes DDPB-3458

## Approach
Move the checks out of the contructor and into the other healthchecks.

## Learning
NA

## Checklist

### Frontend


client/src/AppBundle/Controller/ManageController.php
```diff
@@ -17,35 +17,35 @@
  */
 class ManageController extends AbstractController
 {
-    /** @var array<ServiceAvailabilityAbstract> */
-    private $services = [];
-
     public function __construct(
+    )
+    {
+    }
```
acsauk-mojd, line 23:
```suggestion
    public function __construct(){}
```

🙊

PR #372, opened 2020-06-18
DDPB 3461: Switch off auto discharge on CSV upload

aiesoftware:
## Purpose
In order to prevent unwanted side effects, we want to remove the auto discharge functionality from prof/pa CSV upload. Discharges will require a manual verification stage in the future.

Fixes DDPB-3461

## Approach
Removed the complex discharge logic from the upload
Removed the old debugging logging
Tidied up the `OrgService` a little. 
## Learning
This class has been significantly reduced in lines of code and code complexity in recent months; we are now in a good position for a clean refactor with a much greater reduce to the risk of breaking anything.

## Checklist

### Frontend


api/src/AppBundle/Service/OrgService.php
```diff
@@ -210,6 +210,11 @@ public function addFromCasrecRows(array $data)
 
                 $client = $this->upsertClientFromCsv($row, $namedDeputy);
                 if ($client instanceof Client) {
+
+                    if (!$this->clientHasSwitchedOrganisation($client) && $this->clientHasNewNamedDeputy($client, $namedDeputy)) {
+                        $client->setNamedDeputy($namedDeputy);
+                    }
+
```
aiesoftware, line 217: Whilst ensuring that we no longer move Client's into new orgs, we also must ensure that we don't auto update their named deputy either, by only updating deputies for clients who have not moved orgs.

api/tests/AppBundle/Service/OrgServiceTest.php
```diff
@@ -369,9 +369,9 @@ public function testPAAddFromCasrecRows()
         $this->pa->addFromCasrecRows($dataMove);
         self::$em->clear();
 
-        // check client 3 is now associated with deputy1
-        $this->assertCount(1, self::$fixtures->findNamedDeputyByNumber('00000001')->getClients());
-        $this->assertCount(2, self::$fixtures->findNamedDeputyByNumber('00000002')->getClients());
+        // check client 2 remains with deputy 1
+        $this->assertCount(2, self::$fixtures->findNamedDeputyByNumber('00000001')->getClients());
+        $this->assertCount(1, self::$fixtures->findNamedDeputyByNumber('00000002')->getClients());
         $this->assertCount(1, self::$fixtures->findNamedDeputyByNumber('00000003')->getClients());
```
acsauk-mojd, line 376: Out of scope for this ticket but it would be nice to re-write this into a few individual tests as its not clear what's covered at first glance.

api/src/AppBundle/Service/OrgService.php
acsauk-mojd: Question on this - it looks like the client is always going to have the `$namedDeputy` from the CSV row at this point and it will never be null as its set in `upsertClientFromCsv()` so the second check on `clientHasNewNamedDeputy()` seems redundant? Unless I've missed something and `$namedDeputy` can be null at this point? 🤔

aiesoftware: I did mull over this too, but for an existing client they won't have hit the setter in `upsertClientFromCsv()` and we've have been stung in the past assuming that all existing clients have a named deputy set. I think technically you are correct, if we remove it we get the same result, except we will be needlessly setting it to it's current value for the majority of them, whereas this is an attempt to only do an update if we actually need to - it's arguable that by being more restrictive we somehow don't update one that is needed, but we also stand a chance of not updating one that we shouldn't. I think it is better to not update than to update as not updating does not have any side effects.

api/tests/AppBundle/Service/OrgServiceTest.php
aiesoftware: Completely agree. The `OrgService` is not in a position for a reasonably straight forward rewrite and a much needed rewrite of the tests

api/src/AppBundle/Service/OrgService.php
acsauk-mojd: Makes sense - better to not have a potential data breach 👍

PR #373, opened 2020-06-18
DDPB-3465: Fix access to fixtures link

acsauk-mojd:
My bad - the test I'd written was only checking I could see it as super admin and missed that I couldn't see it as admin.


aiesoftware:
A little picky, but can we also alter the `@Security` annotation on the Controller actions? Sorry, should have spotted first time around.

acsauk-mojd:
Not picky at all - I've added security to all the endpoints involved now.

PR #374, opened 2020-06-19
In 234

jamesrwarren:
## Purpose
As part of the solution to allow larger file sizes across to sirius. The first solution that was decided on from the output of a spike that required the least rework for sirius side and not too much rework in integrations or digideps was that digideps would send the storage ref instead of a full 64 encoded payload. 

The lambda would then pull in this payload and send it on to sirius as usual. There was a PR on the integrations side that will allow the cross account access and this is the PR on Digideps side.

Fixes IN-234

## Approach

In development allow access from any role from sirius dev to assume this role because of branch based permissions. In other environments restrict this to the proper buckets and environments.

## Learning
Can't use wildcards for the trusted principals so we had to improve the naming on the integrations side of the task roles.

## Checklist

### Frontend


shared/document_sync.tf
```diff
@@ -0,0 +1,42 @@
...
+  tags               = local.default_tags
+}
+
+data "aws_iam_policy_document" "integrations_assume_role" {
+  statement {
+    sid    = "AllowIntegrationsLambdaS3"
+    effect = "Allow"
+    principals {
+      type        = "AWS"
+      identifiers = local.account.development ? [local.sirius_root] : [local.sirius_env_lambda]
```
acsauk-mojd, line 18:
```suggestion
      identifiers = local.account.name = "development" ? [local.sirius_root] : [local.sirius_env_lambda]
```

Saves having to add a `development` prop to all accounts then?

jamesrwarren: I did try that as it would have been nicer. Unfortunately terraform doesn't like that sort of statement. It gives a syntax error.

jamesrwarren: Ok, I missed out on this. It does like it with a == and I just must have missed it!

PR #376, opened 2020-06-23
stop database from being deleted on change

jamesrwarren:
## Purpose
To stop the database from being recreated and snapshots lost by terraform on a change. We have added 3 extra flags that should have been there which should stop it happening again!!

These are 
- "database protection" to prevent destruction
- "delete_automated_backups" to false which will stop automated backup being destroyed on destruction of database
- "skip_final_snapshot" to false so that final snapshot will be taken before destruction in the case where we force a destruction.

Fixes DDPB-3469

## Approach
Change the flags in terraform.

## Learning
NA

## Checklist

### Frontend

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -11,10 +11,11 @@ workflows:
           name: build pr
           filters: { branches: { ignore: [ master ] } }
 
-      - apply:
+      - terraform-command:
           name: apply environment
           requires: [ build pr, lint terraform ]
           filters: { branches: { ignore: [ master ] } }
+          tf_command: apply
 
       - run-task:
           name: reset environment
@@ -58,7 +59,7 @@ workflows:
           requires: [ restore environment ]
           filters: { branches: { ignore: [ master ] } }
 
-      - apply:
+      - terraform-command:
           name: destroy environment
           requires: [ approve destroy environment ]
           filters: { branches: { ignore: [ master ] } }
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -77,9 +77,24 @@ workflows:
           tf_tier: shared
           tf_workspace: development
 
+      - apply:
```
acsauk-mojd, line 80: I feel like the naming for this job should change now - we're using apply for plan, apply and destroy. Struggling to think of a catchall though that snappy. Something like `run terraform command` is descriptive. I'd also drop the default for apply if you decide to change the job name and be explicit about passing it in when we want to run apply.

.circleci/config.yml
```diff
@@ -128,16 +143,31 @@ workflows:
...
+          filters: { branches: { only: [ master ] } }
+          tf_workspace: production02
+          tf_command: plan
+
+      - apply:
+          name: plan shared-production
+          requires: [ apply preproduction ]
+          filters: { branches: { only: [ master ] } }
+          tf_tier: shared
+          tf_workspace: production
```
aiesoftware, line 158: `production02`?

.circleci/config.yml
jamesrwarren: Yeah i did want to do that. I call it terraform action in my other integration pipelines but didn't want to make too many changes and risk borking the pipeline but you're right, let's go for it. Might as well do it now.

.circleci/config.yml
jamesrwarren: No shared is interestingly in production.. :)

acsauk-mojd: Yeah I never got this either 🤷

PR #379, opened 2020-06-26
Ddpb 3442

aiesoftware:
## Purpose
Creates command to manage checklist syncing by retrieving checklists from API, transforming to a normalised entity, and passing on to a service that will make the post.

Fixes DDPB-3442

## Approach
Largely a copy of the `DocumentSyncCommand`. At some point, it is very likely we will to adapt `DocumentSyncCommand` to manage but types of sync (reports and checklists), but at this point it was decided to keep them separate whilst some unknown differences may crop up in development.

## Checklist

### Frontend


aiesoftware:
I left it as a separate task, most likely part of the POST ticket, just to keep these incremental releases a little smaller. Focus here was to write the service that orchestrates everything, but without too much implementation elsewhere. Not sure if that is the right approach but it probably won't matter either way as it won't get switched on until it's all done.

acsauk-mojd:
Yeah makes sense - I'm all for smaller incremental releases.

aiesoftware:
@acsauk-mojd Please can I get a reapproval. I have added extra scaffolding code to ensure no errors when the service is invoked whilst there is no API for `documents/queued`. These empty methods have temporarily decreased unit test coverage until we implement those methods.

PR #380, opened 2020-07-01
Add Pa11y accessibility test tool to CircleCI pipeline

Gugandeep:
## Purpose
Add Pa11y Accessibility testing tool to pipeline to alert developers to any areas of the service that may cause issues to users who are using other tools to view webpages.

Fixes DDPB-3325

## Approach
Added a pa11y job into the config for CircleCI. This jobs runs the pa11y container/service. The container grabs all the dependencies needed to run the pa11y tool and brings in the config file pa11y needs. 

After the pa11y container is finished being brought up, prod mode is enabled is remove the symfony profiler from display. The job waits for a small amount of time before running the pa11y test.

During the test, pa11y will report any errors it detects. The command to run the pa11y test ensures that the job, and therefore the pipeline, does not fail when an error is detected.

## Learning
Learnt how to create and configure jobs in CircleCI through the config.yml file. 
Learnt how to create a Docker service and configure dependencies and networks.
Learnt how to install packages in a Docker container through the Dockerfile.
Learnt how to configure pa11y tool to run on multiple URLs.

## Checklist

### Frontend

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -13,7 +13,7 @@ workflows:
 
       - pa11y-ci:
           name: accessibility test
-          filters: { branches: { ignore: [ master ] } }s
+          filters: { branches: { ignore: [ master ] } }
 
       - terraform-command:
           name: apply environment
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -508,8 +508,7 @@ jobs:
             docker-compose --project-name pa11y-ci exec frontend touch /var/www/.enableProd
             docker-compose --project-name pa11y-ci exec admin touch /var/www/.enableProd
             docker-compose --project-name pa11y-ci exec api touch /var/www/.enableProd
-            docker-compose -f docker-compose.yml --project-name pa11y-ci run pa11y || echo "Pa11y found some errors"
-
+            docker-compose -f docker-compose.yml --project-name pa11y-ci run pa11y
 
   run-task:
     executor: terraform/terraform
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -508,7 +508,7 @@ jobs:
             docker-compose --project-name pa11y-ci exec frontend touch /var/www/.enableProd
             docker-compose --project-name pa11y-ci exec admin touch /var/www/.enableProd
             docker-compose --project-name pa11y-ci exec api touch /var/www/.enableProd
-            docker-compose -f docker-compose.yml --project-name pa11y-ci run pa11y try{pa11y-ci}catch(error){console.log("error")}
+            docker-compose -f docker-compose.yml --project-name pa11y-ci run pa11y || echo "Pa11y found some errors"
 
 
   run-task:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -503,7 +503,7 @@ jobs:
       - run:
           name: Run pa11y
           command: |
-            docker-compose -f docker-compose.yml up -d pa11y
+            docker-compose -f docker-compose.yml --project-name pa11y-ci up -d pa11y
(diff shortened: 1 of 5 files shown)


.circleci/config.yml
```diff
@@ -486,6 +491,26 @@ jobs:
...
+            ~/project/.circleci/set_env.sh >> $BASH_ENV
+      - run:
+          name: Run pa11y
+          command: |
+            docker-compose -f docker-compose.yml up -d pa11y
+            docker-compose exec frontend touch /var/www/.enableProdMode
+            docker-compose exec admin touch /var/www/.enableProdMode
+            docker-compose exec api touch /var/www/.enableProdMode
+            sleep 10
+            docker-compose -f docker-compose.yml run pa11y pa11y-ci || echo "Pa11y found some errors"
```
acsauk-mojd, line 512: I think it's worth adding a comment here to say what's going on with the double pipe for our future colleague's sake.

pa11y/Dockerfile
```diff
@@ -0,0 +1,31 @@
...
+# Installs latest Chromium (77) package.
+RUN apk add --no-cache \
+      chromium \
+      nss \
+      freetype \
+      freetype-dev \
+      harfbuzz \
+      ca-certificates \
+      ttf-freefont \
+      nodejs \
```
acsauk-mojd, line 12: I *think* we can remove this as we're already pulling node in from the base image - worth testing locally.

pa11y/Dockerfile
```diff
@@ -0,0 +1,31 @@
...
+      yarn
```
acsauk-mojd, line 13: We could *possibly* remove this and just use npm that comes with node - not sure if it changes how the package managing works but could save some time for the circle job to run if we only have to worry about one package manager.

.circleci/config.yml
```diff
@@ -486,6 +491,26 @@ jobs:
...
+          command: |
+            ~/project/.circleci/set_env.sh >> $BASH_ENV
+      - run:
+          name: Run pa11y
+          command: |
+            docker-compose -f docker-compose.yml up -d pa11y
+            docker-compose exec frontend touch /var/www/.enableProdMode
+            docker-compose exec admin touch /var/www/.enableProdMode
+            docker-compose exec api touch /var/www/.enableProdMode
+            sleep 10
```
mattmachell, line 511: I'd tend to avoid sleep and prefer something that waits for a service to be ready.

pa11y/Dockerfile
mattmachell: Might also be worth looking if there's an official alpine node chromium somewhere (Selenium docker builds might have that for example).

.circleci/config.yml
acsauk-mojd: Good point actually - we're using `waitforit` in Serve-OPG if you want to see how that would work @Gugandeep https://github.com/ministryofjustice/serve-opg/blob/master/.circleci/config.yml

.circleci/config.yml
Gugandeep: Added a comment to explain what the purpose of the double pipe is

pa11y/Dockerfile
Gugandeep: Removed installation of nodejs and yarn packages. Changed yarn command to npm install.

pa11y/Dockerfile
Gugandeep: I've had a look on Docker but couldn't find any official images that fit the match unfortunately.

.circleci/config.yml
Gugandeep: Added a ticket to the backlog to implement this: https://opgtransform.atlassian.net/browse/DDPB-3518

PR #381, opened 2020-07-02
Ddpb 3489

jamesrwarren:
## Purpose

Have a non pinned version of RDS that has to be applied to master initially (if this strategy seems to work well, we can expand this to other versions).Created a weekly job that runs after the apply which will tell us if everything is ok.

The preproduction release will use the RDS version from master and the prod version will use rds version from preprod. 

This should be the current version rather that the version that is to be applied in the maintenance window. 

I have also shoehorned a couple of other minor tweaks into this PR to save brand new ones:

- Minor fix to the monitoring logic as we should be checking the created_on of the report_submission rather than the documents. It seems it's quite regular for docs to be created more than an hour before submission.

- Pinning fargate platform to version 1.3 until we try the 1.4 upgrade.

Fixes DDPB-3489

## Approach
Changes to the terraform to create the logic for the pinning.

## Learning
Fixed versions still only apply in the maintenance window.

One annoying thing to consider with this approach! During the time between where it updates the minor version in TF code and the maintenance window, it's going to say there's a change to apply against the DB for that time. I tried to get round this by adding a tag and using that to update on and ignoring engine version as I thought it would just ignore it as a change but still update it if it saw another change to the resource but it didn't seem to work... So it won't affect anything but it might confuse people a bit.

## Checklist

### Frontend

### DIFF
--- a/environment/terraform.tfvars.json
+++ b/environment/terraform.tfvars.json
@@ -19,7 +19,8 @@
       "sirius_api_account": "649098267436",
       "state_source": "production",
       "elasticache_count": 2,
-      "always_on": true
+      "always_on": true,
+      "copy_version_from": "preproduction"
     },
     "preproduction": {
       "account_id": "454262938596",
@@ -38,7 +39,8 @@
       "sirius_api_account": "492687888235",
       "state_source": "preproduction",
       "elasticache_count": 2,
-      "always_on": true
+      "always_on": true,
+      "copy_version_from": "master"
     },
     "training": {
       "account_id": "515688267891",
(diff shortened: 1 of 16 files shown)


environment/terraform.tfvars.json
```diff
@@ -96,7 +100,8 @@
       "sirius_api_account": "288342028542",
       "state_source": "development",
       "elasticache_count": 1,
-      "always_on": false
+      "always_on": false,
+      "copy_version_from": "NA"
```
acsauk-mojd, line 104: Could we have a slash here? I read it as North America 😂

PR #385, opened 2020-07-09
DDPB-3344: Add audit logging for removal of a deputy from a client

acsauk-mojd:
## Purpose
This change adds further fields to the existing `clientDischarged` audit event to give more context and is required to support some future events.

Fixes DDPB-3344

## Approach
After a chat with @aiesoftware, we decided to wrap the Logger in its own service to smooth over testing. Previously when trying to mock out the controller flow the mocking was getting out of control having to go right down the AWS logging stack. This way we only have to mock out the single notice method.

## Checklist

### Frontend


client/app/config/services.yml
```diff
@@ -103,3 +103,11 @@ services:
             $storage: '@AppBundle\Service\File\Storage\S3Storage'
             $siriusApiGatewayClient: '@AppBundle\Service\Client\Sirius\SiriusApiGatewayClient'
             $restClient: '@AppBundle\Service\Client\RestClient'
+
+    DateTime:
+        class: 'DateTime'
```
acsauk-mojd, line 108: This was really weird...I couldn't get Symfony to start without defining DateTime as a service to inject into the fake clock class. Does anyone know if there's any way round this?

aiesoftware: You're on the right lines with the date provider so you can test it, but you shouldn't need the `FakeClock`. I've used a similar service in api called `DateTimeProvider` which is a service that wraps `new \DateTime`. I inject that into the whatever needs to generate a date, and to test it I just mock that service and tell it to return a hard coded date in the tests.

acsauk-mojd: That's much simpler - I've dropped all the clock classes and switched to using the DateTimeProvider instead 👍

aiesoftware:
Great job on the Controller tests

PR #392, opened 2020-07-20
DDPB-3507 upgrade fargate platform

jamesrwarren:
## Purpose
Upgrade fargate to the current latest version. It's currently pinned at 1.3.0. Version 1.4.0 has some breaking changes in.

Fixes DDPB-3507

## Approach
The approach was to try it out with a straight upgrade and see if it worked. It didn't and many of the services didn't come up then follow the upgrade path and make it work. 

## Learning
Logs showed issues with var which I assumed was covering the fact that we hadn't done the upgrade path... (assumed...)

As such I followed this upgrade path: https://docs.aws.amazon.com/AmazonECS/latest/developerguide/platform_versions.html

Added all the relevant permissions and new endpoints it says are required. Still didn't work. Realised finally that maybe it was just handling errors differently and changed the ga env var that is never used to a fixed value. 

Everything worked.. then spent ages unpicking what I had done and checking it still worked till I realised that the whole upgrade path was redundant for us.

EDIT: Then realised that even though it keeps running without the upgrade path, it doesn't start up without it. Then had everything running nicely but stabiliser wouldn't complete. Spent absolutely ages debugging stabiliser going thourhg and printing lines from the ecs library.. It was because of the manual bits I'd done. 

Tidied everything up and pushed again and it all worked. Masterclass in how not to do a ticket!!

## Checklist

### Frontend


client/docker/confd/templates/parameters.yml.tmpl
```diff
@@ -20,7 +20,7 @@ parameters:
     log_path: /var/log/app/application.log
     ga:
         default: {{ getv "/ga/default" }}
-        gds: {{ getv "/ga/gds" }}
+        gds: ''
```
acsauk-mojd, line 23: Should this still be a blank string?

jamesrwarren: It doesn't get used anywhere and it was causing the container to fall over under 1.4.0 for some reason. I tried just updating the variable it pulls from env file to `""` in the hope it would find it but it didn't and got fed up so blanked it out. Really I should find out why it's not picking it up when it's picking up `ga/default`. I'll do it. I know you're right! :)

PR #393, opened 2020-07-20
DDPB-3310: Invalid ARIA attributes in organisation report overview accordion

acsauk-mojd:
## Purpose
Fix an issue with invalid aria values in an accordion that was identified as part of the DAC report.

Fixes DDPB-3310

## Approach
I had a search around the codebase and couldn't find anywhere else we were appending an EPOCH time string to an accordion ID. From looking at the git history I think this may have come about when we were using a custom solution rather than the GDS version. The date string was originally added there to ensure the IDs were unique but the GDS component automatically appends a number based on how many sections of the accordion there are. The date string was probably left in after we moved to the GDS version.

## Checklist

### Frontend

### DIFF
--- a/pa11y/.pa11yci
+++ b/pa11y/.pa11yci
@@ -8,7 +8,7 @@
          ]
       },
       "hideElements":"svg",
-      "timeout":60000,
+      "timeout":40000,
       "wait":5000
    },
    "urls":[
@@ -24,6 +24,19 @@
       {
          "url": "https://digideps.local",
          "screenCapture": "screenshots/index.png"
+      },
+      {
+         "url": "https://digideps.local/report/{id}/overview",
+         "actions": [
+            "navigate to https://digideps.local/login",
+            "set field #login_email to behat-prof1@publicguardian.gov.uk",
+            "set field #login_password to Abcd1234",
+            "click element #login_login",
+            "wait for path to be /org/",
+            "wait for element .behat-link-pa-report-open to be visible",
+            "click element .behat-link-pa-report-open"
+         ],
+         "screenCapture": "screenshots/report-overview.png"
       }
    ]
 }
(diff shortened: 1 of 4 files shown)


pa11y/.pa11yci
```diff
@@ -24,6 +24,19 @@
       {
          "url": "https://digideps.local",
          "screenCapture": "screenshots/index.png"
+      },
+      {
+         "url": "https://digideps.local/report/{id}/overview",
```
Gugandeep, line 29: I believe the {id} placeholder won't work and you are required to be the actual id of the report in.

pa11y/.pa11yci
```diff
@@ -24,6 +24,19 @@
...
+         "actions": [
+            "navigate to https://digideps.local/login",
+            "set field #login_email to behat-prof1@publicguardian.gov.uk",
+            "set field #login_password to Abcd1234",
+            "click element #login_login",
+            "wait for path to be /org/",
+            "wait for element .behat-link-pa-report-open to be visible",
```
Gugandeep, line 36: `            "wait for element .behat-link-pa-report-open to be visible",`

Is the id prefix supposed to be a `#` instead of a `.`?

pa11y/.pa11yci
acsauk-mojd: You're right but unfortunately, we're never going to know the exact id of a report to view when this runs in the pipeline. I thought it was better to have a good description of the page we're testing and then manually navigate to the login page/further pages in the pa11y actions.

pa11y/.pa11yci
acsauk-mojd: This link in question doesn't have an id attached to it and it felt a bit wrong to add an id just for the sake of a test so I settled on selecting the class of the link. It doesn't matter if there are multiple links on the page as the test isn't linked to a specific report so it seemed like an OK solution.

Gugandeep: Seems there was an issue after looking at the accessibility test in Circle.
I think pa11y is waiting for the URL to be `https://digideps.local/report/{id}/overview` and then it will carry out the actions. 

Maybe setting the URL to `https://digideps.local/` will avoid that error.

acsauk-mojd: Good spot! Weird though, its working OK locally...I'll SSH into the container and see what the screenshot looks like.

PR #399, opened 2020-07-23
Force first main build

acsauk-mojd:
Forcing first main build after rebuilding webhooks.


README.md
```diff
@@ -20,6 +20,7 @@ You must have Docker installed.
 ```sh
 docker-compose run --rm api sh scripts/reset_db_structure.sh
 docker-compose run --rm api sh scripts/reset_db_fixtures.sh
+
```
jamesrwarren, line 23: Think this should be a squirrel

acsauk-mojd: Opportunity missed 😭  next time.

PR #400, opened 2020-07-24
Remove temp fix for getting main branch in pipeline now main is default branch

acsauk-mojd:
I'd previously had to add a manual checkout of `main` while we still had `master` as our default branch. Now we've switched over the temporary fix can be removed.


acsauk-mojd:
Closing in favour of #401

PR #411, opened 2020-07-31
DDPB-3312: Accessibility, update address fields

acsauk-mojd:
## Purpose
The address form when adding/editing contacts has multiple address line fields, however only the first has a label.

Each field should have a label, though we can use .govuk-visually-hidden to hide them for sighted users. Hence the form would look something like:

Address [Line 1 of 3]: _____
[Address Line 2 of 3]: _____
[Address Line 3 of 3]: _____

There is a pattern for this in the GDS manual.

This is probably the case for other address-editing forms. To ensure we’re fixing this consistently, we should create a new reusable macro (e.g. form_address) for entering addresses and use it everywhere an address is added or edited.

See page 19 in DAC report

## Additional tweaks

- Upgraded missing one time services to fargate 1.4.0 to make sure they're ok when it gets switched over.
- Changed monitor script to be 2 hours as that seems to stop alerts because of time difference in how SQL and AWS are setup.

Fixes DDPB-3312

## Checklist

### Frontend

### DIFF
--- a/client/src/AppBundle/Entity/Report/Contact.php
+++ b/client/src/AppBundle/Entity/Report/Contact.php
@@ -49,14 +49,6 @@ class Contact
     /**
      * @JMS\Type("string")
      * @JMS\Groups({"contact"})
-     *
-     * @Assert\Length( max=200, maxMessage="contact.address.maxMessage")
-     */
-    private $address3;
-
-    /**
-     * @JMS\Type("string")
-     * @JMS\Groups({"contact"})
      * @Assert\Length( max=200, maxMessage="contact.address.maxMessage")
      */
     private $county;
--- a/client/src/AppBundle/Entity/Report/Contact.php
+++ b/client/src/AppBundle/Entity/Report/Contact.php
@@ -145,7 +145,6 @@ class Contact
         return $this;
     }
 
-
     public function getCounty()
     {
         return $this->county;
--- a/client/src/AppBundle/Entity/Report/Contact.php
+++ b/client/src/AppBundle/Entity/Report/Contact.php
@@ -49,6 +49,14 @@ class Contact
     /**
      * @JMS\Type("string")
      * @JMS\Groups({"contact"})
+     *
+     * @Assert\Length( max=200, maxMessage="contact.address.maxMessage")
+     */
+    private $address3;
+
+    /**
+     * @JMS\Type("string")
+     * @JMS\Groups({"contact"})
      * @Assert\Length( max=200, maxMessage="contact.address.maxMessage")
      */
     private $county;
@@ -144,6 +152,7 @@ class Contact
 
         return $this;
     }
+
 
     public function getCounty()
     {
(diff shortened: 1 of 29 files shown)


client/src/AppBundle/Resources/views/Report/Decision/start.html.twig
```diff
@@ -19,21 +19,21 @@
...
+            <span class="govuk-details__summary-text">{{ (page ~ '.whatisMentalCapacitySummary') | trans }}</span>
         </summary>
-        <div class="text opg-indented-block">
+        <div class="govuk-details__text">
             <p>{{ (page ~ '.whatisMentalCapacityPara1') | trans(transOptions) }}</p>
             <p>{{ (page ~ '.whatisMentalCapacityPara2') | trans(transOptions)|raw }}</p>
         </div>
     </details>
 
-    <details {% if '104' in report.type %}class="push--bottom"{% endif %}>
```
Gugandeep, line 32: I noticed this if statement has been removed and I just wanted to check if this was intentional.
I've noticed the `push--bottom` class was removed from other places, I'm assuming it was there to improve the look of the page and is no longer needed?

acsauk-mojd: Yeah - I *think* what was happening was the custom classes applied required a bit of margin to give the page some spacing but using the built-in GDS classes everything displays nicely with no spacing issues.

client/src/AppBundle/Resources/views/Macros/macros.html.twig
```diff
@@ -450,3 +450,49 @@ C13.7,13.2,13.2,13.7,12.5,13.7z M12.5,0.5c-6.6,0-12,5.4-12,12s5.4,12,12,12s12-5.
         </div>
     </dl>
 {% endmacro %}
+
+{# Shows the start date, end date and status of a report #}
```
acsauk-mojd, line 454: Think this has come from a different macro?

client/src/AppBundle/Entity/Report/Contact.php
```diff
@@ -145,6 +145,7 @@ public function setAddress2($address2)
         return $this;
     }
 
+
```
acsauk-mojd, line 148: This can go I think?

client/src/AppBundle/Resources/views/Ndr/Asset/Property/step.html.twig
```diff
@@ -5,6 +5,7 @@
 {% set translationDomain = "ndr-assets" %}
 {% trans_default_domain translationDomain %}
 {% set transOptions = {'%client%': ndr.client.firstname | e } %}
+{% set preFix = "form.property." %}
```
acsauk-mojd, line 8: Nice way to handle the different prefixes 👍

jamesrwarren:
> Could of tiny changes but looks good to go. You'll be taking on all the frontend tickets now, yes? 😁
> 
> (I cant approve this as I opened the PR)

Haha, we'll see about that. Will keep picking up small ones to get my skills up on it for sure :)

client/src/AppBundle/Resources/views/Org/ClientProfile/addContact.html.twig
```diff
@@ -5,6 +5,7 @@
 {% set translationDomain = "client-contacts" %}
 {% trans_default_domain translationDomain %}
 {% set page = "addContact" %}
+{% set preFix = "form." %}
```
acsauk-mojd, line 8: Just one tiny comment - this may be more self-explanatory if it was something like `translation-prefix`

PR #412, opened 2020-07-31
DDPB-3536: Limit adding Lay deputies from admin frontend to super admins

acsauk-mojd:
## Purpose
We've tracked down the co-deputy flag bug to instances where a Lay Deputy is created in the admin panel using the Add User form. This flow doesn't carry out a casrec verification step and so the co-deputy flag is never set. We should limit this feature to super admins only to ensure we don't have any further cases pop up.

Fixes DDPB-3536

## Learning
There are a few behat tests that rely on this form to populate data as part of existing older tests. I contemplated using the ensure users exist behat function we have now but it wasn't clear if some of the users created with specific email addresses were used in further tests so this seemed like a good compromise.

## Checklist

### Frontend

### DIFF
--- a/behat/tests/features/admin/05-ndr-enablement.feature
+++ b/behat/tests/features/admin/05-ndr-enablement.feature
@@ -1,7 +1,7 @@
 Feature: Enabling and disabling NDR for Lay deputies
 
   Scenario: Enabling and disabling NDR from admin toggles the availability of the NDR and Report accordingly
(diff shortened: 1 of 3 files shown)


acsauk-mojd:
Closing in order to narrow down NDR creation routes - https://mojdt.slack.com/archives/C7YH87C3A/p1596440005018000

PR #419, opened 2020-08-05
DDPB-3508: Change Document sync payload to send S3 reference

acsauk-mojd:
## Purpose
To support sending files larger than 5MB we've had to switch to sending the S3 resource reference rather than the Base64 encoded file contents. This change updates the DocumentSync flow to use the reference when POSTing to Sirius.

Fixes DDPB-3508

## Approach
In order to support manually testing the ticket, I've had to add some resource replication settings to branch environment S3 buckets. This copies any files uploaded to branch buckets to a development bucket that is referenced in the Sirius integration code to grab the document from S3. I also took the opportunity to add some more logging while syncing documents and to better handle errors in unexpected formats.

## Checklist


api/app/DoctrineMigrations/Version243.php
```diff
@@ -0,0 +1,43 @@
...
+    }
+
+    public function up(Schema $schema) : void
+    {
+        // this up() migration is auto-generated, please modify it to your needs
+        $this->abortIf($this->connection->getDatabasePlatform()->getName() !== 'postgresql', 'Migration can only be executed safely on \'postgresql\'.');
+
+        $this->addSql('ALTER TABLE document ALTER synchronisation_error TYPE TEXT');
+        $this->addSql('ALTER TABLE document ALTER synchronisation_error DROP DEFAULT');
+        $this->addSql('ALTER TABLE document ALTER synchronisation_error TYPE TEXT');
```
aiesoftware, line 27: Is this a deliberate duplicate line? Same below for checklist

client/src/AppBundle/Service/Client/Sirius/SiriusApiGatewayClient.php
```diff
@@ -100,7 +103,8 @@ public function sendSupportingDocument(SiriusDocumentUpload $upload, string $sub
     {
         $reportJson = $this->serializer->serialize(['supporting_document' => ['data' => $upload]], 'json');
 
-        $this->logger->warning("Syncing supporting document ID with UUID: $submissionUuid");
+        $this->logger->warning("Syncing supporting document with UUID: $submissionUuid");
+        $this->logger->warning($reportJson);
```
aiesoftware, line 107: Just checking if any of these logs are leftover debug?

aiesoftware:
A couple of quick checks but looks good

api/app/DoctrineMigrations/Version243.php
acsauk-mojd: This was the default that Symfony generated when I made the entity change...I'll have a play locally and see if there's any difference with only applying this once 👍

client/src/AppBundle/Service/Client/Sirius/SiriusApiGatewayClient.php
acsauk-mojd: N -, I've found it really useful to be able to see the UUIDs when syncing a document along with what's actually being sent. It also has the plus side of showing what's in each payload so if we get any unexpected errors we can go to integrations with what was sent.

PR #421, opened 2020-08-05
DDPB-3532 create development environment

jamesrwarren:
## Purpose
We need a stable environment which can be used by interested parties to access the application. It should be publicly available but not indexed by search engines and it should be clear it's not a production environment to users.

- Also added minor fix to cancel redundant builds script

Fixes DDPB-3532

## Approach
Created a new environment that gets refreshed as part of merge to main. Created a banner to make it obvious that its not prod env and added robots.txt in the web folder to make it non searchable.

## Learning
Was a bit of a faff getting the robots.txt to only be included on the client image and to get the banner to appear only on this image. 

## Checklist

### Frontend

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -24,22 +24,6 @@ workflows:
           task_name: reset_database
           timeout: 180
 
-      - run-task:
-          name: reset development
-          requires: [ lint terraform ]
-          filters: { branches: { ignore: [ main ] } }
-          task_name: reset_database
-          tf_workspace: development
-          timeout: 180
-
-      - run-task:
-          name: integration test development
-          requires: [ reset development ]
-          filters: { branches: { ignore: [ main ] } }
-          task_name: integration_test
-          tf_workspace: development
-          timeout: 3000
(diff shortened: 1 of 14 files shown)


jamesrwarren:
> Looks good - couple of questions though:
> 
> * Is having a refresh on merge to main a good idea? If we're using this for UR testing and testing the Document/Checklist flow on branches it could cause test failures of a users data to be wiped mid session.
> * Should we put some kind expiry date on the resources in the S3 bucket to stop it getting filled up? I was thinking maybe a week or two from creation time

For the first one I think you saw it when I had done the temp hack for stacey to get some good data in for here. I've put it back to how it's supposed to be now.

On your second point, we don't clear down any of our buckets in integrations/pre/training. Maybe we should have a separate story to clear down the buckets for all those envs

acsauk-mojd:
Makes sense - I'm adding 14-day expiry limits to `integration`, `preproduction` and possibly `training` (waiting for confirmation on if they need docs there for a long period of time).

PR #427, opened 2020-08-19
DDPB-3532 see if works better as always_on false

jamesrwarren:
## Purpose
It was brought to our attention that the development environment was running slowly. There didn't seem to be much difference in traffic except that the aws availability check was being hit repeatedly. By changing this single option to false we bypass this check. It's also probably better to be on aurora as the env is not used that often.

Fixes DDPB-3532

## Checklist

### Frontend


acsauk-mojd (approved):
:shipit:

PR #430, opened 2020-08-21
DDPB-3322 - Error summaries do not add "Error: " to page title

Gugandeep:
## Purpose
If form validation produces an error, the <title> tag should be prepended with "Error: " so screen readers read it out as soon as possible.

Fixes DDPB-####

## Approach
Updated `moj_template.html.twig` to add "Error:" to title when there are form errors present. 

## Checklist

### Frontend

### DIFF
--- a/client/src/AppBundle/Resources/views/Layouts/moj_template.html.twig
+++ b/client/src/AppBundle/Resources/views/Layouts/moj_template.html.twig
@@ -5,10 +5,7 @@
     <meta http-equiv="content-type" content="text/html; charset=UTF-8" />
 
     <title>
-        {%  if form is defined %}
-            {% if not form.vars.valid %}Error: {% endif %}
-        {% endif %}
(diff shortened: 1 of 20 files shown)


client/src/AppBundle/Resources/views/Layouts/moj_template.html.twig
```diff
@@ -4,7 +4,12 @@
     <meta charset="utf-8" />
     <meta http-equiv="content-type" content="text/html; charset=UTF-8" />
 
-    <title>{% if not form.vars.valid %}Error: {% endif %}{% block htmlTitle %}Deputy report{% endblock %}</title>
+    <title>
+        {%  if form is defined %}
```
coderste, line 8: Believe this could be condensed into
```
<title>{%  if form is defined and not form.vars.valid %}Error: {% endif %}...</title>
```

coderste (commented):
LGTM 👍🏻

coderste (approved):

PR #431, opened 2020-08-25
DDPB-3317: removed the <aside> HTML tag as it should not be contained within a landmark

coderste:
## Purpose
Improved usability for screenreader users

The <aside> HTML5 tag is considered a landmark tag in and by it's self.
The HTML specification suggestes landmarks should not be contained
within other landmarks and our sidebars are contained within the
<main> tag which is considered a landmark therefore we change the
<aside> tag to a standard <div> tag

Fixes DDPB-3317

## Checklist
- [x] I have updated documentation (Confluence/ADR/tech debt doc) where relevant **N/A**
- [x] I have added tests to prove my work **N/A**

### Frontend
- [x] There are no deprecated CSS classes noted in the profiler **N/A**
- [x] Translations are used and the profiler doesn't identify any missing **N/A**
- [x] Any links or buttons added are screen reader friendly and contextually complete **N/A**

### DIFF
--- a/client/src/AppBundle/Resources/views/Index/index.html.twig
+++ b/client/src/AppBundle/Resources/views/Index/index.html.twig
@@ -108,8 +108,8 @@
             </div>
 
         </div>
-        <div class="column-third">
-            <aside class="opg-related-items" role="complementary">
+        <div class="govuk-grid-column-one-third">
+            <div class="opg-related-items">
                 <h2 id="related-items-title" class="govuk-heading-m">{{ 'relatedLinks.heading' |trans|raw }}</h2>
                 <nav role="navigation" aria-labelledby="related-items-title">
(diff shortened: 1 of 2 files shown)


client/src/AppBundle/Resources/views/Index/index.html.twig
```diff
@@ -108,8 +108,8 @@
             </div>
 
         </div>
-        <div class="column-third">
-            <aside class="opg-related-items" role="complementary">
+        <div class="govuk-grid-column-one-third">
```
coderste, line 111: Changed the class name for this to be in-line with the govuk CSS naming system

Gugandeep (commented):
Looks good!

Gugandeep (approved):

jamesrwarren (approved):

PR #432, opened 2020-08-25
Ddpb 3508b

jamesrwarren:
## Purpose
This change reinstates the file upload limit to 15MB for supporting documents and moves to sending the S3 reference rather than the base64 encoded file contents.

Fixes DDPB-3508, DDPB-3511

## Approach
We needed to be able to test syncing documents on a branch environment so had to enable replication from the branch bucket to a static `development` environment that exists which the Sirius API Gateway points towards when pulling down files.

## Checklist


acsauk-mojd (approved):
:shipit:

PR #433, opened 2020-08-26
DDPB-3324 - Selected boxes are unclear to screenreaders in report preview

Gugandeep:
## Purpose
In the generated report PDF we use “X” in a box to indicate the user's selection:

This is read out to screenreader users as “cross”, which is unclear. This tickets adds the attribute aria-label=”Selected” so that screenreaders read it as “Selected” instead.

Fixes DDPB-3324

## Approach
Add the attribute aria-label=”Selected” so that screenreaders read it as “Selected” instead.

## Learning

## Checklist

### Frontend

### DIFF
--- a/client/src/AppBundle/Resources/views/Macros/macros-review.html.twig
+++ b/client/src/AppBundle/Resources/views/Macros/macros-review.html.twig
@@ -91,9 +91,9 @@
 
         <table class="checkboxes labelvalue inline">
             <tr>
-                <td class="value checkbox">{% if visitsCare.doesClientReceivePaidCare == 'yes' %}X{% else %}&nbsp;{% endif %}</td>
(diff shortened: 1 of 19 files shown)


client/src/AppBundle/Resources/views/Macros/macros-review.html.twig
```diff
@@ -191,9 +191,9 @@
 
             <table class="checkboxes labelvalue inline">
                 <tr>
-                    <td class="value checkbox">{% if param.answer == 'yes' %}X{% else %}&nbsp;{% endif %}</td>
+                    <td class="value checkbox"{% if param.answer == 'yes' %} aria-label=”Selected”>X{% else %}>&nbsp;{% endif %}</td>
```
mattmachell, line 194: I'd check how this is read by AT as the aria-label on td can be problematic (see http://www.davidmacd.com/blog/does-aria-label-override-static-text.html). On a wider note, I'm not convinced by the table usage here, feels like it's solely for layout benefit.

coderste (commented):
LGTM 👍🏻

coderste (approved):

PR #434, opened 2020-08-27
DDPB-3628 fix double click when adding documents to a report

jamesrwarren:
## Purpose
When adding documents to a submitted report, we can hit the button multiple times and this causes additional document submission rows to be entered in the DB.

Fixes DDPB-3628

## Approach
Quick fix to only allow us to submit once as once the button is pressed once, it becomes inactive.

## Learning
Learned a bit about how the data-modules work with CSS.

## Checklist

### Frontend

### DIFF
--- a/client/src/AppBundle/Resources/assets/javascripts/main.js
+++ b/client/src/AppBundle/Resources/assets/javascripts/main.js
@@ -101,7 +101,6 @@ $(document).ready(function () {
   if ($submitButtons !== null) {
     $submitButtons.forEach(function ($el) {
       $el.addEventListener('click', function ($e) {
-        this.classList.add('opg-submit-link--disabled')
         this.classList.add('opg-submit-link--disabled', 'govuk-button--disabled')
       })
     })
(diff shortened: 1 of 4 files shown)


client/src/AppBundle/Resources/assets/javascripts/main.js
```diff
@@ -95,6 +95,16 @@ $(document).ready(function () {
       new ButtonToggler().init($el)
     })
   }
+
+  const $submitButtons = document.querySelectorAll('[data-module="opg-toggleable-submit"]')
+
+  if ($submitButtons !== null) {
+    $submitButtons.forEach(function ($el) {
+      $el.addEventListener('click', function ($e) {
+        this.classList.add('opg-submit-link--disabled')
```
coderste, line 104: Maybe add `this.classList.add('govuk-button--disabled')` as well as the `opg-submit-link--disabled` styling so when the button is pressed once it's also greyed out 👍🏻

coderste (dismissed):

coderste (approved):

PR #435, opened 2020-09-01
DDPB-3630 fix sec group rule for mock end point

jamesrwarren:
## Purpose

Security group rules were missing access from frontend to mock sirius! As part of this fix, I've added a test that would pick this issue up in future (as it was still returning 200 even though the sirius check was timing out).

Also needed to update composer as had a security vulnerability come up that stopped the build.

Fixes DDPB-3630

## Approach
Fixed sec group rule and added a test

## Learning
NA

## Checklist

### Frontend

### DIFF
--- a/behat/tests/features/admin/00-prechecks.feature
+++ b/behat/tests/features/admin/00-prechecks.feature
@@ -1,0 +1,8 @@
+Feature: admin / pre checks
+
+  @infra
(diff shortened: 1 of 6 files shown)


Gugandeep (approved):

Gugandeep (approved):
Looks good to me

PR #438, opened 2020-09-03
DDPB-3321: Error summaries do not auto-focus

coderste:
There was an issue where if we had a form error on the page we didn't
auto-focus onto the error summary. According to the GOV.uk design system
we should auto-focus onto any error summaries on the page to

Fixes DDPB-3321

## Checklist
- [ ] I have updated documentation (Confluence/ADR/tech debt doc) where relevant **N/A**
- [ ] I have added tests to prove my work **N/A**

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -53,6 +53,9 @@ up-app-integration-tests: ## Brings the app up using test env vars (see test.env
 
 	docker-compose -f docker-compose.yml -f docker-compose.dev.yml up -d
 
+down-app: ### Tears down the app
+	docker-compose down -v --remove-orphans
+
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -53,6 +53,9 @@ up-app-integration-tests: ## Brings the app up using test env vars (see test.env
 
 	docker-compose -f docker-compose.yml -f docker-compose.dev.yml up -d
 
+down-app: ### Tears down the app
```
coderste, line 56: Added this extra command to make it easier to bring the app fully down

Gugandeep (approved):

PR #439, opened 2020-09-04
DDPB-3636 adding extra memory to scan task in pre and prod

jamesrwarren:
Scan service is falling over. This should at least alleviate the issue for now. Need to do thorough testing on whether we can make a the settings help minimise this instead.


coderste (approved):
LGTM 👍🏻

Gugandeep (approved):

PR #440, opened 2020-09-05
Ddpb 3636

jamesrwarren:
## Purpose
After seeing some timeouts on availability which were tracked down to clamav becoming unavailable, we have seen a gone on a mad goose chase because of misinterpretation of the logs and metrics. It can run out of memory occasionally when doing a restart but the actual issue is just that the restart can happen during the day. It always seems to recover fine so having the restart at 1am only will help 

Fixes DDPB-3636

## Approach
Change cron job to only run at 1 and put the memory back down to what it was before

## Learning
Learned a lot of useful information about putting traps and extra container logging inside containers. Was completely useless for this but hopefully it will prove useful at some point in the future...

## Checklist

### Frontend

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -34,10 +34,6 @@ workflows:
           requires: [ apply environment ]
           filters: { branches: { ignore: [ main ] } }
 
-#      - pa11y-ci:
-#          name: accessibility test
-#          requires: [ apply environment ]
-#          filters: { branches: { ignore: [ main ] } }
       - approve:
           name: approval for further testing
           type: approval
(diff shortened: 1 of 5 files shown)


.circleci/config.yml
```diff
@@ -34,28 +34,26 @@ workflows:
...
           requires: [ reset environment ]
           filters: { branches: { ignore: [ master ] } }
 
       - run-task:
-          name: integration test
+          name: reset environment for testing
           requires: [ approval for further testing ]
+          filters: { branches: { ignore: [ main ] } }
+          task_name: reset_database
+          timeout: 180
```
acsauk-mojd, line 48: 🙌

acsauk-mojd (approved):
:shipit:

PR #441, opened 2020-09-08
DDPB-3316: New window links don't identify themselves sufficiently

Gugandeep:
## Purpose
From the DACs report is what found that links that open up new windows/tabs weren't identifying themselves sufficiently. The original fix was to ensure that links that open in a new window indicate they do so (for example by adding the text "(opens in new window)”).

However, after looking at the [GOV.UK Design Systems guidance on this](https://design-system.service.gov.uk/styles/typography/#links), we decided to instead to remove the new tab/window functionality from these links.

Fixes DDPB-3316

## Approach
Remove `target="_blank"` from any link element.

## Learning

## Checklist

### Frontend


acsauk-mojd (commented):
All looks good. One Q though - the guidance says its acceptable to use new tab links when its to preserve data in a form. Are there any instances where we lose data by clicking on links?

Gugandeep:
> All looks good. One Q though - the guidance says its acceptable to use new tab links when its to preserve data in a form. Are there any instances where we lose data by clicking on links?

I've had a look through to see what pages these links pop up on. The only one that is on a form would be the login page, and it does not keep any data on the form if you were to open the link and then go back to the page. It's not too bad as it's only two fields you have to re-enter information for. It might be nice to be able to retain information on the login page, but there might also be security concerns too.

acsauk-mojd (approved):
If its only the login page then I think that's fine - good job 🥳

PR #442, opened 2020-09-08
DDPB-3315: Illogical heading structure

coderste:
Heading that are of the same semantic level should use the same HTML
heading level tag. For example all `<h2>` but when a sub-heading is added
it should increment the level by one `<h3>`. This ticket does that while
also removing `<h4>` as they should rarely be needed.

Fixes DDPB-3315

## Checklist
- [ ] I have updated documentation (Confluence/ADR/tech debt doc) where relevant **N/A**
- [ ] I have added tests to prove my work **N/A**


jamesrwarren (approved):
Seems reasonable to me :)

PR #446, opened 2020-09-14
DDPB-3451: Temporary errors should re-queue Documents

acsauk-mojd:
## Purpose
When theres a temporary error during the document sync process we are setting the error status as a permanent error. This leads to manual intervention in the admin panel, but we should handle this automatically (up to a point). 

Fixes DDPB-3451.

## Approach
This change treats any 5XX error responses from Sirius API as a temporary in nature which will lead to the status being set back to QUEUED three times. On the fourth attempt if the service is still not available then we will treat this a permanent error to ensure the document queue doesn't get filled up.

## Learning

## Checklist

### Frontend


Gugandeep (approved):
Looks good!

coderste (approved):

Gugandeep (commented):
Looks good!

Gugandeep (approved):
Approved

PR #447, opened 2020-09-16
Reinstate separate app volumes

acsauk-mojd:
## Purpose
By having a single volume for /app there is some kind of overwriting going on for the value of the client secret app env in admin that determines which roles are allowed to sign in to which side of the frontend. I've reinstated them to how they were before so this should fix local docker login issues for admin.

## Approach

### DIFF
--- a/docker-compose.override.yml
+++ b/docker-compose.override.yml
@@ -5,7 +5,11 @@ version: '3.2'
 services:
   admin:
     volumes:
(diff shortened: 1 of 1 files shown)


Gugandeep (approved):
Looks good to me!

coderste (approved):
LGTM 👍🏻

PR #450, opened 2020-09-18
DDPB-3627 exclude missing sections from checklist pdf

jamesrwarren:
## Purpose
Checklist question l7.2 is an only applicable on certain report types and only shows up to check on the checklist where it exists on those reports. Whilst some users found this confusing, it's how it was designed to work. 

However to avoid confusion it shouldn't then appear on the PDF when it doesn't exist on the checklist. 

Checked all the checklist options with different report types and this was the only instance of this I could find. Full sections were already handled properly. As such the end fix for this is a very straight forward fix

Fixes DDPB-3627

## Approach
Only displays the balance on spreadsheet where that section exists in report

## Learning
Certainly learned a bit more about how checklists worked. Easy fix but interesting area for business logic

## Checklist

### Frontend


acsauk-mojd (approved):
:shipit:

PR #451, opened 2020-09-18
DDPB-3459 update to latest version of wkhtmltopdf

jamesrwarren:
## Purpose
We were getting security alerts on this container. Stop it from having all the high and medium vulnerabilities at least.

Fixes DDPB-3459

## Approach
Used anchore to check for vulnerabilities locally and tried to find more secure base image and most recent version of the PDF generator.

Also updated the ECR script to only notify on medium, high or critical vulnerabilities. We can still see the results (and made formatting better) in the job when we want to tackle low vulnerability issues.

## Learning
NA

## Checklist

### Frontend


acsauk-mojd (approved):
:shipit:

acsauk-mojd (approved):
:shipit:

PR #453, opened 2020-09-18
Lock replication bucket ARN to development provider

acsauk-mojd:
## Purpose
The pipeline was still broken following #449 as the ARN is only available in the development account. This change locks the data source for the bucket to using the development provider to ensure the ARN is always available but replication is disabled in non-branch environments.


jamesrwarren (commented):
Certainly a bit simpler than count mayhem ;)

jamesrwarren (approved):

PR #458, opened 2020-09-21
DDPB-3659 fix to s3 read integration role

jamesrwarren:
## Purpose
The naming was wrong on s3 read integration role. Make it so that it works properly in prod and preprod.

Fixes DDPB-3659

## Approach
Make it work
## Checklist

### Frontend


acsauk-mojd (approved):
✅

PR #459, opened 2020-09-21
DDPB-3591: Drop court order entities and tables

acsauk-mojd:
## Purpose
We're going to recreate how we represent CourtOrders in our entities so this change removes the existing models and adds a migration to drop the tables and references.

Fixes DDPB-3591

## Approach
Originally the ticket has said to set references to court orders to null but as the entity and table will no longer exist this won't be possible.

## Checklist


Gugandeep (approved):
Looks good to me

PR #460, opened 2020-09-21
DDPB-3644 scale down existing infra to manage costs

jamesrwarren:
## Purpose
Scale down infra as we have way too much unused CPU and RAM. In fact previously RAM was set to 256 max for PHP so I have upped that to be inline with the container size.

Integration tests should use the app far more than we would use it for UAT and they all pass with normal times. 

I haven't scaled down prod as I don't deem it to be worth it. In fact for the scan containers I've given them extra RAM again as they seemed happier with the extra memory.

Fixes DDPB-3644

## Approach
Make changes in terraform and to the PHP ini file.

## Checklist

### Frontend

### DIFF
--- a/api/docker/confd/templates/pool.conf.tmpl
+++ b/api/docker/confd/templates/pool.conf.tmpl
@@ -530,7 +530,7 @@ access.log = /dev/null
 
 php_flag[display_errors] = off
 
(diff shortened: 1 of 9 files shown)


acsauk-mojd (approved):
✅

PR #461, opened 2020-09-22
DDPB-3659 rework for prod

jamesrwarren:
## Purpose
Completely missed the naming of prod bucket in last ticket. This will apply permissions to correct name.


acsauk-mojd (approved):
:shipit:

PR #475, opened 2020-10-06
DDPB-3651 backup our backups

jamesrwarren:
## Purpose
Have backups stored in a separate account so in case of account related disaster, we don't lose all our data.

Fixes DDPB-3651

## Approach

I looked into using s3 initially but the export to s3 option in RDS snapshots seemed geared around making the data available for querying (via athena etc). I had to do a bit of reading but it seemed others were going for the option of having the snapshots stored in a separate account as is.

To do this the process is:
- Create KMS key (shared with user from backup account)
- Copy the system snapshots out as manual snapshots using this key
- Share the snapshot with backup account
- From backup account, copy the snapshot to 'manual snapshots'
- From source account, remove the manual snapshot (don't remove the system snapshots of course!!!)

## Learning
Lots of permission based issues. I initially tried to use lambda and orchestrate with step function (as I had started using them in migration project). It became too complicated and I backtracked.. still some good L&D...

Went with ECS task scheduled by cloudwatch event in the end. Not the most elegant but an old favourite...

*INFO* reviewers may wonder about using the module. In tf 0.13.x we can now put counts on modules. I ummed and arred about whether to put this in shared or environment. I settled on environment as it gives separation of kms keys and had more of the resources we needed to import to the module.

## Checklist

### Frontend


dr_backup/dr_backup.py
```diff
@@ -0,0 +1,304 @@
...
+        print(f"Snapshot {id} exists, deleting...")
+
+        if cluster:
+            rds_client.delete_db_cluster_snapshot(
+                DBClusterSnapshotIdentifier=id
+            )
+        else:
+            rds_client.delete_db_snapshot(
+                DBSnapshotIdentifier=id
+            )
```
acsauk-mojd, line 28: Could be my non-existent python skills but could these blocks be merged with the ones above if all we're checking `cluster = True`?

dr_backup/dr_backup.py
```diff
@@ -0,0 +1,304 @@
...
+    copy_individual_snapshot(
+        backup_client,
+        shared_snapshot_identifier,
+        backup_snapshot_identifier,
+        shared_kms_key_id,
+        aws_region
+    )
+
+    # Delete shared manual backup
+    delete_snapshot(client, target_snapshot_identifier)
```
acsauk-mojd, line 293: Not sure I've followed all of the above but is there any way we could get the wrong snapshot here?

environment/disaster_recovery/ecs.tf
```diff
@@ -0,0 +1,148 @@
...
+  }
+}
+
+module "dr_backup_security_group" {
+  source = "../security_group"
+  rules  = local.dr_backup_sg_rules
+  name   = "dr-backup"
+  tags   = var.default_tags
+  vpc_id = var.aws_vpc_id
+}
```
acsauk-mojd, line 148: Don't know if this is possible in modules but we've tended to follow the pattern of extracting SGs into their own file for ease of update.

acsauk-mojd (commented):
Nice job - a couple of comments on some of the python stuff but feel free to ignore as I have super limited python experience.

dr_backup/dr_backup.py
jamesrwarren: Delete snapshot only deletes from manual snapshot section so not the automated ones and it has a different identifier. It should be fine.

environment/disaster_recovery/ecs.tf
jamesrwarren: Yeah put in separate file. Good point

dr_backup/dr_backup.py
jamesrwarren: Nah just didn't want to do the print twice but yeah saves a couple of lines of code so done it

environment/disaster_recovery/kms.tf
```diff
@@ -6,21 +6,21 @@ resource "aws_kms_key" "db_backup" {
...
+  //      identifiers = [
+  //        aws_iam_role.cross_acc_backup.arn
+  //      ]
+  //    }
+  //    resources = ["*"]
+  //    actions = [
+  //      "kms:CreateGrant",
+  //      "kms:DescribeKey"
+  //    ]
+  //  }
```
acsauk-mojd, line 23: Should we just drop this if we're not using it anymore?

environment/disaster_recovery/ecs.tf
```diff
@@ -73,14 +73,14 @@ resource "aws_iam_role" "dr_backup" {
...
-    ]
-  }
+  //  statement {
+  //    sid       = "allowAssumeAccess"
+  //    effect    = "Allow"
+  //    resources = [aws_iam_role.cross_acc_backup.arn]
+  //    actions = [
+  //      "sts:AssumeRole"
+  //    ]
+  //  }
```
acsauk-mojd, line 83: Should we drop?

environment/disaster_recovery/ecs.tf
```diff
@@ -46,10 +46,10 @@ locals {
...
         value = var.account.account_id
       },
-      {
-        name  = "BACKUP_ACCOUNT_ROLE",
-        value = aws_iam_role.cross_acc_backup.arn
-      },
+      //      {
+      //        name  = "BACKUP_ACCOUNT_ROLE",
+      //        value = aws_iam_role.cross_acc_backup.arn
+      //      },
```
acsauk-mojd, line 52: Drop?

acsauk-mojd (commented):
Makes sense - but should we just get rid of the stuff you've moved to org-infra?

acsauk-mojd (approved):
:shipit:

PR #480, opened 2020-10-08
DDPB-3717: Configure Cloudwatch logs in localstack

acsauk-mojd:
## Purpose
Ever since we added audit logging the behat tests broke part way through running the test suite locally with an error showing the app was calling out to the real cloudwatch endpoint. This change configures Cloudwatch logs to run in localstack so the behat tests (and the app running in prod mode) doesn't break when trying to send audit logs to Cloudwatch.

Fixes DDPB-3717

## Approach

## Learning
Either due to our docker setup or the way localstack works we're still relying on a PHP script that runs after localstack is up (we were already using this for some SSM setup). Its not pretty but gets the job done for now. Potentially upgrading localstack could fix this but that's outside the scope fo this ticket.

## Checklist


jamesrwarren (approved):
Awesome!

PR #484, opened 2020-10-12
DDPB-3670: Move and reimplement one audit log from client to API

acsauk-mojd:
## Purpose
As part of Symfony upgrade preparations, we've decided to move the audit logging from the client-side to the API side in order to move away from overwriting services in the DI container during tests. This approach also allows us to write the business logic (and tests) for certain entity update events once and then forget about it rather than having to manually implement it in each place in the app that updates the entity.

Fixes DDPB-3670

## Approach
I've hooked into the Doctrine lifecycle events rather than adding this to an entity class or as a service. This means that any time a user entity email address is updated in any part of the app we will have a log created.

The following subtask under the main logging and emails ticket will move the remainder of the log events over.

## Learning
I'd forgotten that when dealing with the services.yaml and config files a cache reset is required to recreate the new service container in order to reflect any changes.

## Checklist


jamesrwarren (commented):
Looks good. Will let the devs have a look over it though before approving

Gugandeep (commented):
Looks good!

I see you've implemented wait for it which is great. Gives us a head start on the following ticket: [DDPB-3518](https://opgtransform.atlassian.net/browse/DDPB-3518)

api/docker/confd/templates/parameters.yml.tmpl
```diff
@@ -3,6 +3,18 @@
 
 # This file is a "template" of what your parameters.yml file should look like
 parameters:
+    cloudwatch_logs_client_params:
+       version: 'latest'
+       region: 'eu-west-1'
+{{ if exists "/cloudwatch/logs/localstack" }}
```
coderste, line 9: Could you explain this part for me please? When would this exist and when wouldn't it?

api/src/AppBundle/Service/Audit/AuditEvents.php
```diff
@@ -0,0 +1,163 @@
...
+            'changed_on' => $this->dateTimeProvider->getDateTime()->format(DateTime::ATOM),
+            'changed_by' => $changedBy,
+            'subject_full_name' => $subjectFullName,
+            'subject_role' => $subjectRole,
+        ];
+
+        return $event + $this->baseEvent(AuditEvents::EVENT_USER_EMAIL_CHANGED);
+    }
+
+    public function clientEmailChanged(
```
coderste, line 84: No PHPDoc above this function

coderste: Also do we know the type this function returns? If so should we typehint it?

api/src/AppBundle/Service/Audit/AwsAuditLogHandler.php
```diff
@@ -0,0 +1,174 @@
...
+    public function __construct(CloudWatchLogsClient $client, $group, $level = Logger::NOTICE, $bubble = true)
+    {
+        $this->client = $client;
+        $this->group = $group;
+
+        parent::__construct($level, $bubble);
+    }
+
+    /**
+     * {@inheritdoc}
```
coderste, line 44: Like this :D didn't even know it was a thing 👍🏻

coderste (changes requested):
Just the PHPDocs missing on the one function other than that this is great work 👏🏻 👍🏻

api/docker/confd/templates/parameters.yml.tmpl
acsauk: So we use confd to handle some of the config values here in different environments. I'm not a huge fan of it as the syntax isn't very clear with what its doing and Symfony has its own environment system that we can hook into with things like `env.test` and `config.test` etc but that's a different discussion 😂 

This is basically saying if there's an environment variable called CLOUDWATCH_LOGS_LOCALSTACK then compile the values down. We only have this variable in environments where we want localstack to run so its excluded any places that use real infra. The confd process itself runs after [composer is run](https://github.com/ministryofjustice/opg-digideps/blob/main/client/composer.json#L38) as a post hook and is something we'll need to update when we do the Symfony upgrade as the package that handles that won't work in Symfony 4+ seeing as they use [Symfony Flex](https://symfony.com/doc/current/setup/flex.html) instead.

api/src/AppBundle/Service/Audit/AuditEvents.php
acsauk: Good spot - actioned 👍

api/src/AppBundle/Service/Audit/AwsAuditLogHandler.php
acsauk: Yeah - one thing in PHP that encourages us to type less 🙃

api/docker/confd/templates/parameters.yml.tmpl
coderste: Ah! okay that makes more sense to me now 👍🏻

coderste (approved):
LGTM 👍🏻

Gugandeep (commented):
Latest changes looks good!

PR #496, opened 2020-10-19
DDPB-3242 - Migrate to Aurora

TomTucka:


environment/aurora/main.tf
```diff
@@ -0,0 +1,88 @@
...
+
+  lifecycle {
+    ignore_changes  = [replication_source_identifier]
+    prevent_destroy = true
+  }
+}
+
+resource "aws_rds_cluster_instance" "cluster_instances" {
+  count                           = var.aurora_serverless ? 0 : var.instance_count
+  auto_minor_version_upgrade      = var.auto_minor_version_upgrade
```
acsauk, line 33: I can see we default this to false but does the terraform workflow update that var at all?

acsauk: Probably phrased that a bit weird...I guess what I really need to know is if the var value is static or dynamic (similar to tfvars etc).

TomTucka: It's static, It will only change if `auto_minor_version_upgrade` is passed into the module call. I would leave it static while we migrate then it should be a simple change once we've migrated.

sixdaysandy (approved):

PR #497, opened 2020-10-19
DDPB-3184 create enc to test adding encryption to s3

jamesrwarren:
## Purpose

We currently encrypt data on s3 using a policy. To show that that the buckets use encryption we can set it at the bucket level. AES256 encryption which the policy specifies is the same as the aws kms key uses so they should be compatibly and we can leave in the policy as it was previously working.

Fixes DDPB-3184

## Approach
Got a bit nervous about how this would affect existing data so spun it up without the encryption, ran the end to end tests then manually ran the encryption on the buckets through terraform and checked I could still access the existing files and that I could upload new files through the app.

Checked the integrations still work fine in environment.

image

## Checklist

### Frontend

### DIFF
--- a/environment/s3.tf
+++ b/environment/s3.tf
@@ -33,6 +33,14 @@ resource "aws_s3_bucket" "pa_uploads" {
     }
   }
 
+  server_side_encryption_configuration {
+    rule {
+      apply_server_side_encryption_by_default {
+        sse_algorithm = "aws:kms"
+      }
+    }
+  }
+
   replication_configuration {
     role = aws_iam_role.replication.arn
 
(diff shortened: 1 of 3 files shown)


environment/s3.tf
```diff
@@ -31,6 +31,14 @@ resource "aws_s3_bucket" "pa_uploads" {
     }
   }
 
+  server_side_encryption_configuration {
+    rule {
+      apply_server_side_encryption_by_default {
+        sse_algorithm = "aws:kms"
```
sixdaysandy, line 37: Don't you want this to be AES256 to match the bucket policy, same for the other one?

jamesrwarren: so under the hood aws kms seems to use aes 256 and i thought it needed a bespoke key but yeah you're right. seems it doesn't any more and actually that seems to pass all tests fine so yes that's a very good point.

sixdaysandy (approved):

PR #498, opened 2020-10-20
DDPB-3726: Make general feedback email sending testable 

acsauk:
## Purpose
This ticket was originally around moving the email sending functionality from `client` to `api` in order to hook into entity lifecycle events to unblock testing and remove DI container setting in `client` unit tests. After playing around with some of the emails we're sending it didn't feel right to move this to `api` as some of the personalisation details we send to Notify are available in the client but not required for the API CRUD calls on entities. This would have meant sending extra data to the API purely to support this approach so instead, I revised how we're testing email sending in controllers and came up with an alternative approach.

Fixes DDPB-3726

## Approach
In order to support testing email sending in controllers, I've overridden the `NotifyClient` and `RestClient` classes when running the app under the test environment. This unlocked being able to collect emails sent and assert on them as well as mock out any calls to our internal api.

When the next ticket in this series is played we'll move the other emails being sent in controllers to follow this approach and once all completed we can remove `config_unittest.yml` and `AbstractControllertestCase` which are there to support DI setting.

One big change that was required in order to get this ticket working was to overhaul what we've been referring to as 'Prod mode' (using either `dd-mode-prod` or `make prod-mode`). Under the hood, this was actually switching the app between `dev` mode with debug enabled and `test` mode with debug disabled. This led to me losing about a days worth of work trying to work out why the tests were breaking when in 'prod mode' when actually the app was in test mode 🤦 . In the end, I've renamed the functions from `dev-mode` and `prod-mode` to `debug-enabled` and `debug-disable` with the app now always running in dev mode but the profiler and toolbar being either enabled or disabled respectively.

## Learning
I'd originally tried to create a mock for `GuzzleClient` in `RestClient` in order to implement the [officially supported guzzle mocking](https://docs.guzzlephp.org/en/stable/testing.html) but due to the way we structure `RestClient` I couldn't get this working. I still think there's a lot of value in this approach for other tests so we may want to look at ripping out as much as we can from `RestClient` and seeing if we can implement a standardised client configuration package - something like [EightPointGuzzle](https://packagist.org/packages/eightpoints/guzzle-bundle) has worked well in the past.

I also did a revisit to what I know about the[ app mode system](https://symfony.com/doc/3.4/configuration/environments.html) that Symfony uses in order to work out the weirdness I was seeing when changing between what we were previously calling prod and dev mode.

## Checklist

### Frontend


Gugandeep (dismissed):
Looks good to me

coderste (commented):
LGTM 👍🏻

coderste (approved):

Gugandeep (approved):

PR #501, opened 2020-10-22
DDPB-3564 - Secret Management

sixdaysandy:
## Purpose
Bring secret management (creation/deletion) under the control of terraform.

Fixes DDPB-3564

## Approach
As there are three different approaches to secrets, I've chosen to unify them by having each environment have it's own set of secrets generated at the account/shared level.  The downside to this is that secrets must be created at the environment level before a new persistent environment can be created.
I've taken the approach of passing a list of environments per workspace into a module with a list of secrets that are required, I feel this is the cleanest and easiest to manage approach.

## Checklist

### Frontend

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -310,8 +310,8 @@ orbs:
               password: $DOCKER_ACCESS_TOKEN
         resource_class: small
         environment:
-          TF_VERSION: 0.13.3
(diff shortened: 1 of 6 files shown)


TomTucka (approved):

TomTucka (approved):

acsauk (approved):
:shipit:

PR #502, opened 2020-10-22
Push release time to OPG Metrics

acsauk:
## Purpose
Adding a step to push digideps release times to OPG metrics. This will give us an overview of when releases are pushed out to production. This won't be totally accurate as there will be a short delay between apply production and the metrics job running but its POC for now and we can look at exporting the actual release time to a variable and sharing it with the metrics job in the future.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -211,6 +211,10 @@ workflows:
           tf_workspace: production02
           tf_command: apply
 
+      - add-to-metrics:
+          name: add to metrics
+          requires: [ apply production ]
+
       - run-task:
           name: backup production
           requires: [ apply production ]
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -603,6 +607,27 @@ jobs:
...
+    docker:
+      - image: cimg/base:2020.10
+    resource_class: small
+    steps:
+      - run:
+          name: PUT release time to OPG metrics
+          command: |
+            EPOCH_MILLISECONDS=$(date +%s%3N)
+
+            curl --location --request PUT ${METRICS_ENDPOINT} \
```
mattmachell, line 620: Worth checking the default curl timeout on circle if using this method.

acsauk: I can't see anything curl specific but the default timeout for a command on CircleCI is 10 minutes. Have you run into this issue before @mattmachell?

mattmachell: Was just thinking in terms of builds failing quickly and feedback cycles, and also how we handle failure modes here. I'd set in at 30 seconds to avoid builds hanging. `-m 30` for 30 seconds timeout (https://ec.haxx.se/usingcurl/usingcurl-timeouts).

acsauk: Ah I was coming at it from the opposite perspective 😂 makes sense, I'll add that in

jamesrwarren (approved):

PR #516, opened 2020-11-03
DDPB-3723: Refactor remaining events to make testable

acsauk:
## Purpose
Moves the logic and setup code around logging audit events from controllers to API classes and removes the existing tests that rely on overwriting the Symfony DI container during test runs.

Fixes DDPB-3723

## Approach
Initially, we'd planned to move the audit logging code to the API layer of the app to trigger from Doctrine lifecycle events. This worked well in practice but in doing so we lost one critical piece of information - what triggered the entity change. Doctrine lifecycle events are completely closed off and so adding in extra context to them could have involved creating additional listeners which would have added a layer of confusion on top of the process and required passing information to the API that technically it doesn't need. Instead, I've doubled down on abstracting any internal API calls into the `Client/Internal/XXXXApi` namespace and adding the logging code there so its called every time we request the API to make entity changes.

## Learning
I lost quite a bit of time trying to write a mock service for the `RestClient` that we could inject when the app is in unit_test mode but this turned out to be a dead-end as it required far more services to be mocked out due to the way we handle auth in the app. From looking at the code it seemed a little over-engineered and something we could look at simplifying in the future.

## Checklist


coderste (approved):
LGTM good work 👏🏻

coderste (approved):

coderste (approved):

PR #526, opened 2020-11-16
Revert to using v2 endpoint for client details

acsauk:
## Purpose
Fixes timeouts on the client details page and when searching for users by case number with a case reference linked to a client in a large Org.

Fixes DDPB-3781

## Approach
As part of the audit logging refactor we had reverted to using the old client endpoint rather than v2. Testing locally with a large org (1000 client and 150 users) there were no issues but on prod we've witnessed timeouts so this suggests its more than the users and clients causing the slowdown. I'll put a ticket in the backlog to try to recreate the problem locally but this fix should get us back to a working state.

## Checklist

### DIFF
--- a/api/src/AppBundle/v2/Assembler/UserAssembler.php
+++ b/api/src/AppBundle/v2/Assembler/UserAssembler.php
@@ -1,0 +1,8 @@
+<?php
+
+
+namespace AppBundle\v2\Assembler;
+
+class UserAssembler
+{
+}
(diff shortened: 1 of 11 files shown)


api/src/AppBundle/v2/Assembler/UserAssembler.php
```diff
@@ -0,0 +1,8 @@
+<?php
+
+
+namespace AppBundle\v2\Assembler;
+
+class UserAssembler
```
Gugandeep, line 6: Is this empty class required for something?

acsauk: Ah good spot - no I was making this while trying to fix some missing data coming back from the API. I'll drop it out.

Gugandeep (approved):
Looks good to me

jamesrwarren (approved):

PR #530, opened 2020-11-23
DDPB-3675 - Implement Cross Browser Testing using BrowserStack + Behat

Gugandeep:
## Purpose
Implement Cross Browser Testing into the pipeline to allow for regular testing of the app across multiple browsers we should be compliant with.

Fixes DDPB-3675

## Approach
The implementation makes use of our current Behat implementation. We are using BrowserStack as our provider for cross browser testing. Our Behat implementation is current using the Mink extension, and fortunately the Mink extension comes with a BrowserStack driver which would do a lot of the setup and configuration for us.

The 3 main parts of this ticket were creating and running our Behat tests, connecting to BrowserStack, and finally implementing this as a job in our pipeline. 

We decided to create a few small and simple tests initially. Quite a few of our already made tests are not compatible with the Selenium WebDriver that was required to use BrowserStack. There was a bit of configuration needed to make sure that we could easily run different browsers using the BrowerStack driver provided by Mink. We made use of the profiles functionality provided by Behat to achieve this.

Connecting to BrowserStack was done via the tunnelling software that they provide. Running the binary and making use of the driver provided by Mink would allow our tests to be run locally and the output sent to BrowserStack.

Lastly, we are implementing this cross browser testing as a step in your weekly integration run. The process, in brief, is it to get the required secrets from AWS, run the tunnelling binary, run the Behat tests, and finally stop the tunnelling binary.

## Learning
[Behat Docs](https://docs.behat.org/en/latest/guides.html)
[Mink Docs](http://mink.behat.org/en/latest/)
[MinkExtension Github](https://github.com/Behat/MinkExtension)
[BrowserStack Docs](https://www.browserstack.com/docs/)

## Checklist

### Frontend


acsauk (approved):
:shipit:

PR #539, opened 2020-12-01
DDPB-3555 add password validation for pwned passwords

jamesrwarren:
## Purpose
Stop users from using passwords that are too common by taking the top 100k known pwned passwords and making sure password is not in the list.

When we move to symfony 4 we should use the constraint that does this check automatically! It's cleverer than what we are doing.

Fixes DDPB-3555

## Approach
Went back and forth on approach a bit. Initially I wanted the list in the database but in the end after realising the size of the list (tiny) I went for a hybrid approach of DLing file locally where it doesn't already exist or if it's older than 24 hours and using that for the comparison. The check takes milliseconds so seems fine doing it this way.

I created a new bespoke constraint and added some unit tests and behat tests to check it.

## Learning
Constraints, unit testing constraints, (although not in the code, I learned quite a bit more about doctrine!)

## Checklist

### Frontend

### DIFF
--- a/behat/tests/features-v2/deputy-management/deputy/edit-myself.feature
+++ b/behat/tests/features-v2/deputy-management/deputy/edit-myself.feature
@@ -162,7 +162,15 @@ Feature: A deputy user edits their details
     And I press "change_password_save"
     Then the following fields should have an error:
       | change_password_plain_password_first |
-        # valid new password
+      #too common password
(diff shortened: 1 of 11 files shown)


Gugandeep (approved):
Looks good to me

acsauk (approved):
:shipit:

PR #541, opened 2020-12-02
DDPB-3733 do not fail healthcheck on sirius down

jamesrwarren:
## Purpose
We should keep the status in manage availability but we don't want the healthcheck to affect our uptime stats as it's not a vital part of our service and docs will just remain queued if it's down.

Fixes DDPB-3733

## Approach
Don't return unhealthy status if it's just sirius that's down.

## Learning
NA
## Checklist

### Frontend


Gugandeep (approved):
Looks good!

charlesmmarshall (approved):

PR #542, opened 2020-12-04
DDPB-3802: Fix empty query string

acsauk:
## Purpose
Fixes the date range picker to allow selecting date ranges when downloading the DAT file for CasRec/

Fixes DDPB-3802

## Approach
So we're still not 100% sure if this is PHP 7.4 related or not but it seems at some point the behaviour of `http_build_query()` changed to stop accepting DateTimes as valid entries so:

```
http_build_query([
                'fromDate' => $query->getStartDate(),
                'toDate' => $query->getEndDate(),
                'orderBy' => $query->getOrderBy(),
                'order' => $query->getSortOrder()
            ]
```

resulted in missing out fromDate and toDate in the query string sent to the API.

Unfortunately, the tests covering this function ALSO used `http_build_query()` to build the expected query string so when the behaviour changed the tests still passed. 

## Learning
Make sure we're hard coding test expectations and not using implementation code when writing tests.

## Checklist

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -281,7 +281,7 @@ workflows:
       - cross-browser-test:
           name: cross browser test
           requires: [ apply integration ]
-          filters: { branches: { ignore: [ main ] } }
+          filters: { branches: { only: [ main ] } }
       - run-task:
           name: integration test
           requires: [ reset integration ]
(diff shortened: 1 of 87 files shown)


Gugandeep (approved):
Good job finding the issue and getting a fix out before the weekend!

coderste (approved):

PR #543, opened 2020-12-07
ddpb3772b make clam less flake in integration

jamesrwarren:
## Purpose
For some reason we're not giving clam enough memory in integration and development.  Also removed an env var in envrc that shouldn't have been there.

## Checklist

### Frontend


acsauk (approved):
:shipit:

PR #544, opened 2020-12-07
DDPB-3752: Add audit logging for adding and removing a user from an org

acsauk:
## Purpose
Provides an audit trail for org membership updates.

Fixes DDPB-3752

## Approach
Following our established pattern of:

- Moving direct RestClient calls from controller to an API wrapper (in this case `OrganisationApi`)
- Emitting an event describing whats happened after a successful API call
- Consuming the event in an event subscriber and carrying out any other actions required (in this case, audit logging)

## Checklist


client/src/AppBundle/Controller/Admin/OrganisationController.php
```diff
@@ -245,11 +252,13 @@ public function deleteUserAction(Request $request, $id, $userId)
         $form->handleRequest($request);
 
         $organisation = $this->restClient->get('v2/organisation/' . $id, 'Organisation');
-        $user = $this->restClient->get('user/' . $userId, 'User');
+        $userToremove = $this->restClient->get('user/' . $userId, 'User');
```
coderste, line 255: Think the variable is missing part of camelCase 👍🏻

client/src/AppBundle/EventSubscriber/OrgUserMembershipSubscriber.php
```diff
@@ -0,0 +1,57 @@
+<?php declare(strict_types=1);
```
coderste, line 1: Would be good if we could DocComment the functions 👍🏻

client/src/AppBundle/Service/Audit/AuditEvents.php
```diff
@@ -161,4 +155,45 @@ public function userDeleted(string $trigger, string $deletedBy, string $subjectF
 
         return $event + $this->baseEvent($eventType);
     }
+
```
coderste, line 158: This function could do with a comment -- takes a fair few parameters

coderste (approved):

coderste (approved):

PR #546, opened 2020-12-10
DDPB-3449: Append file extensions to files missing a file extension

acsauk:
## Purpose
A common issue we run into when users upload a document is accidentally (or intentionally?) removing the file extension from the filename when they rename or edit a file. As we rely on the file extension to set the mime type of the file prior to sending to Sirius this would fail during the synchronisation step. This change instead automatically detects the file type based on file content and appends the relevant extension on the to file automagically for the user.

Fixes DDPB-3449

## Approach
Following some convos during planning, I looked for a package that would do the file introspection and settled on The PHP League's [mime-type-detection](https://github.com/thephpleague/mime-type-detection). Its got a sizeable user base, lots of GitHub stars and as its a TPL package we can be fairly confident that it should be supported going forward.

## Learning
I also took the opportunity to make time a specific external dependency for the S3Upload service to aid testing. This led to some confusion when trying to get the same level of precision that microtome() provides until I realised we were using epoch time + milliseconds rather than just epoch time.

## Checklist


jamesrwarren (approved):
This all looks great to me

PR #547, opened 2020-12-10
DDPB-3816: Fix failing lay CSV upload

acsauk:
## Purpose
Fixes the broken CSV upload.

Fixes DDPB-3816

## Approach
We had reports that the CSV upload process was hanging in production. Looking in cloudwatch there was an error related to the Ajaxcontroller that gets called when splitting the CSV into chunks of 2000 rows - a class wasn't being injected into the controller function that required it. 

I've fixed the typehint now but the reason our tests didn't pick this up is all of the CSVs we upload are under 2000 rows long. As chunks are set at 2000 rows the Ajaxcontroller was never called in tests. I'm looking into writing a JS enabled test in behat to cover this off but want to get the fix out now so the Lay team can get the most recent lays added to the system.

### DIFF
--- a/client/src/AppBundle/Controller/Admin/AjaxController.php
+++ b/client/src/AppBundle/Controller/Admin/AjaxController.php
@@ -4,7 +4,7 @@ namespace AppBundle\Controller\Admin;
 
 use AppBundle\Controller\AbstractController;
 use AppBundle\Service\Client\RestClient;
-use Predis\Client;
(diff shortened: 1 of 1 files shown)


Gugandeep (approved):
You're right it was a very simple fix!
Approved! :thumbsup:

jamesrwarren (approved):
Excellent! Nice and concise fix! :)

PR #548, opened 2020-12-11
DDPB-3645 scheduled workspace cleanup

jamesrwarren:
## Purpose

So we don't have to destroy environments as part of the pipeline and get left with random environments and loads of empty workspaces, we need a script that runs through them and destroys them.

Fixes DDPB-3645

## Approach

Went for an approach of locking them for 24 hours. Although it says 24 hours this basically means that they will remain for this working day and the next as the job runs at midnight each night. If you want to extend the duration you can either rerun the whole pipeline, change the expiry in dynamo or probably the quickest is just rerun the protect bit of the pipeline again. 

The bulk of the work here was actually clearing up all our old workspaces. Manually went through them with the script to make sure everything was in a fit state... took ages..

## Learning
NA

## Checklist

### Frontend

### DIFF
--- a/environment/scripts/workspace_cleanup.sh
+++ b/environment/scripts/workspace_cleanup.sh
@@ -1,13 +1,9 @@
-#!/usr/bin/env sh
+#!/usr/bin/env bash
 
 if [ $# -eq 0 ]
   then
     echo "Please provide workspaces to be removed."
 fi
-
-function getWorkspaces {
-  terraform workspace list
-}
 
 in_use_workspaces="$@"
 reserved_workspaces="default production preproduction development develop integration training production02"
@@ -17,6 +13,7 @@ all_workspaces=$(terraform workspace list|sed 's/*//g')
 
 unset TF_WORKSPACE
 export TF_VAR_OPG_DOCKER_TAG=""
+export TF_EXIT_CODE="0"
 
 for workspace in $all_workspaces
 do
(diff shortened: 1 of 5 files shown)


environment/scripts/workspace_cleanup.sh
```diff
@@ -0,0 +1,44 @@
...
+export TF_EXIT_CODE="0"
+
+for workspace in $all_workspaces
+do
+  case "$protected_workspaces" in
+    *$workspace*)
+      echo "protected workspace: $workspace"
+      ;;
+    *)
+      if [[ $workspace == *"prod"* ]] || [[ $workspace == *"dependabot"* ]]
```
acsauk, line 25: We have a dependabot workspace?

shared/workspace-cleanup.tf
```diff
@@ -0,0 +1,4 @@
+module "workspace-cleanup" {
+  source  = "github.com/TomTucka/terraform-workspace-manager/terraform/workspace_cleanup"
```
acsauk, line 2: Fine for now but are we planning to have this as an MOJ repo?

acsauk (dismissed):
:shipit: - all good, two questions out of curiosity but go ahead and merge.

environment/scripts/workspace_cleanup.sh
jamesrwarren: I was supposed to have removed this. Good spot! Fixed now

shared/workspace-cleanup.tf
jamesrwarren: Was thinking of pulling out the table setup and just using it as is but we're using the repo for the executable as well and it's Tom rather than external but yeah you're right, i'll ask him about whether we can move this to moj

acsauk (approved):
:shipit:

PR #550, opened 2020-12-17
Add path to live doc

acsauk:
## Purpose
Adding the path to live diagram so we don't lose it in the sea of document links we have flying around.


coderste (dismissed):
LGTM

acsauk:
This branch is cursed - I've done console resource deletions, manual destroys, pipeline destroys, rebases but Terraform still complains of a log resource to already exists. Starting a new branch instead of wasting any more time - a curse on your house `a`

PR #555, opened 2020-12-20
Ddpb 3428

jamesrwarren:
## Purpose
Add audit logging for unsubmit and resubmit events.

Fixes DDPB-3428

## Approach
Followed the patterns from previous jobs. Feel in hindsight it may have been better to split the send email and log for the submit into different events rather than have them as one submit event however it seemed easier to code and made sense that way at the time (EDIT: got picked up on this in code review and wasn't tricky to change in the end).

I also promised some tests based on actual log events. Whilst I can see how to do this (as it's all set up nicely for it already), by the time I got it all working as is I didn't have time. Will do it for the next one.

## Learning
Here's a tip which used up about 2 hours of head scratching!! You need to completely rebuild the containers for subscribers to recognise different numbers of params as it seems it registers this kind of thing on startup.. spend ages trying to work out where it was getting the values from... you live and learn!

## Checklist

### Frontend

### DIFF
--- a/client/src/AppBundle/Controller/Admin/Client/ReportController.php
+++ b/client/src/AppBundle/Controller/Admin/Client/ReportController.php
@@ -443,7 +443,7 @@ class ReportController extends AbstractController
             if ($form->has('confirm') && $form['confirm']->getData() === 'yes' && $report->isSubmitted()) {
                 $this->reportApi->unsubmit(
                     $report,
-                    $this->get('security.token_storage')->getToken()->getUser(),
+                    $this->getUser(),
                     AuditEvents::TRIGGER_UNSUBMIT_REPORT
                 );
                 $this->upsertChecklistInformation($report);
--- a/client/src/AppBundle/Controller/Admin/Client/ReportController.php
+++ b/client/src/AppBundle/Controller/Admin/Client/ReportController.php
@@ -5,7 +5,6 @@ namespace AppBundle\Controller\Admin\Client;
 use AppBundle\Controller\AbstractController;
 use AppBundle\Entity\Report\Checklist;
 use AppBundle\Entity\Report\Report;
-use AppBundle\Entity\User;
 use AppBundle\Exception\ReportNotSubmittedException;
 use AppBundle\Form\Admin\CloseReportConfirmType;
 use AppBundle\Form\Admin\CloseReportType;
@@ -16,7 +15,6 @@ use AppBundle\Form\Admin\ManageSubmittedReportType;
 use AppBundle\Form\Admin\ManageReportConfirmType;
 use AppBundle\Service\Client\Internal\ReportApi;
 use AppBundle\Service\Client\RestClient;
-use AppBundle\Service\Client\Internal\OrganisationApi;
 use AppBundle\Service\Audit\AuditEvents;
 use AppBundle\Service\ParameterStoreService;
 use AppBundle\Service\ReportSubmissionService;
@@ -97,9 +95,6 @@ class ReportController extends AbstractController
 
     /** @var ReportApi */
     private $reportApi;
-
-    /** @var OrganisationApi */
-    private OrganisationApi $organisationApi;
 
     private LoggerInterface $logger;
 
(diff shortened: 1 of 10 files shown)


client/src/AppBundle/Controller/Admin/Client/ReportController.php
```diff
@@ -436,7 +441,11 @@ public function manageConfirmAction(Request $request, $id)
             $this->restClient->put('report/' . $report->getId(), $report, ['report_type', 'report_due_date']);
 
             if ($form->has('confirm') && $form['confirm']->getData() === 'yes' && $report->isSubmitted()) {
-                $this->unsubmitReport($report);
+                $this->reportApi->unsubmit(
+                    $report,
+                    $this->get('security.token_storage')->getToken()->getUser(),
```
acsauk, line 446: All controllers have a built in `getUser()` function which saves us having to dip in and grab a service from the container:

```suggestion
                    $this->getUser(),
```

client/src/AppBundle/EventSubscriber/ReportSubmittedSubscriber.php
```diff
@@ -16,24 +19,59 @@ class ReportSubmittedSubscriber implements EventSubscriberInterface
...
+    {
+        if ($event->getSubmittedReport()->getUnSubmitDate() !== null) {
+            try {
+                $this->logReportSubmittedEvent($event);
+            } catch (\Exception $e) {
+            }
+        }
+        $this->sendEmail($event);
+    }
+
```
acsauk, line 54: We've tended to try to follow SRP with event subscribers to make testing as simple as possible and increase re-use. Could we get this split out to `sendEmail()` and `log()` (or something similar)?

client/src/AppBundle/Service/Client/Internal/ReportApi.php
```diff
@@ -7,6 +7,8 @@
 use AppBundle\Entity\Report\Report;
 use AppBundle\Entity\User;
 use AppBundle\Event\ReportSubmittedEvent;
+use AppBundle\Event\ReportUnsubmittedEvent;
+use AppBundle\Event\ReportResubmittedEvent;
```
acsauk, line 11: Is this being used?

acsauk (changes requested):
Nice job! One request for splitting out the email sending and logging to their own functions but otherwise good to go 💪

client/src/AppBundle/Service/Client/Internal/ReportApi.php
jamesrwarren: Nope, good spot, twas a hangover from when i was splitting submission into submission and resubmission but that approach created way too much duplicated code..

client/src/AppBundle/Controller/Admin/Client/ReportController.php
jamesrwarren: I thought this but php storm was saying nay (expecting entity rather than interface error) but it seems to work anyway. Cheers.

client/src/AppBundle/EventSubscriber/ReportSubmittedSubscriber.php
jamesrwarren: Yeah should have done this. Done now.

client/src/AppBundle/Controller/Admin/Client/ReportController.php
acsauk: Yeah I get that too - its one of those times where the IDE isn't quite smart enough to know we're going to get a proper entity user object back.

client/src/AppBundle/Service/Client/Internal/ReportApi.php
acsauk: I've done this a couple of times with other events - end up looking at what I've made and its basically the same as another event 😂

client/src/AppBundle/EventSubscriber/ReportSubmittedSubscriber.php
```diff
@@ -36,4 +49,17 @@ public function sendEmail(ReportSubmittedEvent $event)
             $this->mailer->sendReportSubmissionConfirmationEmail($event->getSubmittedBy(), $event->getSubmittedReport(), $newReport);
         }
     }
+
+    public function log(ReportSubmittedEvent $event)
```
acsauk, line 53: Sorry, had a last-minute thought to make this a bit clearer - we should probably make it clear to someone reading this subscriber that this logging only happens when a report is being resubmitted so something like `logResubmittedReport()`? Last comment, promise 😬

acsauk (changes requested):
Sorry made one last comment - I should have picked it up in initial review - apologies!

client/src/AppBundle/EventSubscriber/ReportSubmittedSubscriber.php
jamesrwarren: Yeah no worries. Good point!

acsauk (approved):
:shipit:

PR #556, opened 2020-12-29
DDPB-3130 add logging bucket

jamesrwarren:
## Purpose

We only have the main bucket that is of interest here. It already had versioning but needed logging

Fixes DDPB-3130

## Approach
added logging and made versioning have better expiration rules

## Checklist

### Frontend


acsauk (approved):
:shipit:

PR #557, opened 2020-12-30
DDPB-3290 - Folder changes to prepare for Symfony upgrade

acsauk:
## Purpose
Prepping for the Symfony upgrade we need to make some changes to the folder structure of the app along with some other conventions that [change in V4+](https://github.com/symfony/symfony/blob/4.4/UPGRADE-4.0.md):

Completed:

Symfony 3.x | Symfony 4.x
-- | --
app/config/ | config/
app/config/*.yml | config/*.yaml and config/packages/*.yaml
app/DoctrineMigrations | DoctrineMigrations
app/Resources/**BundleName**/views/ | templates/bundles/**BundleName**/
app/Resources/ | src/Resources/
var/logs/ | var/log/
web/ | public/
web/app.php | public/index.php
web/app_dev.php | public/index.php

Followup PR (to keep changelog into three digits (!)):

Symfony 3.x | Symfony 4.x
-- | --
app/config/parameters.yml.dist | config/services.yaml and .env.dist
app/config/parameters.yml | config/services.yaml and .env
src/AppBundle/Resources/assets/ | assets/
src/AppBundle/Resources/translations/ | translations/
src/AppBundle/Resources/views/ | templates/
src/AppBundle/ | src/

Additionally:

- Changed `SYMFONY_ENV` and `SYMFONY_DEBUG` to `APP_ENV` and `APP_DEBUG` (along with references in terraform, Makefile, Symfony console and index.php)
- Added the Dotenv component to support loading .env files locally while also ensuring we don't load .env files when the app is running hosted rather than locally
- Added an entry on manage availability page to show if the app is running in debug mode
- Moved the mock-sirius docker service into the main docker-compose file to ensure its cleaned up when brining docker down
- Moved the env files from the docker folder to root so it was more obvious we have them available

Fixes DDPB-3290 (partially)

## Checklist


client/public/index.php
```diff
@@ -0,0 +1,44 @@
...
+use Symfony\Component\Debug\Debug;
+use Symfony\Component\Dotenv\Dotenv;
+use Symfony\Component\HttpFoundation\Request;
+
+require dirname(__DIR__).'/vendor/autoload.php';
+require dirname(__DIR__) . '/app/AppKernel.php';
+
+// We don't want to load the .env files if we are running on a server - ENVIRONMENT is only injected via terraform
+if (empty($_SERVER['ENVIRONMENT'])) {
+    (new Dotenv())->load(dirname(__DIR__).'/frontend.env', dirname(__DIR__).'/admin.env');
```
mattmachell, line 12: Curious why the need for this? Is it a symfony-ism? Docker should take care of loading the env files and they should be accessible from inside the PHP runtime without having to load  in like this? Unless something is really weird about our setup.

client/public/index.php
```diff
@@ -0,0 +1,44 @@
...
+if ($trustedProxies = $_SERVER['TRUSTED_PROXIES'] ?? $_ENV['TRUSTED_PROXIES'] ?? false) {
+    Request::setTrustedProxies(explode(',', $trustedProxies), Request::HEADER_X_FORWARDED_ALL ^ Request::HEADER_X_FORWARDED_HOST);
+}
+
+if ($trustedHosts = $_SERVER['TRUSTED_HOSTS'] ?? $_ENV['TRUSTED_HOSTS'] ?? false) {
+    Request::setTrustedHosts([$trustedHosts]);
+}
+
+$_SERVER += $_ENV;
+$_SERVER['APP_ENV'] = $_ENV['APP_ENV'] = ($_SERVER['APP_ENV'] ?? $_ENV['APP_ENV'] ?? null) ?: 'dev';
```
mattmachell, line 37: Future you will probably hate having to parse what this chain of ternary operators is doing. Might be worth breaking it out to be more readable.

client/public/index.php
acsauk: The main driver behind this was how we were handling the environment mode and debug profiling options via a hacky local command to create a file (see https://github.com/ministryofjustice/opg-digideps/blob/66727991d164521a97a4cbbc635fc057dc9a29b6/client/web/app_dev.php#L24). I was having trouble getting Symfony to take notice of externally loaded env vars we were defining in docker-compose and as Symfony made the Dotenv component a built-in part of version 4+ I thought it would be a good idea to try to standardise how we're handling env vars.

I'll have another play around with seeing how env vars are respected to be sure there wasn't something else that was the issue as I agree it would be nice to just have one place to inject them.

client/public/index.php
acsauk: Agree - I'd taken this from some official Symfony files but even as I was adding it I had to manually parse through in a REPL to make sure it was doing what I thought it was.

jamesrwarren (commented):
All seems sensible to me

client/public/index.php
acsauk: Awesome, its working without the dotenv component so I'll drop to keep things simpler.

jamesrwarren (approved):

PR #558, opened 2020-12-30
DDPB-3755 amend the circle workflow to fit our desired design

jamesrwarren:
## Purpose
With this ticket there's a number of things we have done:

- Brought in parallel plans for all envs
- Built the majority of env during the app build stage
- Use images from build stage to run client and api tests instead of rebuilding the local env for them
- Included a step so we can unprotect env (especially good for where we have added long protection to an env)
- Followed the desired plan mostly but deviated from it where it made sense. 
- Added a smoke test for prod (currently running against preprod so we can be double sure it won't do anything bad against a non branch env)
- Optional workflow for destroy env set before apply steps at Alex request to be able to get rid of broken envs without resorting to command line (see final point below in "Things I tried but didn't work out well")
- Seems to be a problem pulling from build docker container on commonpasswords file so made it part of CI but kept it as optional in build step for local builds (so we don't have to store the full file locally (not that it's big but it will go out of date and seems unnecessary). This means if there's still a genuine issue with site then it will not build but we should stop getting those weird connection issues.
- Added some documentation 

image

Things I tried but didn't work out well:
- Caching between builds (this looks promising but we aren't on plan for docker layer caching so can only cache vendor for example and seemed like it wasn't worth it).
- Using plans from previous plans for the builds. This is probably still a good idea but the code was getting messy(-er) because of the way I have set up the full plans beforehand. Was a little concerned that a bad variable replacement would make it use the wrong plan (like the prod plan for example!) so decided to leave this for now..
- Looked into adding a separate, run anytime, destroy stage but circle just isn't made for this kind of thing so have done above workaround. 

Fixes DDPB-3755

## Approach
As above

## Learning
Using images between steps makes sense (though because of the tar and untar time it doesn't save as much as we would like). We probably should have done this sooner.

## Checklist

### Frontend


coderste (approved):
LGTM 👍🏻

coderste (approved):

docker-compose.test.yml
```diff
@@ -0,0 +1,14 @@
...
+    frontend:
+        image: frontend:latest
+    admin:
+        image: admin:latest
+    api:
+        image: api:latest
+    localstack-init:
+        image: localstack-init:latest
+    wkhtmltopdf:
+        image: wkhtmltopdf:latest
```
acsauk, line 14: May be good to add ci in the filename for this just to keep it in line with the other ci docker-compose files.

.circleci/config.yml
```diff
@@ -516,20 +722,13 @@ jobs:
...
-            --terms_to_waitfor="apply,plan" \
-            --prod_job_terms=" production,shared-production"
-          working_directory: ~/project/ecs_helper/cmd/circle_builds
       - run:
           name: Set environment
           command: ~/project/.circleci/set_env.sh >> $BASH_ENV
+      - run:
+          name: Download common passwords file
+          command: wget -q -O commonpasswords.txt "https://www.ncsc.gov.uk/static-assets/documents/PwnedPasswordsTop100k.txt"
+          working_directory: ~/project/client/extra
```
acsauk, line 731: Cant tell from the wget flags but is this now not failing when we get a 500 and continuing with the build?

acsauk (commented):
Awesome - one Q on the password intermittent failure we've been seeing recently.

.circleci/config.yml
jamesrwarren: It's quiet but still does a non zero error code. I will make it non quiet anyway as you're right, there's nothing insecure in there and it would be good to see the failure!

docker-compose.test.yml
jamesrwarren: Yes good point!

acsauk (approved):
:shipit:

PR #559, opened 2021-01-04
DDPB-3833 try out step scaling to avoid pointless alarms

jamesrwarren:
## Purpose
Our previous scaling had two very annoying issues and one slightly annoying issue:

1) Because the alarms were purely based on simple metrics once it would reach minimum count, it would just stay in alarm permanently. Because min count was set on scaling policy this didn't do anything but it makes it unnerving for anyone looking at the service. According to aws documentation this is fine but I personally don't like to see loads of alarms muddying the waters..

2) Because it was based on two separate metrics it's highly possible for the autoscaling to cause more problems than it solves by trying to scale up and scale down at the same time. For example if we have high CPU but normal memory then one policy will be trying to start containers whilst another will be scaling down.

3) This isn't too much of an issue but threshold alarms for ecs just don't seem good for small scale infra as they try and keep levels at a certain volume. Step scaling gives more control. I can see threshold alarms maybe being of value when you have like 10 tasks or something and you're removing or adding to keep levels consistent. Our use case is more the ability to scale up from our standard 1 to 3 tasks when there's high load and to go back down to that minimum.

Fixes DDPB-3833

## Approach
What we want instead is something a bit more clever. After some experimentation, came up with using math expressions to combine various metrics in a more logical way that doesn't alarm when service is already at lowest task count and takes into consideration both memory and cpu for scaling. Notice we use AND on scale down and OR on scale up. This logic is still mutually exclusive.

Here's example of it working.

image

image

Also included a really rough script to put load on our front service. This is nothing near a load testing script but I felt it might be useful to have in the future for other work where we just need to quickly stress the system.

## Learning
Had to relearn how the multithreading channels in go worked as had completely forgotten..

## Checklist

### Frontend


docs/scripts/stress_service.go
```diff
@@ -0,0 +1,47 @@
...
+
+// The purpose of this script is to put some basic stress on the application for testing of alarms and scaling etc
+// It is not meant to be used as a full load testing suite.
+// Replace the base url with the branch you wish to test and make sure you are connected to the VPN.
+func main() {
+	baseUrl := flag.String("base_url", "ddpb0000.complete-deputy-report.service.gov.uk", "a string")
+	urlSuffix := flag.String("url_suffix", "manage/availability", "a string")
+	batchSize := flag.Int("batch_size", 50, "an int")
+	numberOfBatches := flag.Int("number_of_batches", 100, "an int")
+	waitBetweenBatches := flag.Int("wait_between_batches", 1, "an int")
```
acsauk, line 31: Just checking, do all of these need to be exported as env vars when running the script?

docs/scripts/stress_service.go
```diff
@@ -0,0 +1,47 @@
...
+	flag.Parse()
+
+	url := "https://" + *baseUrl + "/" + *urlSuffix
+	var respStatus int
+	chStatus := make(chan int)
+
+  for i := 0; i < *numberOfBatches; i++ {
```
acsauk, line 38: Do we need the `*` here? I'm reading that as when `i` is less than nothing multiplied by `numberOfBatches`.

acsauk (dismissed):
Seems like a good change - I did remember seeing long-running alarms in AWS so will be nice to have them silenced. Couple of Qs on the load test script but nothing worth blocking merging if its working as expected 👍

docs/scripts/stress_service.go
jamesrwarren: No they have defaults but just means you can overwrite them on command line if you so wish

docs/scripts/stress_service.go
jamesrwarren: Nah it's a pointer cos it's taken from flag so flag is a pointer to an int :)

charlesmmarshall (dismissed):

docs/scripts/stress_service.go
acsauk: Of course it is - I need to get my hands back on some Go code!

acsauk (approved):
:shipit:

PR #561, opened 2021-01-05
Composer update - lets see what breaks

acsauk:
Its been a while since we did a composer update - so I did one. Its removed a bunch of unused packages from client and updated almost every package we use. Hopefully, all package maintainers are making non-breaking changes 😁


jamesrwarren (approved):
Seems to have passed the tests so all good there :)

PR #562, opened 2021-01-06
DDPB-3755b fix workflow dependency in main pipeline

jamesrwarren:
didn't quite get workflow dependencies right in main.


acsauk (approved):
:shipit:

PR #563, opened 2021-01-07
DDPB-3290: Move remaining files to Symfony 4 structure

acsauk:
## Purpose
Using the new file structure in preparation for Symfony 4 update. Although there are loads of files here the main changes were:

- Moving contents of `src/AppBundle` -> `src/`
- Moving contents of `src/Resources/*` -> `templates/`, `translations/` and `/assets`
- Changing the namespace AppBundle to App and updating any references 
- Using file locations in some config files rather than previous AppBundle namespaced references
- Adopting the more modern twig reference style (e.g. `AppBundle::two:three.html.twig` -> `AppBundle/two/three.html.twig`) and update all references in controllers, services and templates
- Removing database reset between each API unit test suite (due to pipeline failures after passing in env var for the app mode)

Fixes DDPB-3290

## Learning
A happy accident resulted in halving the run time of the API tests. We were doing a database drop and fixture load using the Symfony console between each step which is where the slowdown was coming from, but due to some subtle changes in how we're setting the app mode during test runs following these updates that approach no longer worked. I had to tweak some tests to make then isolated which is why there are some test changes present but the end result is quicker tests 💪 .

Oh also, I faced my regex fears and managed to work out the right regex for doing all the twig updates:

Replace all instances of `App:` to have an `@` at the front
```
/App:/

Replace text:

@$0
```

Change `:` to `/` in twig references:
```
/(?=\b\d*:\b).(?=.*\bhtml.twig\b)/

Replace text:

/
```

## Checklist


mattmachell:
>A happy accident resulted in halving the run time of the API tests.

Well that's great to hear!

coderste (approved):

coderste:
That approval wasn't final btw ^ I was testing to see if you could approve without clicking 'viewed' on all files :D turns out you can

coderste (approved):
Approved this time 👍🏻

PR #564, opened 2021-01-11
Ddpb 3836

jamesrwarren:
## Purpose
Move the multi deputy functionality from the registration part of the form to the client verification part of the form so that users created by case managers through the admin console will still adhere to this functionality.

This shouldn't have any affect on self registrations as they can't do anything until they have gone through the verification step anyway.

Fixes DDPB-3836

## Approach
Add a behat test to cover this scenario and add a missing unit test to cover the whole verify function.

## Learning
NA
## Checklist

### Frontend

### DIFF
--- a/behat/tests/features/admin/01-admin.feature
+++ b/behat/tests/features/admin/01-admin.feature
@@ -14,6 +14,11 @@ Feature: admin / admin
     Then I should see "behat-admin-user@publicguardian.gov.uk" in the "users" region
     And the response status code should be 200
 
+  Scenario: login and add lay user
+    Given I am logged in to admin as "admin@publicguardian.gov.uk" with password "Abcd1234"
+    And I create a new "NDR-enabled" "Lay Deputy" user "Joe" "Bloggs" with email "joe.bloggs@publicguardian.gov.uk" and postcode "SW1"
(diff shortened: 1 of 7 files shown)


behat/tests/features/admin/01-admin.feature
```diff
@@ -14,6 +14,46 @@ Feature: admin / admin
...
+      | 12345XYZ | Smith   | D003      | Bloggs      | SW1          | OPG102    |
+      | 12345XYZ | Smith   | D004      | Doe         | SW1          | OPG102    |
+    When I activate the user "joe.bloggs@publicguardian.gov.uk" with password "Abcd1234"
+    And I go to "logout"
+    And I go to "/login"
+    And I fill in the following:
+      | login_email     | joe.bloggs@publicguardian.gov.uk |
+      | login_password  | Abcd1234 |
+    And I press "login_login"
+    Then the url should match "/user/details"
```
acsauk, line 33: For consistency we could probably replace this with:

```suggestion
And I am logged in as "joe.bloggs@publicguardian.gov.uk" with password "Abcd1234"
```

acsauk (changes requested):
Nice job - this will save a fair amount of manual updates. One change for consistency on the behat test but otherwise good to go 👍

PR #567, opened 2021-01-13
Ddpb 3836b

jamesrwarren:
## Purpose
Move the multi deputy functionality from the registration part of the form to the client verification part of the form so that users created by case managers through the admin console will still adhere to this functionality.

This shouldn't have any affect on self registrations as they can't do anything until they have gone through the verification step anyway.

Fixes DDPB-3836

## Approach
Add a behat test to cover this scenario and add a missing unit test to cover the whole verify function.

## Learning
NA
## Checklist

### Frontend


acsauk (approved):
:shipit:

PR #572, opened 2021-01-15
DDPB-3852: Fix registration flow and add missing eventSubscriber

acsauk:
## Purpose
Removing some old court order code that was leftover that resulted in some registrations not completing and adding a missing send email listener for self-registration events.

Fixes DDPB-3852.

## Learning
We clearly have some holes in our behat tests. The sooner we get on the rewrite the better.

## Checklist

### Frontend

### DIFF
--- a/client/templates/Manage/api-collector.html.twig
+++ b/client/templates/Manage/api-collector.html.twig
@@ -1,4 +1,4 @@
-{% extends 'WebProfilerBundle/Profiler/layout.html.twig' %}
+{% extends '@WebProfiler/Profiler/layout.html.twig' %}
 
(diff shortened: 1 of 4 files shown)


client/templates/Manage/api-collector.html.twig
```diff
@@ -1,4 +1,4 @@
-{% extends 'WebProfilerBundle/Profiler/layout.html.twig' %}
+{% extends '@WebProfiler/Profiler/layout.html.twig' %}
```
jamesrwarren, line 1: yay :)

jamesrwarren (approved):

PR #574, opened 2021-01-18
DDPB-3850 make co_deputy logic make sense

jamesrwarren:
## Purpose
So previous attempt to fix this was missing some of the context. The point of the co_deputy_client_confirmed flag is really to check check for the redirect route for a joining co deputy. 
It's not really needed to be active on the main deputy. However, as it is a useful signifier currently of whether the co deputy has been added, I have left this in. I have no idea why it was being set as part of registration. It shouldn't be used as a check for whether a case has multiple deputies as their are calls to check against multiple count in casrec for that case_nnumber that achieve that.

Fixes DDPB-3850

## Approach
I've re-added the functionality from /lay route to /ndr route. As these are the only two routes that can invite deputies, this functionality is now consistent. Also as it doesn't rely on the value of the field and instead on casrec table, we shouldn't need to update any cases.

## Learning
NA

## Checklist

### Frontend


client/src/Controller/Ndr/NdrController.php
```diff
@@ -105,11 +110,14 @@ public function indexAction(Redirector $redirector)
 
         $clients = $user->getClients();
         $client = !empty($clients) ? $clients[0] : null;
-        $coDeputies = !empty($client) ? $this->getCoDeputiesForClient($user) : [];
+
+        $clientWithCoDeputies = $this->restClient->get('client/' . $client->getId(), 'Client', ['client', 'client-users', 'user']);
```
acsauk, line 114: Could we replace this with:

```suggestion
        $clientWithCoDeputies = $this->clientApi->getWithUsersV2($client->getId()));
```

Trying to move restclient calls into the appropriate API classes when we come across them.

acsauk (changes requested):
Cool makes sense - I think I was the one that made the change to call out to casrec rather than rely on the flag so should have also removed setting it during reg flow. 

Just the one change then this is ready to go 👍

acsauk (approved):
:shipit:

api/src/v2/Controller/ClientController.php
```diff
@@ -61,7 +61,7 @@ public function getByIdAction(int $id): JsonResponse
 
     /**
      * @Route("/case-number/{caseNumber}", methods={"GET"})
-     * @Security("has_role('ROLE_ADMIN') or has_role('ROLE_AD')")
+     * @Security("has_role('ROLE_ADMIN') or has_role('ROLE_AD') or has_role('ROLE_DEPUTY')")
```
acsauk, line 64: I think we may need to tighten up the security on this to protect against exposing client info accidentally to people who shouldn't see it. As we limited the use of the endpoint to admins only prior to this change there may be instances where we were relying on the security rule. Could we get some kind of check on if the logged-in user is not an admin and not associated with the client then we throw a 403 or similar?

jamesrwarren: Added this check now

acsauk (approved):
:shipit:

Gugandeep (approved):
Re-ticked

PR #577, opened 2021-01-26
Ddpb 3854 

jamesrwarren:
## Purpose
As part of our ITHC we were informed of improvements we could make to our handling of form fields with regards to css. This PR tackles the issues from a validation perspective and also an execution perspective.

Fixes DDPB-3854

## Approach
We are adding checks to the majority of text fields that applies a validation rule that the most useful chars to a would be hacker can't be saved. Whilst it would be nice to be even more restrictive, it's not practical due to weird real chars in addresses and names etc that need to be handled.

In any case this is mainly to stop any future hiccups on the behalf of a dev as the other part of this is to remove all instances of raw from our templates except to display our css stylesheets. Anything where a var could be passed in should have raw removed.

Finally added a pre commit hook to not allow any more instances of raw in the templates (where not for parsing css file) and a further check in our pipeline that gives a nice message telling you not to do it!

## Learning
After chatting with Alex we had a cool plan for mass testing of formatting changes that involved outputting html snapshots from each step of main branches and comparing them to the same from this. Unfortunately it was too messy to look at in the end as it seemed to format the html quite differently even when things looked the same on screen (different lines). Shame as it could be an interesting idea if we got it right. I do think there's some merit in this approach for future formatting work but we would have to maybe use a smaller selection of snapshots and parse both lots of files.

## Checklist

### Frontend


acsauk:
Noticed a couple of places during the pro journeys that had some formatting and translating issues but overall looks great!
Screenshot 2021-01-27 at 11 46 04
Screenshot 2021-01-27 at 11 45 20
Screenshot 2021-01-27 at 11 36 47

jamesrwarren:
> Noticed a couple of places during the pro journeys that had some formatting and translating issues but overall looks great!
These are fixed now along with other bits I've picked up manually and from behat.

jamesrwarren:
Closing and opening another one as ripping this in half and wanted to keep the other bits just in case..

PR #582, opened 2021-01-28
Ddpb 3821

jamesrwarren:
## Purpose
As part of the spike into guardduty, none of the following changes actually affect gd in any way but they are some tweaks I made during the spike. 

The changes include:
- Adding VPC logging. Need to monitor how expensive this is. I've gone for same retention as other projects but it does log a lot of rows so will have to monitor the costs...
- Fixing a dangling RDS logs resource. It stop environments being recreated with the same name as it doesn't currently get destroyed properly.

Before destroy:
image

After destroy:
image

- Don't allow tf lock files to be committed.
- Update the certificate spec as it was only issue stopping us upgrading to tf 0.14.5. However I decided not to upgrade TF right now as we have some big tickets going through pipe.

Outcome of DDPB-3821

## Learning
NA

## Checklist

### Frontend


acsauk (approved):
Nice - hopefully solves our pipeline recreates now 🙌

PR #584, opened 2021-01-28
DDPB-3860 add timeout to request to sirius

jamesrwarren:
## Purpose
Sirius was still causing us issues when it goes down. Despite putting in a fix to still bring back 200 when sirius is down for our availability as seen in screenshot below. It still seems that it would just hang above and beyond what healthcheck is expecting (defaults to 5 secs). As such have tried this against non responsive endpoint and it fails against sirius check in 2 secs and sends back 200 which is what we want it to do when sirius is down.

image

I know it's a bit cheeky bu I also noticed we ought to scale down our min serverless config. 2 is the minimum it allows. This is because RDS is by far our biggest cost in dev atm:

image

Fixes DDPB-3860

## Approach
Add timeouts to sirius request

## Learning
NA

## Checklist

### Frontend

### DIFF
--- a/api/Dockerfile
+++ b/api/Dockerfile
@@ -1,4 +1,4 @@
-FROM composer AS composer
+FROM composer:2.0.8 AS composer
 
 WORKDIR /app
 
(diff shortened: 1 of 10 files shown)


acsauk (approved):
Awesome - feels like we're finally free of Sirius downtime issues with this change 👍

PR #587, opened 2021-02-01
DDPB-3697 rightsize redis

jamesrwarren:
## Purpose
This should just work. After running the behat tests locally and looking at the keys stored in redis it all seems to be:

image

Both of these should be unique. The short one is actually based on userid it seems which is unique in the user table.

image

As such this all should just work fine.

EDIT: This didn't work just fine. If session is being stored across admin and frontend like behat does when running certain tests then it can cause variables to not be set. The solution is to add admin and front to the prefixes so then it will not use session from the other app type as it's own.

Fixes DDPB-3697

## Approach
Changed the local config. Checked everything still worked. Updated terraform.

## Learning
NA

## Checklist

### Frontend


Makefile
```diff
@@ -81,6 +82,12 @@ reset-database: ## Resets the DB schema and runs migrations
 reset-fixtures: ## Resets the DB contents and reloads fixtures
 	docker-compose run --rm api sh scripts/reset_db_fixtures_local.sh
 
+redis-clear: ## Puts app in dev mode and disables debug (so the app runs faster, but no toolbar/profiling)
```
acsauk, line 85: The description here needs an update/removal

environment/checklist_sync.tf
```diff
@@ -122,9 +122,10 @@ locals {
       { "name": "S3_BUCKETNAME", "value": "pa-uploads-${local.environment}" },
       { "name": "APP_ENV", "value": "${local.account.app_env}" },
       { "name": "OPG_DOCKER_TAG", "value": "${var.OPG_DOCKER_TAG}" },
-      { "name": "ADMIN_HOST", "value": "https://${aws_route53_record.admin.fqdn}" },
+      { "name": "ADMIN_HOST", "value": "https://${aws_route53_record.frontend_redis.fqdn}" },
```
acsauk, line 125: Should this remain as admin?

environment/document_sync.tf
```diff
@@ -118,9 +118,10 @@ locals {
       { "name": "S3_BUCKETNAME", "value": "pa-uploads-${local.environment}" },
       { "name": "APP_ENV", "value": "${local.account.app_env}" },
       { "name": "OPG_DOCKER_TAG", "value": "${var.OPG_DOCKER_TAG}" },
-      { "name": "ADMIN_HOST", "value": "https://${aws_route53_record.admin.fqdn}" },
+      { "name": "ADMIN_HOST", "value": "https://${aws_route53_record.frontend_redis.fqdn}" },
```
acsauk, line 121: Same as checklist, should this still be admin?

acsauk (commented):
A couple of Qs on same var values but awesome job - should be a nice saving. Incidentally, you may have fixed an issue I was having with some behat fixture steps - will pull this branch in and see!

environment/checklist_sync.tf
jamesrwarren: omg yes!

acsauk (approved):
:shipit:

PR #590, opened 2021-02-03
Ddpb 3854b

jamesrwarren:
As part of our ITHC we were informed of improvements we could make to our handling of form fields with regards to css. This PR tackles the issues from a validation perspective and also an execution perspective.

Fixes DDPB-3854

##Approach

We are adding checks to the majority of text fields that applies a validation rule that the most useful chars to a would be hacker can't be saved. Whilst it would be nice to be even more restrictive, it's not practical due to weird real chars in addresses and names etc that need to be handled.

In any case this is mainly to stop any future hiccups on the behalf of a dev as the other part of this is to remove all instances of raw from our templates except to display our css stylesheets. Anything where a var could be passed in should have raw removed.

Finally added a pre commit hook to not allow any more instances of raw in the templates (where not for parsing css file) and a further check in our pipeline that gives a nice message telling you not to do it!

##Learning

After chatting with Alex we had a cool plan for mass testing of formatting changes that involved outputting html snapshots from each step of main branches and comparing them to the same from this. Unfortunately it was too messy to look at in the end as it seemed to format the html quite differently even when things looked the same on screen (different lines). Shame as it could be an interesting idea if we got it right. I do think there's some merit in this approach for future formatting work but we would have to maybe use a smaller selection of snapshots and parse both lots of files.

## Checklist

### Frontend


client/templates/Components/Form/_input.html.twig
```diff
@@ -7,12 +7,12 @@
                 <span class="govuk-caption-xl">{{ label.caption }}</span>
             {% endif %}
             <label class="govuk-label govuk-label--xl {{ labelClass | default('') }}" for="{{ element.vars.id }}">
-                {{ labelRaw is defined and labelRaw ? label.text | raw : label.text }}
+                {{ labelRaw is defined and labelRaw ? label.text : label.text }}
             </label>
```
Gugandeep, line 11: Is this check necessary if the end result for both cases is using label.text?

client/templates/Report/Report/_report_row.html.twig
```diff
@@ -14,9 +14,6 @@
         <p class="text flush--bottom govuk-body">
             {% block bottomLeftContent %}
             {% endblock %}
-            {#{{ bottomLeftContent | raw }}#}
-            {#{{ (page ~ '.editReportingPeriodPara') | trans(transOptions) }}#}
```
Gugandeep, line 18: Was this translation removed all together?

client/templates/Components/paginator.html.twig
```diff
@@ -19,7 +19,11 @@
 <div>
     {% if totalRecords <= recordsPerPage %}
         <p data-test-id="pager-summary">
-            {{ messages.singlePage | transchoice(totalRecords) | raw }}
+            {% if totalRecords == 1 %}
+            {{  messages.singlePage.singular | transchoice(totalRecords) }}
+            {% else %}
+            {{  messages.singlePage.plural | transchoice(totalRecords) }}
```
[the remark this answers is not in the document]
jamesrwarren: yeah we didn't need to discern between zero and more than 1 cos they both are counted as plural

client/templates/Components/Form/_input.html.twig
jamesrwarren: Good spot!

client/templates/Report/Report/_report_row.html.twig
jamesrwarren: It was already commented out so I just removed the commented out text

Gugandeep: Ah didn't spot that it was commented out.

client/templates/Components/paginator.html.twig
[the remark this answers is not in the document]
Gugandeep: Great stuff.

Gugandeep (approved):
Great work. Looks good to go!

PR #591, opened 2021-02-08
DDPB-3886: Fix active lay report

acsauk:
## Purpose
The active lay report was timing out on prod as there were approximately 23K Users being returned. This change streamlines the process to allow the full dataset to be returned.

Fixes DDPB-3886

## Approach
I replicated creating a large dataset using a factory locally and running in a test multiple times. I backed up the test database and restored to the API database. We could look at building this functionality into our fixtures so we have a representative DB size to work with when we need to test anything with high loads. 

## Checklist

### Frontend

### DIFF
--- a/api/src/Controller/StatsController.php
+++ b/api/src/Controller/StatsController.php
@@ -35,7 +35,7 @@ class StatsController extends RestController
     }
 
     /**
-     * @Route("stats/activeLays", methods={"GET"})
+     * @Route("stats/deputies/lay/active", methods={"GET"})
      * @Security("has_role('ROLE_SUPER_ADMIN')")
      */
     public function getActiveLays()
--- a/api/src/Controller/StatsController.php
+++ b/api/src/Controller/StatsController.php
@@ -14,8 +14,10 @@ class StatsController extends RestController
     private QueryFactory $QueryFactory;
     private UserRepository $userRepository;
 
-    public function __construct(QueryFactory $QueryFactory, UserRepository $userRepository)
-    {
+    public function __construct(
+        QueryFactory $QueryFactory,
+        UserRepository $userRepository
+    ) {
         $this->QueryFactory = $QueryFactory;
         $this->userRepository = $userRepository;
     }
@@ -30,5 +32,14 @@ class StatsController extends RestController
         $query = $this->QueryFactory->create($params);
 
         return $query->execute($params);
+    }
+
+    /**
+     * @Route("stats/activeLays", methods={"GET"})
+     * @Security("has_role('ROLE_SUPER_ADMIN')")
+     */
+    public function getActiveLays()
+    {
+        return $this->userRepository->findActiveLaysInLastYear();
     }
 }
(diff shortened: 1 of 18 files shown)


api/src/Controller/StatsController.php
```diff
@@ -31,4 +33,13 @@ public function getMetric(Request $request)
 
         return $query->execute($params);
     }
+
+    /**
+     * @Route("stats/activeLays", methods={"GET"})
```
mattmachell, line 38: Would be temped to break this up into a more RESTful set of resources for future reuse. Something like stats/deputies/lay/active

acsauk: Nice - will update 👍

api/src/Entity/Repository/UserRepository.php
```diff
@@ -206,20 +206,36 @@ public function findInactive($select = null)
...
-            ->setParameter('login_cutoff', $oneYearAgo)
-            ->setParameter('lay_deputy_role', User::ROLE_LAY_DEPUTY);
-
-        return $qb->getQuery()->getResult();
+        $oneYearAgo = (new DateTime())->modify('-1 Year')->format('Y-m-d');
+
+        $conn = $this->getEntityManager()->getConnection();
+
+        $sql = <<<SQL
+SELECT u.id,
```
jamesrwarren, line 218: Looks good :)

jamesrwarren (dismissed):

api/src/Entity/Repository/UserRepository.php
```diff
@@ -206,20 +206,36 @@ public function findInactive($select = null)
...
+u.registration_date,
+u.last_logged_in,
+c.firstname as client_first_name,
+c.lastname as client_last_name,
+COUNT(r.id) as submitted_reports
+FROM dd_user as u
+LEFT JOIN deputy_case as dc on u.id = dc.user_id
+LEFT JOIN client as c on dc.client_id = c.id
+LEFT JOIN report as r on c.id = r.client_id
+WHERE r.submit_date is not null AND u.role_name = 'ROLE_LAY_DEPUTY' AND u.last_logged_in > :oneYearAgo
```
mattmachell, line 232: Just a random query, but what fidelity do we need this data in? Do we want > :oneYearAgo or >= :oneYearAgo

mattmachell:
Oh, one other question. Do we use the Aurora cluster's read endpoint at the moment in DD? For this kind of MI query it's probably best to use it.

api/src/Entity/Repository/UserRepository.php
acsauk: Good point - its anyone whos logged in within the last year. I'll whack an `=` on.

acsauk:
I'm pretty much in the dark about how we use Aurora other than I know we updated to it a couple of months back - any idea @jamesrwarren?

mattmachell:
> I'm pretty much in the dark about how we use Aurora other than I know we updated to it a couple of months back

Might be that moving to a read-only endpoint is a bit beyond scope for this then and should be a distinct followup ticket. Sirius have the read only endpoint configured and send all their MI queries to it to stop overloading the read/write one.

jamesrwarren (approved):

PR #592, opened 2021-02-10
DDPB-3874 - Contacts report section behat re-write

acsauk:
## Purpose
The first of our new style behat tests! This tests the contacts section with the following requirements:

- No contacts
- Single contact
- Multiple contacts
- Section navigation arrows

Fixes DDPB-3874

## Approach
A few false starts while trying to auth without having to fill a form in (I gave up in the end and stored by attempts [here](https://github.com/ministryofjustice/opg-digideps/tree/behat-auth-attempt)) but I like how this has come out. Forcing lots of the individual steps behind a plain English description allowed me to use fake data in tests which can then be asserted on without having to type random static values into the gherkin syntax itself. I'm going to run this by Stacey and see if it makes sense so there may be an update or two following that.

Every scenario creates its own data and clears out previous data automatically by using a `@BeforeScenario` filter step so this way it will be impossible to fall back in to bad habits and rely on data created in previous steps to pass future steps. This also FINALLY makes TDD starting from an integration level possible by having a short(ish) feedback loop when running single scenarios.

I've also started a `v2` folder for contexts (I know, I know, they're multiplying at the moment) with the view to separate out any new behat code we write there so we can safely delete all the old code once we cover off all features of the app.

## Learning
Our `RestClient` and auth approach is really really difficult to emulate in tests. I'm not sure if its behat getting in the way or not but I just couldn't get the app to respect the `AuthToken` added to headers on subsequent requests once provided after a successful auth call to the API (see link above in Approach). Anyone that can get this working will get a crate of whatever beverage they want.

## Checklist

### Frontend


api/src/TestHelpers/BehatFixtures.php
```diff
@@ -0,0 +1,57 @@
...
+    public function loadFixtures()
+    {
+        $purger = new ORMPurger($this->entityManager);
+        $purger->purge();
+
+        // Add admin users
+        $adminUser = (new User())
+            ->setFirstname('Admin')
+            ->setLastname('User')
+            ->setEmail('admin@publicguardian.gov.uk')
```
mattmachell, line 37: We probably want to have a pattern for fixtures relating to a particular test to avoid potential collisions in parallel runs and the DB vs ones that will never change.

acsauk: Great point - I'll add something in for that

coderste (approved):

acsauk:
Closing as this env is cursed and I'm not about to waste half a day trying to fix it 😭

PR #593, opened 2021-02-12
[DDPB-3288] Fix Symfony v4.x deprecations

coderste:
## Purpose
Symfony 4.x versions introduced a lot of deprecations which will be finalised in Symfony 5 (e.g. functions deprecated in v4.x are removed in v5). We have removed some dependancies and upgraded the Symfony version to 4

Fixes DDPB-3288

## Checklist


api/src/Controller/ManageController.php
```diff
@@ -62,7 +62,7 @@ private function dbInfo()
             if ($e instanceof \PDOException && $e->getCode() === 7) {
                 $returnMessage = 'Database service not reachabe (' . $e->getMessage() . ')';
             }
-            if ($e instanceof \Doctrine\DBAL\DBALException) {
+            if ($e instanceof \Exception) {
```
acsauk, line 65: I wouldn't necessarily say an `Exception` type signifies the migrations table is missing...if there's no alternative to `Doctrine\DBAL\DBALException` we may as well lose the if block completely and leave the generic error message defined above.

client/composer.json
```diff
@@ -25,28 +25,18 @@
...
-            "Sensio\\Bundle\\DistributionBundle\\Composer\\ScriptHandler::installAssets",
-            "Sensio\\Bundle\\DistributionBundle\\Composer\\ScriptHandler::installRequirementsFile"
+            "Incenteev\\ParameterHandler\\ScriptHandler::buildParameters"
         ],
         "post-update-cmd": [
-            "Incenteev\\ParameterHandler\\ScriptHandler::buildParameters",
-            "Sensio\\Bundle\\DistributionBundle\\Composer\\ScriptHandler::buildBootstrap",
-            "Sensio\\Bundle\\DistributionBundle\\Composer\\ScriptHandler::clearCache",
-            "Sensio\\Bundle\\DistributionBundle\\Composer\\ScriptHandler::installAssets",
-            "Sensio\\Bundle\\DistributionBundle\\Composer\\ScriptHandler::installRequirementsFile"
```
acsauk, line 49: Cool if it working but what were these post-install commands actually doing? Does the cache get cleared automatically now post-install?

acsauk (commented):
Excellent work - couple of comments but looking good.

I'm not sure if you've got [pre-commit](https://pre-commit.com/) installed but it will automatically take care of the missing EOFs on some of the config files you have there before you commit changes courtesy of [this](https://github.com/ministryofjustice/opg-digideps/blob/main/.pre-commit-config.yaml).

api/src/Command/MigrationsMigrateLockCommand.php
```diff
@@ -25,13 +25,6 @@ class MigrationsMigrateLockCommand extends MigrationsMigrateDoctrineCommand
     protected function configure(): void
     {
         parent::configure();
-
-        $this
-            ->setName('doctrine:migrations:migrate-lock')
-            ->setDescription('Same as doctrine:migrations:migrate, but locking the database.')
-            ->setHelp('')
-            ->addOption('release-lock', null, InputOption::VALUE_NONE, 'Release lock and exit.')
-        ;
```
acsauk, line 34: Was there an incompatibility reason this was removed? I'm kind of hesitant to stop locking the DB when we carry out migrations...

coderste: Yeah I removed this to try get the app running as it said the command was invalid. I've added it back now as the `MigrationsMigrateDoctrineCommand` no longer exists but I've changed it to just `Command`

jamesrwarren (approved):
I do hereby declare this ticket ticked! 🎉

acsauk (approved):
Aaaaaand breath 😁

PR #594, opened 2021-02-17
DDPB-3895: Send a Google Analytics event to Google

acsauk:
## Purpose
Adds in some scaffolding to standardise how we send Google Analytics events to Google.

Fixes DDPB-3895

## Approach
I've added some Javascript code that checks for elements containing `data-attribute=ga-event` which then extracts other `data-*` values and pings of an event to Google for tracking. I've updated the twig form extensions we have to have to include a `form_submit_ga()` function so we can extend this to other forms where we need to without having to touch markup. We may want to add more of these helpers but that seemed out of scope for this ticket.

I also made the decision to update our twig extension helpers to all user `TwigFunction` rather than `Simple_TwigFunction` which was deprecated. I was going to leave this for a separate PR but I needed the newer function to get this code working which would have a mixture of old and new functions in the file so it felt better to do it all at once.

I've also added the first of our javascript unit tests using Jest. We still have a gap in our testing when it comes to behat/integration tests to make sure this all works together but at least we have some initial coverage for the code I just added. I've put a ticket in the backlog (https://opgtransform.atlassian.net/browse/DDPB-3905) to update our pipeline to run the JS tests as part of our PR checks.

Finally, I added a new runbook for GA to formalise how we work with the event side of things. As we start to use GA more we should keep this updated.

## Learning
My JS is so rusty it took a good couple of days to get into the swing of how we deal with JS in the app. I've chosen to use the `class` sugary syntax out of habit from working with PHP day in day out and added a babel plugin to handle older browsers. This did raise a question of style and approach to writing JS as there are a number of different ways we do it in the app. Something for a dev discussion I think.

## Checklist

### Frontend

### DIFF
--- a/client/assets/javascripts/__tests__/modules/googleAnalyticsEvents.test.js
+++ b/client/assets/javascripts/__tests__/modules/googleAnalyticsEvents.test.js
@@ -36,17 +36,6 @@ const setDocumentBody = () => {
     `
 }
 
-const simulateClick = (element) => {
-  // Create our event (with options)
-  const event = new global.MouseEvent('click', {
-    bubbles: true,
-    cancelable: true,
-    view: window
-  })
-
-  element.dispatchEvent(event)
-}
-
 beforeAll(() => {
   setDocumentBody()
   GoogleAnalyticsEvents.init()
@@ -71,8 +60,8 @@ describe('googleAnalyticsEvents', () => {
   describe('clicking button', () => {
     describe('when gtag is loaded', () => {
       it('dispatches gtag event', () => {
-        simulateClick(document.getElementById('button1'))
-        simulateClick(document.getElementById('button2'))
+        document.getElementById('button1').click()
+        document.getElementById('button2').click()
 
         expect(window.gtag).toHaveBeenCalledWith(
           'event',
--- a/client/assets/javascripts/__tests__/modules/googleAnalyticsEvents.test.js
+++ b/client/assets/javascripts/__tests__/modules/googleAnalyticsEvents.test.js
@@ -1,0 +1,32 @@
+import GoogleAnalyticsEvents from '../../modules/googleAnalyticsEvents'
+
+const setDocumentBody = () => {
+  document.body.innerHTML = `
+        <div>
+            <button data-attribute="gae">1</button>
+            <button data-attributettribute="gae">2</button>
+        </div>
+    `
+}
+
+describe('googleAnalyticsEvents', () => {
+  describe('init', () => {
+    it('attaches event listeners to elements with data-attributes=gae', () => {
+      setDocumentBody()
+      const buttons = document.querySelectorAll('button[data-attribute="gae"]')
+
+      const spies = []
+
+      buttons.forEach(button => {
+        spies.push(jest.spyOn(button, 'addEventListener'))
+      })
+
+      GoogleAnalyticsEvents.init()
+
+      spies.forEach(spy => {
+        expect(spy).toHaveBeenCalledTimes(1)
+        expect(spy).toHaveBeenCalledWith('userStartsURSection', expect.any(Function))
+      })
+    })
+  })
+})
--- a/client/assets/javascripts/__tests__/modules/googleAnalyticsEvents.test.js
+++ b/client/assets/javascripts/__tests__/modules/googleAnalyticsEvents.test.js
@@ -1,4 +1,4 @@
-import GoogleAnalyticsEvents from '../../modules/googleAnalyticsEvents'
+import { GoogleAnalyticsEvents } from '../../modules/googleAnalyticsEvents'
 import { describe, it, jest } from '@jest/globals'
 
 const setDocumentBody = () => {
@@ -22,15 +22,15 @@ const setDocumentBody = () => {
     `
 }
 
-const simulateClick = (elem) => {
+const simulateClick = (element) => {
   // Create our event (with options)
-  const event = new MouseEvent('click', {
+  const event = new global.MouseEvent('click', {
     bubbles: true,
     cancelable: true,
     view: window
   })
 
-  elem.dispatchEvent(event)
+  element.dispatchEvent(event)
 }
 
 describe('googleAnalyticsEvents', () => {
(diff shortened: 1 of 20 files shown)


Gugandeep (dismissed):
Looks good to me!

client/assets/javascripts/modules/googleAnalyticsEvents.js
```diff
@@ -0,0 +1,29 @@
+class GoogleAnalyticsEvents {
+  static init () {
+    const elements = document.querySelectorAll('button[data-attribute="ga-event"]')
+
+    elements.forEach(element => {
```
mattmachell, line 5: Better to use event delegation for this (https://davidwalsh.name/event-delegate). Listen for click at the document level and only respond if it has a matching ga-event attribute.

client/assets/javascripts/modules/googleAnalyticsEvents.js
```diff
@@ -0,0 +1,29 @@
+class GoogleAnalyticsEvents {
```
mattmachell, line 1: Worth checking that ES6 class semantics work on all our target browsers and if we need to polyfill for older browsers (IE-old for example).

client/assets/javascripts/modules/googleAnalyticsEvents.js
```diff
@@ -0,0 +1,29 @@
...
+          event_label: eventElement.dataset.label
+        }
+    }
+  }
+
+  static sendEvent (event) {
+    const eventElement = event.target
+    const eventInfo = this.extractEventInfo(eventElement)
+
+    global.gtag('event', eventInfo.action, eventInfo.event_params)
```
mattmachell, line 25: I would be tempted to add some defensive code here to check that gtag exists before trying to use it, to avoid async loading issues.

client/assets/javascripts/modules/googleAnalyticsEvents.js
acsauk: Nice, that's a much better approach 👍

docs/runbooks/GOOGLE-ANALYTICS.md
```diff
@@ -0,0 +1,76 @@
...
+We use Google Analytics to track events within the app to highlight patterns in user behaviour. This can be used to analyse or evaluate a change made to a page, for example:
+
+ - Do additional form fields result in a drop in the amount of users submitting a form?
+ - Has changes to the style of a button resulted in more people clicking it?
+ - Do users scroll all the way to the bottom of the page or just click the continue button?
+
+ Google Analytics accepts a number of data points related to an event:
+
+ #### Category (required)
+ Describes the page, module or component the event belongs to.
```
mattmachell, line 13: Think we should give examples, maybe even a table of the expected categories in DigiDeps by current pages to give a feel for how this will work? 

Looking at the convention from the post you reference, I wonder if we could make it a bit cleaner.

Category: {Page Title}:{Sub Section i.e. in a form, optional}
Action: {Event }:{Element Type}
Label: {Human summary and additional detail}

Category: Finances : Outgoings
Action: Click : Help Link
Label: Clicked link to help page  /help

docs/runbooks/GOOGLE-ANALYTICS.md
```diff
@@ -0,0 +1,76 @@
...
+| Add Client Details Form |
+| Add Report Dates |
+| User Details Form |
+| Set Password Form |
+
+#### Actions
+
+| Action |
+|---|
+| Clicked Sign Up Button |
```
mattmachell, line 43: Adding onto my comment above, if we did action as: {event} : { Element Type} : {Element Specifics} it would be still human readable, but allow for easier sorting/grouping etc.

`Click : Button : Sign Up`

client/assets/javascripts/modules/googleAnalyticsEvents.js
acsauk: I've dropped in `"@babel/plugin-transform-classes"` to take care of older IE versions. The only exception was where a call extends a built-in type which we aren't doing so we should be good but I'll run the browserstack tests to be sure.

client/assets/javascripts/modules/googleAnalyticsEvents.js
acsauk: Agree - added in. I ended up wrapping the gtag global function in a globals namespace so I could test it. Not sure if there are any problems with this approach? I don't touch JS often enough to know if having globals is even frowned on but it made testing near impossible.

docs/runbooks/GOOGLE-ANALYTICS.md
acsauk: Makes perfect sense - I've updated the readme and events to follow this now.

client/assets/javascripts/__tests__/modules/googleAnalyticsEvents.test.js
```diff
@@ -0,0 +1,103 @@
...
+              id='button1'
+              data-attribute="ga-event"
+              data-action="form-submitted"
+              data-category="user-journeys"
+              data-label="button-clicks"
+            >1</button>
+            <button
+              id='button2'
+              data-attribute="ga-event"
+              data-action="back-to-report"
```
mattmachell, line 31: Something that just occurred to me. Worth name-spacing the data attributes to avoid collisions with any later scripts that have to act on the same element. i.e data-ga-action

acsauk: Yep good point - updated!

client/assets/javascripts/__tests__/modules/googleAnalyticsEvents.test.js
```diff
@@ -0,0 +1,101 @@
...
+          'back-to-report',
+          { event_category: 'testing', event_label: 'site-interaction' })
+      })
+    })
+
+    describe('when gtag is not loaded', () => {
+      it('does not dispatch gtag event', () => {
+        window.globals.gtag = null
+
+        simulateClick(document.getElementById('button1'))
```
mattmachell, line 94: You can trigger click event by calling document.getElementById('button1').click() assuming you have a viable DOM, which could simplify the test harness here.

acsauk: Much nicer - updated 👍

jamesrwarren (approved):

PR #597, opened 2021-02-22
DDPB-2914 add bucket backup replication

jamesrwarren:
## Purpose
Added all the digideps side for bucket replication to a backup account. The org infra side has already been merged in.

Fixes DDPB-2914

## Approach

Tried the full setup in a PR on our side first to check everything lined up then pulled out the backup account stuff and opened a PR in org infra with it. 

## Learning

Couple of things.. you need a filter setting to have two replication setups even if you don't use it. That was a bit of a gotcha. Also the whole dynamic for_each setup of all the vars to do per env setup on the org infra side (not in this PR) was a bit of learning to get it working as I wanted. We may want to remove dev or even preprod from the replication in the future but it's nice to have the option to replicate and proves it works through real envs before hitting prod.

## Checklist

### Frontend

### DIFF
--- a/environment/s3.tf
+++ b/environment/s3.tf
@@ -6,11 +6,20 @@ locals {
   noncurrent_expiration_days = contains(local.long_expiry_workspaces, local.environment) ? 365 : 7
 }
 
+data "aws_region" "current" {}
+
+// GET DEV REPLICATION BUCKET
 data "aws_s3_bucket" "replication_bucket" {
   bucket   = "pa-uploads-branch-replication"
   provider = aws.development
 }
 
+// DATA SOURCE FOR DEFAULT KEY
+data "aws_kms_alias" "source_default_key" {
+  name = "alias/aws/s3"
+}
+
+// CREATE THE MAIN BUCKET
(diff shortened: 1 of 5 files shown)


environment/s3.tf
```diff
@@ -70,11 +106,7 @@ resource "aws_s3_bucket_public_access_block" "pa_uploads" {
   restrict_public_buckets = true
 }
 
-resource "aws_s3_bucket_policy" "pa_uploads" {
-  bucket = aws_s3_bucket_public_access_block.pa_uploads.bucket
-  policy = data.aws_iam_policy_document.pa_uploads.json
-}
-
+// BASE POLICY FOR PA_UPLOADS
```
acsauk, line 109: I find these comments throughout super useful - its sometimes hard to look at terraform and know what its doing without having the docs up 👍

acsauk (approved):
Nice job 👍 

:shipit:

environment/s3.tf
jamesrwarren: Yeah they're a bit ugly but I actually find them useful myself too so left them in.

PR #598, opened 2021-03-03
DDPB-3776: Feedback on submission confirmation page

acsauk:
## Purpose
Adds a user research feedback from after the submission page so we can contact deputies interested in attending UR sessions.

Fixes DDPB-3776.

## Approach
I've taken time on this ticket to make sure the behat tests are in the new format we have which has further improved the scaffolding around our behat test runs. Namely:

- Returning additional info related to fixture users created between each scenario. In addition to email we now get:
    - Client Id
    - Current report ID, type and if its an NDR or annual report
    - Previous report ID, type and it it was an NDR or annual report (think submitted reports) 

This change allows us to have fixture users available that are in various stages of submitting a report:
- Not started
- Completed but not submitted
- Submitted

This takes care of an annoying issue we had where we had to step through and manually complete field in the report to get to a certain state. Now we have the users ready to go at the start of every test so no more repetition in testing the same fields over and over 🙌 .

I also took the opportunity to use UUIDs as the primary key for the two new entities I created here. We can use this going forward for any new entities we create to make our primary keys more resilient to change and reduce malicious users from exploiting the incremental id style we use.

## Learning
This was actually the first time in over a year that I created an entity from scratch that had a relationship in place with another entity and it took me an AGE to work out the right annotation to use for a one-to-one relationship. The main point I took away was that you can't have both a `@JoinColumn` and `@Column` annotation on an entity property otherwise it ignores the value in `@JoinColumn` and won't generate foreign keys during migration generation.

## Checklist

### Frontend


jamesrwarren (dismissed):
Thanks for the full walkthrough and explanation on all this Alex, happy to tick now :)

jamesrwarren (approved):

PR #603, opened 2021-03-10
DDPB-3911: Create elevated admin role

acsauk:
## Purpose
In preparation for giving case managers/admin users more responsibilities, we need something between the existing admin and super admin rights. This change creates an `ELEVATED_ADMIN_ROLE` that has the power to edit and add all user roles except for super admins. This is essentially the same as the existing admin role but by splitting this out we can assign certain actions to be limited to super and elevated admins only.

Fixes DDPB-3911

## Approach
I'v rejigged some of the UserVoter code to take into account the similarities for adding and editing a user. I also expanded the tests to cover all roles and ensured that PAs cant edit Profs (and vice versa) as they were previously allowed to do.

## Checklist

### Frontend


api/config/security.yml
```diff
@@ -130,7 +130,7 @@ security:
 
     # https://opgtransform.atlassian.net/wiki/spaces/DEPDS/pages/191299647/User+roles
```
Gugandeep, line 131: Does the page in the wiki require updating?

client/templates/Report/Debt/edit.html.twig
```diff
@@ -23,6 +23,7 @@
     {% for debt in form.debts %}
         <div class="{{ debt.moreDetails is defined ? ' js-details-expander ' : '' }}">
             {{ form_row(debt.debtTypeId) }}
+
```
Gugandeep, line 26: Was this intentional?

Gugandeep (dismissed):
Good work. Looks good to go.

api/config/security.yml
acsauk: Very good point (and I'd never seen this page before 😂 ) I've updated it now

client/templates/Report/Debt/edit.html.twig
acsauk: I had not - removed now 👍

client/src/Security/ClientVoter.php
```diff
@@ -0,0 +1,121 @@
...
+        $user = $token->getUser();
+
+        if (!$user instanceof User) {
+            // the user must be logged in; if not, deny access
+            return false;
+        }
+
+
+        switch ($attribute) {
+            case self::VIEW:
```
coderste, line 61: Do we need this case if it's not doing anything?

coderste (commented):
Just left a question about one of the switch statements other than that this looks good to me 👍🏻

coderste (approved):

client/src/Security/ClientVoter.php
acsauk: Stacking the cases above each other like this is just saying that if the attribute is view or edit then return the result of `$this->canManage()`

PR #604, opened 2021-03-11
DDPB-3898 - Implement soft delete functionality for organisations

Gugandeep:
## Purpose
Implement soft delete functionality for organisations to improve our data retention.
Currently organisations are hard deleted and before doing so, users and clients are removed from the organisation.
This means that it is would be very hard to track the history of a client or user around an organisation if an issue arose.

Fixes DDPB-3898

## Approach
Replicated soft delete functionality from other entities (client).

Deleting an organisation now updates the `deleted_at` column for that organisation.
The frontend of the app will now only return organisations which do not have a value in the `deleted_at` column.
Deleting an organisation has certain restrictions - there must not be any active entities still attached to the organisation.
In this case, there should be no `Users` or `Clients` with both the the `deleted_at` and `archived_at` set to `null`.

## Learning
Had a lot of issues with what seemed like a simple migration.
Previous migrations were failing due to the addition of the `deleted_at` column in what was a new migration version.
To fix this issue, we located the version in which the `organisation` table was first created and added the extra column in its creation. 

## Checklist

### Frontend


api/src/Service/RestHandler/OrganisationRestHandler.php
```diff
@@ -95,7 +97,28 @@ public function create(array $data): Organisation
...
-     * @param array $data
+     * @param int $id
+     * @return bool
+     * @throws OptimisticLockExceptionAlias
+     * @throws \Doctrine\ORM\ORMException
+     */
+    public function delete(int $id)
+    {
+        if (null === ($organisation = $this->orgRepository->find($id))) {
+            throw new \Doctrine\ORM\ORMException();
```
Gugandeep, line 108: Not sure if this is the right exception to be throwing.

api/src/Service/RestHandler/OrganisationRestHandler.php
```diff
@@ -11,6 +11,8 @@
 use Doctrine\ORM\OptimisticLockException as OptimisticLockExceptionAlias;
 use Symfony\Component\Validator\Validator\ValidatorInterface;
 
+use function Amp\Promise\rethrow;
```
acsauk, line 14: Is this being used anywhere?

api/src/Service/RestHandler/OrganisationRestHandler.php
acsauk: From looking at the class it seems to be intended for use when something has gone wrong within Doctrine rather than if doctrine hasn't been able to find an entity. Unless we're going to handle the exception based on the exception type thrown then a regular Exception will probably be fine here.

If you do decide to go down a custom exception route then see the `/Exception` folder for how others have been structured.

api/tests/App/Entity/Repository/OrganisationRepositoryTest.php
```diff
@@ -0,0 +1,152 @@
...
+        self::assertTrue($result);
+
+        //Org2 with 1 user
+        //And 1 soft deleted client
+        //And 1 archived client
+        //should have active entities.
+        $this->fixtures->addClientToOrganisation($clientArchived2->getId(), $orgs[1]->getId());
+        $this->em->flush();
+        $result = $this->sut->hasActiveEntities($orgs[1]->getId());
+        self::assertTrue($result);
```
acsauk, line 132: The tests all look great but it may be easier to read if the different cases are split out into their own functions:

```suggestion
/** @test */
    public function testHasActiveEntities_no_entities_in_org_returns_false()
    {
         ...
     }
     
     /** @test */
    public function testHasActiveEntities_soft_deleted_client_in_org_returns_false()
    {
         ...
     }
     
     etc
```

This way when tests fail in the future you know exactly what's gone wrong rather than having to work out which scenario in a big bunch of assertions have failed.

api/tests/App/Entity/Repository/OrganisationRepositoryTest.php
```diff
@@ -0,0 +1,152 @@
...
+use App\Entity\Repository\OrganisationRepository;
+use App\Entity\Organisation;
+use DateTime;
+use Doctrine\Common\DataFixtures\Purger\ORMPurger;
+use Doctrine\ORM\EntityManagerInterface;
+use Doctrine\ORM\Mapping\ClassMetadata;
+use Symfony\Bundle\FrameworkBundle\Test\WebTestCase;
+use Tests\Fixtures;
+
+use function Symfony\Component\String\s;
```
acsauk, line 14: Is this being used?

acsauk (changes requested):
Nice Job! Couple of comments on making the tests more user friendly but the code looks sound 👍

PR #606, opened 2021-03-12
DDPB-3919 update dependencies and fix activation issues

jamesrwarren:
## Purpose
Whilst looking into the activation issues, a few things came up. This addresses some things:

- Issues caused by double clicking it seems (probably better way of stopping this but added some tags around the submits to stop it)
- there was a missing template
- there was a template issue where we're calling characters letters
- Also chucking in an upgrade to dependency that dependabot mentioned

Fixes DDPB-3919

## Approach
NA

## Learning
The activation issue was very tricky to work out but obvious in hindsight (shout out to Charles for his help). 
The webpack upgrade was a bit picky. We can't upgrade node any further with the versions of packages we have. There's also a fun error based on the docker version being used by circle for remote_docker images that means this is the package selection chosen...

## Checklist

### Frontend

### DIFF
--- a/client/package.json
+++ b/client/package.json
@@ -12,6 +12,7 @@
     "@babel/eslint-parser": "^7.12.16",
     "@babel/plugin-proposal-class-properties": "^7.12.13",
     "@babel/preset-env": "^7.12.1",
+    "@jest/globals": "^26.6.2",
     "babel-jest": "^26.6.3",
     "babel-loader": "^8.1.0",
     "copy-webpack-plugin": "^5.1.2",
@@ -19,7 +20,6 @@
     "del": "^3.0.0",
     "govuk-elements-sass": "^3.1.0",
     "jest": "^26.6.3",
-    "@jest/globals": "^26.6.2",
     "jquery": "^3.5.1",
     "jshint": "^2.12.0",
     "mini-css-extract-plugin": "^0.8.2",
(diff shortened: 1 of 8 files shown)


client/package.json
```diff
@@ -29,7 +29,7 @@
     "standard": "^14.3.4",
     "stickyfilljs": "^2.1.0",
     "url-loader": "^3.0.0",
-    "webpack": "^4.44.2",
+    "webpack": "^5.24.4",
```
acsauk, line 32: I'm a little bit wary of jumping to a new major version without doing some checks on if any JS features break. Could we go to 4.46 instead? That should update the elliptical dependabot warning.

client/src/Controller/AbstractController.php
```diff
@@ -14,7 +14,7 @@ abstract class AbstractController extends Controller
      */
     protected function renderError(string $description, $statusCode = 500)
     {
-        $text = $this->renderView('TwigBundle/Exception/template.html.twig', [
+        $text = $this->renderView('bundles/TwigBundle/Exception/template.html.twig', [
```
acsauk, line 17: Oh snap was this broken the whole time? Sure I had tested this locally when I made the change 😭

client/package.json
jamesrwarren: I did do some checks and it all looked fine but now you've said that I don't want egg on my face when something breaks so putting it to that suggested version ;)

acsauk (approved):
Good detective work @jamesrwarren and @charlesmmarshall - I'm thinking we should find a way to update all of our buttons to have this on by default as when is it ever a good thing to make two form submission? Will add a ticket to the backlog :shipit:

PR #608, opened 2021-03-12
Ddpb 3921

jamesrwarren:
## Purpose

Every day we get an annoying alert around response times from admin because of a casrec csv upload. It's distracting and might mean we miss genuine alerts through not looking at it properly.

Fixes DDPB-3921

## Approach

So I have combined the admin metric to be a combined metric that does the old response time check but overlays a check for a particular route in the logs over it (the upload). It will now only trigger when the response times are triggering and the upload hasn't happened in the last minute. This is super safe as they only do it once a day..

The bit I thought was a bit clever was how I tested it. I set up two routes in the app: badlongresponse and goodlongresponse. They both had sleeps in them to make them take 4 seconds to respond. 

The theory was that the orig alam would trigger on both but our new improved alarm would only trigger on badlongresponse endpoint as we had added that path in the combined metric. So i had a curl looping that would trigger them and took screenshots of results:

good long response
image

You can see here how the additional check overlays with the response and stops it triggering :)
image

bad long response
image

## Learning
All the above ;)

## Checklist

### Frontend

### DIFF
--- a/client/src/Controller/ManageController.php
+++ b/client/src/Controller/ManageController.php
@@ -70,28 +70,6 @@ class ManageController extends AbstractController
     }
 
     /**
-     * @Route("/badlongresponse", methods={"GET"})
-     */
-    public function badLongResponse()
-    {
-        sleep(4);
-        return new Response(
-            '<html><body>Long response</body></html>'
-        );
-    }
-
-    /**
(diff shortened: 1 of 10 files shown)


acsauk (approved):
Nice nice nice - very inventive approach :shipit:

PR #610, opened 2021-03-18
DDPB-3898 - Implement soft delete functionality for organisations

Gugandeep:
## Purpose
Implement soft delete functionality for organisations to improve our data retention.
Currently organisations are hard deleted and before doing so, users and clients are removed from the organisation.
This means that it is would be very hard to track the history of a client or user around an organisation if an issue arose.

Fixes DDPB-3898

## Approach
Replicated soft delete functionality from other entities (client).

Deleting an organisation now updates the `deleted_at` column for that organisation.
The frontend of the app will now only return organisations which do not have a value in the `deleted_at` column.
Deleting an organisation has certain restrictions - there must not be any active entities still attached to the organisation.
In this case, there should be no `Users` or `Clients` with both the the `deleted_at` and `archived_at` set to `null`.

## Learning
Had a lot of issues with what seemed like a simple migration.
Previous migrations were failing due to the addition of the `deleted_at` column in what was a new migration version.
To fix this issue, we located the version in which the `organisation` table was first created and added the extra column in its creation. 

Unfortunately we then realised that this would not work in production. Production has a list of all the migrations it has already done, and it would not pickup and run the new change done to an old migration.

So, back to solution one - creating a new migration that created the 'deleted_at' column.
Faced the same issue as last time, but this time was able to debug it.
Realised that a workaround could be made where we just retrieve the fields needed to run that migration.
This avoids the code needlessly trying to look up the 'deleted_at' column.

Seems like a this might be underlying issue where although we can store various versions of our database, we can't store versions of what our code looked like at the time. For further issues where we make changes that might affect previous migrations, it's most likely we just need to change how the migration is implemented.

## Checklist

### Frontend


api/src/Entity/Repository/OrganisationRepository.php
```diff
@@ -12,13 +12,43 @@ class OrganisationRepository extends EntityRepository
...
     }
 
+    /**
+     * @return array
+     */
+    public function getNonDeletedArray(): array
+    {
+        $query = $this
+            ->getEntityManager()
+            ->createQuery('SELECT o FROM App\Entity\Organisation o WHERE o.deletedAt IS NULL');
```
jamesrwarren, line 32: I have no idea if this works or not so take this comment with a pinch of salt but I wonder if this will do the same thing without the WHERE clause because of that gedmo malarkey. I think it's nice having it be more specific but not sure if the whole gedmo thing might confuse ppl in the future. Like maybe we could remove the where clause and write a comment? I don't know. To be honest it's probably fine as is. Just a thought :)

acsauk (changes requested):
All looks good to go - although I just messed up your day by merging a PR that has the same version number as you 😭  once that's updated I'm ready to tick

acsauk (approved):
:shipit: woohoo, LGTM!

PR #613, opened 2021-03-28
DDPB-3474 actions section behat tests

jamesrwarren:
## Purpose

Capture possible permutations around the actions section of the report form

image

Fixes DDPB-3474

## Approach

After talking with Alex, he suggested that we should not use the old behat links. This was a slight challenge for some elements. Also he suggested not to use variables where possible (though in one instance it did just make total sense to use variables and this can be reused for all report sections).

As such an approach that seems sensible is having  specific code in the section trait that calls more generic functions in the reporting sections feature context.

I have tried where possible to navigate the app rather than just visiting urls. The purpose of this is to make sure the correct links are where we expect them on each page and work as expected. I did have one issue here..

## Learning

I feel we could do with a nice way of hitting the right link when editing without resorting to behat tags. Couldn't quite work it out..

## Checklist

### Frontend

### DIFF
--- a/behat/tests/bootstrap/v2/common/IShouldBeOnTrait.php
+++ b/behat/tests/bootstrap/v2/common/IShouldBeOnTrait.php
@@ -64,10 +64,50 @@ trait IShouldBeOnTrait
     }
 
     /**
+     * @Then I should be on the financial decision actions page
+     */
+    public function iAmOnActionsPage1()
+    {
+        $this->iAmOnPage('/report\/.*\/actions\/step\/1$/');
+    }
+
+    /**
+     * @Then I should be on the concerns actions page
+     */
+    public function iAmOnActionsPage2()
+    {
+        $this->iAmOnPage('/report\/.*\/actions\/step\/2$/');
+    }
+
+    /**
+     * @Then I should be on the actions report last step page
+     */
+    public function iAmOnActionsLastStepPage()
+    {
+        $this->iAmOnPage('/report\/.*\/actions\/summary\?from\=last\-step$/');
+    }
+
+    /**
+     * @Then I should be on the actions report summary page
+     */
+    public function iAmOnActionsSummaryPage()
+    {
+        $this->iAmOnPage('/report\/.*\/actions\/summary/');
+    }
+
+    /**
+     * @Then I should be on the Lay main overview page
+     */
+    public function iAmOnLayMainPage()
+    {
+        $this->iAmOnPage('/lay$/');
+    }
+
+    /**
      * @Then I should be on the Lay reports overview page
      */
     public function iAmOnReportsOverviewPage()
     {
-        $this->iAmOnPage('/lay$/');
+        $this->iAmOnPage('/report\/.*\/overview$/');
     }
 }
--- a/behat/tests/bootstrap/v2/common/IShouldBeOnTrait.php
+++ b/behat/tests/bootstrap/v2/common/IShouldBeOnTrait.php
@@ -80,23 +80,15 @@ trait IShouldBeOnTrait
     }
 
     /**
-     * @Then I should be on the actions report last step page
-     */
-    public function iAmOnActionsLastStepPage()
-    {
-        $this->iAmOnPage('/report\/.*\/actions\/summary\?from\=last\-step$/');
-    }
-
-    /**
      * @Then I should be on the actions report summary page
      */
     public function iAmOnActionsSummaryPage()
     {
-        $this->iAmOnPage('/report\/.*\/actions\/summary/');
+        $this->iAmOnPage('/report\/.*\/actions\/summary.*/');
     }
 
     /**
-     * @Then I should be on the Lay main overview page
+     * @Then I should be on the Lay homepage
      */
     public function iAmOnLayMainPage()
     {
--- a/behat/tests/bootstrap/v2/reporting/sections/ActionsSectionTrait.php
+++ b/behat/tests/bootstrap/v2/reporting/sections/ActionsSectionTrait.php
@@ -1,0 +1,143 @@
+<?php declare(strict_types=1);
+
+namespace DigidepsBehat\v2\Reporting\Sections;
+
+trait ActionsSectionTrait
+{
+    private int $answeredYes = 0;
+    private int $answeredNo = 0;
+    private array $comments = [];
+
+    /**
+     * @Given I view the actions report section
+     */
+    public function iViewActionsSection()
+    {
+        $activeReportId = $this->loggedInUserDetails->getCurrentReportId();
+        $reportSectionUrl = sprintf(self::REPORT_SECTION_ENDPOINT, $activeReportId, 'actions');
+        $this->visitPath($reportSectionUrl);
+    }
+
+    /**
+     * @Given I view and start the actions report section
+     */
+    public function iViewAndStartActionsSection()
+    {
+        $this->iViewActionsSection();
+        $this->clickLink('Start actions');
+    }
+
+    /**
+     * @Given I choose no and save on financial decision actions section
+     */
+    public function iChooseNoOnFinancialDecisionActionsSection1()
+    {
+        $this->fillInActionsForm(
+            'no',
+            'action[doYouExpectFinancialDecisions]'
+        );
+    }
+
+    /**
+     * @Given I choose yes and save on financial decision actions section
+     */
+    public function iChooseYesOnFinancialDecisionActionsSection1()
+    {
+        $this->fillInActionsForm(
+            'yes',
+            'action[doYouExpectFinancialDecisions]',
+            'first comment',
+            'action[doYouExpectFinancialDecisionsDetails]'
+        );
+    }
+
+    /**
+     * @Given I choose no and save on concerns actions section
+     */
+    public function iChooseNoOnDoYouHaveConcernsActionsSection()
+    {
+        $this->fillInActionsForm(
+            'no',
+            'action[doYouHaveConcerns]',
+        );
+    }
+
+    /**
+     * @Given I choose yes and save on concerns actions section
+     */
+    public function iChooseYesOnDoYouHaveConcernsActionsSection()
+    {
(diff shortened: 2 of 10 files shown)


behat/tests/bootstrap/v2/common/IShouldBeOnTrait.php
```diff
@@ -63,11 +63,51 @@ public function iAmOnContactsAddAnotherPage()
...
+        $this->iAmOnPage('/report\/.*\/actions\/summary\?from\=last\-step$/');
+    }
+
+    /**
+     * @Then I should be on the actions report summary page
+     */
+    public function iAmOnActionsSummaryPage()
+    {
+        $this->iAmOnPage('/report\/.*\/actions\/summary/');
+    }
```
acsauk, line 96: Are these two essentially the same page or does having the query param change the content of the page?

behat/tests/bootstrap/v2/common/IShouldBeOnTrait.php
```diff
@@ -63,11 +63,51 @@ public function iAmOnContactsAddAnotherPage()
...
+
+    /**
+     * @Then I should be on the Lay main overview page
```
acsauk, line 99: Maybe:

`Then I should be on the Lay homepage`?

behat/tests/bootstrap/v2/reporting/sections/ActionsSectionTrait.php
```diff
@@ -0,0 +1,143 @@
...
+        );
+    }
+    public function fillInActionsForm($answer, $actionName, $comment = null, $commentName = null)
+    {
+        $this->selectOption($actionName, $answer);
+        if ($answer === "no") {
+            $this->answeredNo += 1;
+        } elseif ($answer === "yes") {
+            $this->answeredYes += 1;
+        }
```
acsauk, line 84: Love this - was trying to work out how to assert on multiple yes and no responses ⭐

behat/tests/bootstrap/v2/reporting/sections/ActionsSectionTrait.php
```diff
@@ -0,0 +1,143 @@
...
+        if ($comment != null) {
+            $this->fillField($commentName, $comment);
+            array_push($this->comments, $comment);
+        }
+        $this->pressButton('Save and continue');
+    }
+
+    /**
+     * @When I should see the expected action report section responses
```
acsauk, line 93: `@Then I should see...` might flow better here?

behat/tests/bootstrap/v2/reporting/sections/ActionsSectionTrait.php
```diff
@@ -0,0 +1,143 @@
...
+            assert(str_contains($table->getHtml(), $comment));
+        }
+    }
+
+    /**
+     * @Then I follow edit link on concerns question
+     */
+    public function iFollowEditLinkConcernActionsPage()
+    {
+        //this should be replaced with actual link click but could not identify it properly
```
acsauk, line 138: I'm assuming there was no id, unique class or label we could hook in to here? Something worth discussing at a higher level to see if we can make a change to always apply an id to links based on the element (or something along those lines)

behat/tests/bootstrap/v2/reporting/sections/ActionsSectionTrait.php
```diff
@@ -0,0 +1,143 @@
...
+        $countPositiveReponse = 0;
+        foreach ($tableEntry as $entry) {
+            if (trim(strtolower($entry->getHtml())) === "no") {
+                $countNegativeReponse += 1;
+            } elseif (strtolower(trim($entry->getHtml())) === "yes") {
+                $countPositiveReponse += 1;
+            }
+        }
+        assert($countNegativeReponse == $this->answeredNo);
+        assert($countPositiveReponse == $this->answeredYes);
```
acsauk, line 115: Could we get a bit of spacing between the blocks here? Makes things a little easier to read. Side note, we should agree on the rules for this and add to pre-commit hooks.

behat/tests/bootstrap/v2/reporting/sections/ReportingSectionsFeatureContext.php
```diff
@@ -54,4 +55,81 @@ public function nextSectionShouldBe(string $sectionName)
...
+                        $statusCorrect = true;
+                    }
+                }
+            }
+        }
+        if (!$statusCorrect) {
+            $this->throwContextualException(
+                sprintf('Report section status not as expected. Status: %s not found. ', $status)
+            );
+        }
```
acsauk, line 111: Same as above, could we get a bit of spacing between the blocks?

behat/tests/bootstrap/v2/reporting/sections/ReportingSectionsFeatureContext.php
```diff
@@ -54,4 +55,81 @@ public function nextSectionShouldBe(string $sectionName)
...
+        if (!$tableEntry) {
+            $this->throwContextualException('A dd element was not found on the page');
+        }
+        $furtherInfoNeeded = false;
+        foreach ($tableEntry as $entry) {
+            if (str_contains(trim(strtolower($entry->getHtml())), "please answer this question")) {
+                $furtherInfoNeeded = true;
+            }
+        }
+        assert($furtherInfoNeeded);
```
acsauk, line 133: Aaaand again 🙏

behat/tests/features-v2/reporting/sections/actions/actions.feature
```diff
@@ -0,0 +1,44 @@
+@v2
+Feature: Actions
+
+  Scenario: I complete the actions section with negative responses
```
acsauk, line 4: I get what this means but I think the scenario makes more sense in a second/third person perspective:

`A user has made no financial decisions and has no concerns`

We already know we're in the actions section as we've set the context for the person reading the test under the `Feature` header.

behat/tests/features-v2/reporting/sections/actions/actions.feature
```diff
@@ -0,0 +1,44 @@
...
+    When I choose no and save on financial decision actions section
+    Then I should be on the concerns actions page
+    When I choose no and save on concerns actions section
+    Then I should be on the actions report last step page
+    And I should see the expected action report section responses
+    When I follow link back to report overview page
+    Then I should be on the Lay reports overview page
+    And I should see "actions" as "finished"
+
+  Scenario: I complete the actions section with positive responses
```
acsauk, line 17: Similar to above, something like `A user has made financial decisions and has concerns` describes the feature a little clearer.

behat/tests/features-v2/reporting/sections/actions/actions.feature
```diff
@@ -0,0 +1,44 @@
...
+    And I view and start the actions report section
+    Then I should be on the financial decision actions page
+    When I choose yes and save on financial decision actions section
+    Then I should be on the concerns actions page
+    When I choose yes and save on concerns actions section
+    Then I should be on the actions report last step page
+    And I should see the expected action report section responses
+    And I should see the expected action comments
+
+  Scenario: I half fill in responses and come back and edit them
```
acsauk, line 28: I guess here the test is on editing existing responses so maybe something like `A user partially completes the section and then edits their responses` focuses the reader on what we're testing?

behat/tests/features-v2/reporting/sections/actions/actions.feature
```diff
@@ -0,0 +1,44 @@
...
+    And I press report sub section back button
+    Then I should see text asking to answer the question
+    When I follow link back to report overview page
+    Then I should see "actions" as "not finished"
+    When I view the actions report section
+    Then I should be on the actions report summary page
+    When I follow edit link on concerns question
+    And I choose no and save on concerns actions section
+    And I follow link back to report overview page
+    Then I should see "actions" as "finished"
```
acsauk, line 44: Nice - I missed this out on the recent scenario I did so will backfill once this is merged.

acsauk (changes requested):
Great solution to asserting on multiple of the same answer on pages 💪  couple of changes required on the scenario descriptions and some styling but looks like a solid bunch of tests!

I'll post a discussion in the dev channel now on what our preferred spacing is in the project so we can add a pre-commit hook in to reduce any future comments on spacing between blocks.

behat/tests/bootstrap/v2/common/IShouldBeOnTrait.php
jamesrwarren: Hmm yeah that sounds nicer

behat/tests/bootstrap/v2/reporting/sections/ActionsSectionTrait.php
jamesrwarren: I think we can actually do this better. I didn't want to waste further time in this ticket but I've got some improvements to try on next one of these.

behat/tests/bootstrap/v2/reporting/sections/ActionsSectionTrait.php
jamesrwarren: Good point. I don't even know why we have to put the tags in the actual codey bit as they can be used in the behatty bit as Ands, Thens or whatever

behat/tests/bootstrap/v2/reporting/sections/ActionsSectionTrait.php
jamesrwarren: yeah nothing specific. the attribute of the href.. but then that would still be doing a visitpath instead of a clicklink. So yeah maybe we should apply an id as you say to these.

behat/tests/bootstrap/v2/common/IShouldBeOnTrait.php
jamesrwarren: yeah i suppose they do seem to be. Merging to one

acsauk (approved):
:shipit:

Taking another bite out of the behat apple 😎

PR #621, opened 2021-04-07
DDPB-3476: Behat report sections - Documents

acsauk:
## Purpose
Covers off the documents section of the report in new style behat tests.

Fixes DDPB-3476.

## Approach
I've really doubled down on trying to make the BDD as understandable and domain-specific as possible in this ticket. Hopefully, I've got the balance right between step naming and asserting in the context files but wide open to feedback if things could be clearer.

I've also added our first professional fixture users in to pass a test on org user specific alerting. I've only used them to log in so far so we could find things aren't quite right when we come to use them for more involved tests but there's a basis to build all other org fixture users out from. We should also keep an eye on the time it takes to build out the fixtures between each scenario. I'm kind of on the fence with at the end of the exercise just outputting some SQL that can be executed if the run time starts to creep up too high but I don't think we're there yet.

## Learning
Used an ifdef in the makefile to allow passing command line vars to make commands without having to have a new make command 🥳 

## Checklist

### DIFF
--- a/api/src/TestHelpers/BehatFixtures.php
+++ b/api/src/TestHelpers/BehatFixtures.php
@@ -227,7 +227,7 @@ class BehatFixtures
 
     private function addClientsAndReportsToLayDeputy(User $deputy, bool $completed = false, bool $submitted = false)
     {
-        $client = $this->clientTestHelper->createClient($this->entityManager, $deputy);
+        $client = $this->clientTestHelper->generateClient($this->entityManager, $deputy);
         $report = $this->reportTestHelper->generateReport($this->entityManager, $client);
 
         $client->addReport($report);
@@ -248,7 +248,7 @@ class BehatFixtures
 
     private function addOrgClientsAndReportsToOrgDeputy(User $deputy, Organisation $organisation, bool $completed = false, bool $submitted = false)
     {
-        $client = $this->clientTestHelper->createClient($this->entityManager, $deputy, $organisation);
+        $client = $this->clientTestHelper->generateClient($this->entityManager, $deputy, $organisation);
         $report = $this->reportTestHelper->generateReport($this->entityManager, $client);
 
         $client->addReport($report);
--- a/api/src/TestHelpers/BehatFixtures.php
+++ b/api/src/TestHelpers/BehatFixtures.php
@@ -3,13 +3,14 @@
 
 namespace App\TestHelpers;
 
+use App\Entity\Organisation;
 use App\Entity\User;
 use Doctrine\Common\DataFixtures\Purger\ORMPurger;
 use Doctrine\ORM\EntityManagerInterface;
 use Exception;
 use Symfony\Component\Security\Core\Encoder\UserPasswordEncoderInterface;
 
-// Not extending AbstractDataFixture so we can use this in test runs rather commands
+// Not extending AbstractDataFixture so we can use this in test runs rather than just commands
 class BehatFixtures
 {
     private EntityManagerInterface $entityManager;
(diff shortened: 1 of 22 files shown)


behat/tests/bootstrap/v2/common/AlertsTrait.php
```diff
@@ -0,0 +1,30 @@
+<?php declare(strict_types=1);
+
+namespace DigidepsBehat\v2\Common;
+
+trait AlertsTrait
+{
+    public function assertOnAlertMessage(string $alertMessage)
```
jamesrwarren, line 7: Ah this is cool :)

api/src/TestHelpers/BehatFixtures.php
```diff
@@ -137,21 +184,48 @@ private function createAdminUsers()
...
+        $this->addClientsAndReportsToLayDeputy($this->layCompleted, true, false);
 
         $this->laySubmitted = $this->userTestHelper
-            ->createUser(null, User::ROLE_LAY_DEPUTY, sprintf('lay-submitted-%s@publicguardian.gov.uk', $this->testRunId));
-        $this->addClientsAndReportsToDeputy($this->laySubmitted, true, true);
+            ->createUser(null, User::ROLE_LAY_DEPUTY, sprintf('lay-submitted-%s@t.uk', $this->testRunId));
+        $this->addClientsAndReportsToLayDeputy($this->laySubmitted, true, true);
+    }
+
+    private function createProfs()
```
jamesrwarren, line 207: So glad you've given us this fixture setup to make use of as I would have found this challenging. Great stuff :)

behat/tests/features-v2/reporting/sections/documents/documents.feature
```diff
@@ -0,0 +1,74 @@
...
+    Then I should be on the documents summary page
+    When I remove one document I uploaded
+    Then the documents summary page should contain the documents I uploaded
+
+  Scenario: A user uploads one supporting document that has an invalid file type
+    Given a Lay Deputy has not started a report
+    When I view and start the documents report section
+    And I have documents to upload
+    And I upload one document with an unsupported file type
+    Then I should see an 'invalid file type' error
```
jamesrwarren, line 56: Not sure if you want to check the page you're on for the errors too or if that's overkill

behat/tests/features-v2/reporting/sections/documents/documents.feature
```diff
@@ -0,0 +1,74 @@
+@v2
```
jamesrwarren, line 1: I have been adding an additional tag here for the individual feature in addition to the v2 tag for ease of  manual running. Totally not necessary but either I should remove the tags or you add them. Up to you :)

jamesrwarren (approved):
All looks good to me. I've added some comments but none that specifically need changes I don't think.

coderste (dismissed):

behat/tests/features-v2/reporting/sections/documents/documents.feature
acsauk: Ah yeah thats a fair point - I'll update 👍

behat/tests/features-v2/reporting/sections/documents/documents.feature
acsauk: Mmmm I think in this instance the focus of the test is more so on ensuring an error is visible rather than the flow of the user journey leading to a certain page (or not in this case). Definitely for other features where we expect to land on a specific page after an action it makes sense to assert on that as part of the BDD.

api/src/TestHelpers/BehatFixtures.php
acsauk: Haha I wasn't looking forward to it and it was a lot of trial and error. Like I said in the PR comments, there may be some of the setup which is quite right when we come to do more org specific pages but at least we have a starting user now.

jamesrwarren (approved):

PR #626, opened 2021-04-16
DDPB-3956: Remove final traces of teams and clear up team user client associations

acsauk:
## Purpose
Getting rid of the team related tables and removing associations between clients and non-lay deputies.

Fixes DDPB-3956

## Approach
We were alerted to a case of org members still seeing a client list even after being removed from the org. After further investigation, we found that the old mechanism for team members to view clients/reports was still in place (essentially using the same mechanism as lay deputies by associating the client with the team user in the deputy_case table).

We've already run the client association removal code on prod so this is just recording the fact in source control but the migration included here also drops the dd_team and user_team tables and constraints. 

## Checklist

### DIFF
--- a/api/src/Migrations/Version250.php
+++ b/api/src/Migrations/Version250.php
@@ -1,0 +1,36 @@
+<?php
+
+declare(strict_types=1);
+
+namespace DoctrineMigrations;
+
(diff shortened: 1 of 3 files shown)


jamesrwarren (approved):
Awesome! Good to get rid of this!

PR #628, opened 2021-04-16
Ddpb 3954 - add basic browserstack check for v2 JS issue

jamesrwarren:
## Purpose
A minimum set of tests we can run against browserstack. This PR includes the following:

- Entries for 3 common browsers through browserstack and a feature that tests a known path that broke because of JS issues in IE11
- A new way of doing asserts
- Removed asserts from iAmOn type checks as they do an error with context as part of them anway
- A README that gives a good explanation of current Browserstack usage 
- Brought in some useful helper functions that are needed for selenium to work on IE11

If we force a failure on assert the new assert fail format looks like this:

image

Fixes DDPB-3954

## Approach
Implement above features

## Learning
Resolution, max screen size and java helper functions for element selection are all needed tools for getting anything to run properly on IE11 and to a certain extent firefox with selenium. Spent a long time being bad as JS before realising someone some time in the distant past had solved this issue so copied in their solution.

## Checklist

### Frontend


behat/tests/bootstrap/v2/reporting/sections/AccountsSectionTrait.php
```diff
@@ -319,24 +319,24 @@ public function iShouldSeeTheExpectedAccountsOnSummaryPage()
             $headHtml = trim(strtolower($tableHeader->getHtml()));
             assert(
                 str_contains($headHtml, $this->accountList[$tRowKey]['accountType']),
-                sprintf('matching account %s ', $this->accountList[$tRowKey]['accountType'])
+                $this->formatAssertResponse($this->accountList[$tRowKey]['accountType'], 'Not found', 'Accounts Type', $this->getCurrentUrl())
```
acsauk, line 322: I wonder if we could wrap assert itself in a helper function that does all of this? We have the same value passed to `str_contains()` and `formatAssertResponse()` and $this->getCurrentUrl() could just be called in the function rather than passed as an argument leaving you with:

- HTML to match against,
- Value to match against
- Found (or not found) value (could leave this as a null field at the end so null values default to 'Not found')
- Name of element matching against ($comparisonSubject)

behat/tests/bootstrap/v2/reporting/sections/VisitsAndCareSectionTrait.php
```diff
@@ -0,0 +1,44 @@
...
+    /**
+     * @When I enter that I do not live with client
+     */
+    public function iDoNotLiveWithClientVisitsAndCareSection()
+    {
+        $this->iFillFieldForCrossBrowser('visits_care_doYouLiveWithClient_1', 'no');
+    }
+
+    /**
+     * @When I can fill in a text box with how often I visit client
```
acsauk, line 36: Maybe make this a tiny bit more explicit about expecting to see the box (I know its implied but just so its clear for other readers):

`@Then I can see and fill in a text box with how often I visit client`

behat/tests/features-v2/cross-browser/js-basic-test.feature
```diff
@@ -0,0 +1,8 @@
+@v2 @js-basic-check
```
acsauk, line 1: Could we get an extra `@js` tag here too? There are some built in behat features to do with selecting drivers we may want to hook in to in the future that relies on a plain ole `@js` tag to work.

docs/CROSS_BROWSER_TESTS.md
```diff
@@ -0,0 +1,50 @@
...
+
+All the profiles are controlled form the `behat.yml` file as usual and more can be added as needed.
+
+
+## Pipeline
+
+We have a basic setup in our pipeline though we are using the V2 rewrite to make something more concrete that we can use.
+More on this soon...
+
+[browser-stack]: https://www.browserstack.com/browser
```
acsauk, line 50: ❤️

acsauk (commented):
Potential suggestion around further streamlining the assert stuff but I'm open to being convinced otherwise!

behat/tests/bootstrap/v2/reporting/sections/AccountsSectionTrait.php
jamesrwarren: Ok I've done this. It does make it slightly less flexible but it looks a bit neater.

behat/tests/bootstrap/v2/reporting/sections/VisitsAndCareSectionTrait.php
jamesrwarren: Done

behat/tests/features-v2/cross-browser/js-basic-test.feature
jamesrwarren: Done

behat/tests/bootstrap/v2/common/DebugTrait.php
```diff
@@ -77,12 +77,11 @@ public function debug($name)
         echo "$file";
     }
 
-
-    public function formatAssertResponse(
-        $expecting,
+    public function bespokeAssert(
```
acsauk, line 80: The name of this function could be more explicit about its intended use. To me, it seems to be narrowly focused on if a string appears in another string so something like `assertStringContainsString` could be clearer for future function callers.

behat/tests/bootstrap/v2/common/DebugTrait.php
```diff
@@ -77,12 +77,11 @@ public function debug($name)
...
+        $expected,
         $found,
         string $comparisonSubject,
-        string $url
+        bool $exactMatch
```
acsauk, line 84: This feels like it should be broken out to its own function rather than adding a conditional to this one to keep them focused on a single responsibility. Similar to the above comment, something like `assertStringsAreEqual()` with the same error formatting could work?

Taking that into account maybe all of this is better off in an `AssertTrait`?

acsauk (commented):
Couple of suggestions on trying to stick with SRP for the assert function as well as where the code should live now its focused on asserting rather than debugging an error out.

behat/tests/bootstrap/v2/common/DebugTrait.php
jamesrwarren: hehe ok, will go with that

behat/tests/bootstrap/v2/common/DebugTrait.php
jamesrwarren: done

behat/tests/bootstrap/v2/common/AssertTrait.php
```diff
@@ -0,0 +1,58 @@
...
+
+        return sprintf(
+            $message,
+            $expected,
+            $found,
+            $comparisonSubject,
+            $this->getCurrentUrl()
+        );
+    }
+}
```
acsauk, line 58: 😍

acsauk (approved):
:shipit: excellent - thanks for bearing with me there 😁

PR #629, opened 2021-04-16
DDPB-3961: Add PHPStan to pre-commit

acsauk:
## Purpose
Adding PHPStan to pre-commit hooks to ensure any new code we write has some base level of static analysis applied. 

I've also formalised running PHPStan under a couple of makefile commands so we can get depressed about the thousands of errors its spits out. At some point, it would be nice to tackle them level by level but for now, having the new code we've written checked is a good first step.

Fixes DDPB-3961.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -21,6 +21,8 @@ repos:
     hooks:
     - id: php-cs-fixer
       files: \.(php)$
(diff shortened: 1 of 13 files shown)


jamesrwarren (approved):
I'm assuming vendor/bin is the official path struct for this, in which case tick tick :)

PR #630, opened 2021-04-19
[DDPB-3481] added behat tests for money in high assets accounts

coderste:
This adds the integration tests for money in section of the report when
the client has a high asset account

Also fixed a bug where the £{{ limit }} was being printed like so rather
than being printed with the value inside. This was happening across
multiple pages

Fixes DDPB-3481


acsauk (changes requested):
Added some comments on making the BDD a bit more user/behaviour driven rather than being focused on the form. If you need a hand with setting up the asserting on expected values on the summary page then give me a shout and we can pair up for a bit 👍 

We probably also want some tests around:
- Adding an item of income from the summary page
- Editing an item of income from the summary page

And one last thought - seeing as we'll want to ensure we don't get the options for low assets we can follow the pattern for filenaming feature files where we separate a differentiator using a `.` so the feature file would become:

`money-in.high-assets.feature`

Ready to have:

`money-in.low-assets.feature`

on a future ticket.

behat/tests/features-v2/reporting/sections/money-in-high-assets/money-in.high-assets.feature
```diff
@@ -0,0 +1,60 @@
...
+    When I edit the money in value
+    Then the money in summary page should contain the edited value
+
+  Scenario: A user submits multiple items of income forms with valid values
+    Given a Lay Deputy has not started a report
+    When I view and start the money in report section
+    And I have a dividend to report on
+    And I enter a valid amount
+    When I add another item
+    And I select state pension
```
acsauk, line 48: Following the same pattern as dividends, this could be `And I have a state pension to report on`

acsauk (changes requested):
Looking great, one tiny change and then this is ready to go 🚀

acsauk:
Oh sorry just realised as well, we probably want to cover off adding an income item via the summary page. Can see you covered off editing from the summary page already but it doesn't look like there's an add there.

behat/tests/features-v2/reporting/sections/money-in-high-assets/money-in.high-assets.feature
coderste: Doh! I dunno how I missed that 😂

acsauk (dismissed):
🚢   great job! We'll revisit the summary assertions once we have a generalised way of asserting on lists of items on summary pages.

acsauk (approved):
:shipit:

PR #632, opened 2021-04-26
DDPB-3962: Remove whitespace between filename and extension during document syncing

acsauk:
## Purpose
Removes whiteapace during document syncing process to take into account users who uploaded documents for a report prior to releasing the code that sanitises the filename on document upload.

Fixes DDPB-3962

## Approach
Unfortunately, we’re only able to do half of this ticket. Being able to add a missing file extension requires the actual file the user has uploaded. We made the switch from sending the document data to the S3 reference a while back to in order to add missing file extensions we’d need to revert to pulling the file down from S3 which wouldn’t be a good move (mainly as this was leading to some out of memory errors for larger files).

To at least get some value out of the ticket I've ensured there is a clear error message showing the file needs to be manually synced to Sirius if there is a missing file extension.

## Checklist

### Frontend

### DIFF
--- a/client/src/Service/DocumentSyncService.php
+++ b/client/src/Service/DocumentSyncService.php
@@ -1,25 +1,25 @@
-<?php declare(strict_types=1);
+<?php
 
+declare(strict_types=1);
 
 namespace App\Service;
 
 use App\Entity\Report\Document;
 use App\Entity\Report\Report;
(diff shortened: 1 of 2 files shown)


Gugandeep (approved):
Looks good to me.

Shame we couldn't also do the missing file extension fix, but at least having the error message for those without an extension will be useful.

PR #635, opened 2021-04-29
DDPB-3957: Move behat to API container

acsauk:
## Purpose
Behat is currently in its own container which is great for separation of concerns but not so great for writing concise or easy to follow code for our integration tests. Whenever we want to do anything fixture related we need to:

- Write a behat function to call out to an admin endpoint
- Write an admin endpoint to call out to an API endpoint
- Write an API endpoint to make a callout to a service 
- Add a function to a service to carry out the fixture request

This makes the flow of the code hard to follow and brittle for any future changes to controllers in the frontend/admin and API. It also means we have behat related code in three separate parts of our repo. By moving behat to the API container we consolidate the code in one place and then only need to:

- Add a function to a service to carry out the fixture request and call from a behat step

Fixes DDPB-3957

## Approach
I've replaced the two new endpoints we made as part of our recent behat rewrite with one call to a behat fixtures service. We only install the behat dependencies as part of our dev app environment so we shouldn't have any extra bloat in prod.

To standardise our namespacing I've made the following changes in the API app:

- Behat tests are now under `/tests/Behat` using the namespace `App\Tests\Behat`
- Unit tests are now under `/tests/Unit` using the namespace `App\Tests\Unit`

## Checklist


api/Dockerfile
```diff
@@ -90,6 +90,15 @@ COPY api.env api.env
 COPY postgres.env postgres.env
 COPY phpstan.neon .
 
+# Behat
+RUN mkdir -p /tmp/html
+RUN mkdir -p /tmp/sql
+
+RUN ["chown", "www-data", "/var/www/tests/Behat/run-smoke-tests.sh"]
```
jamesrwarren, line 97: Nothing wrong with doing it this way but maybe a bit inconsistent doing the array type commands here but just standard spaces elsewhere? :)

api/tests/Behat/bootstrap/Common/AuthenticationTrait.php
```diff
@@ -46,15 +46,15 @@ public function iAmAtAdminLogin()
...
         $this->iAmLoggedInToAdminAsWithPassword('admin@publicguardian.gov.uk', 'DigidepsPass1234');
         $this->visitAdminPath($url);
         $token = $this->getSession()->getPage()->getContent();
         $this->visitAdminPath('/logout');
 
-        $page = $pageType === 'activation' ? 'activate' : 'password-reset';
+        $page = 'activation' === $pageType ? 'activate' : 'password-reset';
 
-        if ($admin === '' || $admin === false) {
+        if ('' === $admin || false === $admin) {
```
jamesrwarren, line 57: I know this is the auto formatting but I find this totally not how my brain works hehe.

api/tests/Behat/bootstrap/v2/Common/BaseFeatureContext.php
```diff
@@ -29,7 +31,7 @@ class BaseFeatureContext extends MinkContext
 
     public const BEHAT_FRONT_RESET_FIXTURES = '/behat/frontend/reset-fixtures?testRunId=%s';
     public const BEHAT_FRONT_USER_DETAILS = '/behat/frontend/user/%s/details';
-    public const REPORT_SECTION_ENDPOINT = '%s/%s/%s';
+    public const REPORT_SECTION_ENDPOINT = '/%s/%s/%s';
```
jamesrwarren, line 34: Was a bit weary about this looks ok as handled by mink which can do either. 👍

api/tests/Behat/features/deputy/03-report/01-102/08-documents.feature
```diff
@@ -38,7 +38,7 @@ Feature: Report documents
       | report_document_upload_files   |
 
     # check vba-eicar file error
-    When I attach the file "pdf-doc-vba-eicar-dropper.pdf" to "report_document_upload_files"
+    When I attach the file "malicious-vba.pdf" to "report_document_upload_files"
```
jamesrwarren, line 41: Hehe. Remember this ;)

docker-compose.yml
```diff
@@ -126,6 +126,7 @@ services:
             APP_DEBUG: ${APP_DEBUG:-0}
         env_file:
             - ./api/api.env
+            - ./api/tests/Behat/test.env
```
jamesrwarren, line 129: Is this needed as, aren't you still using the test container for tests but just with the context of API? Or maybe these env vars are needed for API container to run?

composer.json
```diff
@@ -0,0 +1,2 @@
+{
```
jamesrwarren, line 1: is this supposed to be here?

jamesrwarren (commented):
Awesome stuff. Couple of questions. Otherwise looks good. Will approve once rebased etc as discussed.

api/Dockerfile
acsauk: Yeah good point - it was a copy-paste from the other Dockerfile. I shall update 👍

api/tests/Behat/bootstrap/v2/Common/BaseFeatureContext.php
acsauk: Yeah I had a failing test for some reason due to the missing first slash which is odd as it must have passed for you?

api/tests/Behat/features/deputy/03-report/01-102/08-documents.feature
acsauk: 🔪

docker-compose.yml
acsauk: I had originally dropped this but added them back in desperation during the file scan fiasco. I'll experiment with dropping them both out.

composer.json
acsauk: Oops, no thats me trying to install stuff in the root. I'll drop.

docker-compose.yml
acsauk: Yep works fine - removed.

jamesrwarren (approved):

PR #636, opened 2021-04-30
DDPB-3973 upgrade postgres part1

jamesrwarren:
## Purpose
Notified about security issue with postgres version. Part 1 is upgrade everything except prod. Check it's all dandy after the weekend then do another micro PR to upgrade prod.

Added create before delete on instances and removed the prevent destroy from the instances (NOT the cluster).
Based on documentation, data is stored at the cluster level and we can recreate instances with no issues (I've also done this on serve recently and it was fine). We have to remove the prevent destroy as tf wants to recreate the instances but not the cluster for the upgrade.

https://docs.aws.amazon.com/AmazonRDS/latest/AuroraUserGuide/USER_DeleteCluster.html

Fixes DDPB-3973

## Approach
NA

## Learning
NA
## Checklist

### Frontend


acsauk (approved):
:shipit:

PR #638, opened 2021-05-04
Ddpb 3973b

jamesrwarren:
## Purpose
part 2 update prod

Fixes DDPB-3973

## Approach
Turns out we really didn't need to do it this way as we can do targetted apply on cluster and let it upgrade overnight then push pipeline through the next day...

## Checklist

### Frontend


acsauk (approved):
:shipit:

PR #639, opened 2021-05-05
DDPB-3977 update composer to latest version

jamesrwarren:
## Purpose
No longer need to pin composer. Fix security vulnerability in composer.

Fixes DDPB-3977

## Approach
Needed extra argument to ignore platform req for php as otherwise it expects php8

## Checklist

### Frontend

### DIFF
--- a/api/Dockerfile
+++ b/api/Dockerfile
@@ -5,7 +5,7 @@ WORKDIR /app
 # Install composer dependencies
 COPY composer.json .
 COPY composer.lock .
-
+# Remove '--ignore-platform-req=php' when we move to php 8
 RUN composer install --prefer-dist --no-interaction --no-scripts --ignore-platform-req=php
 
 COPY app app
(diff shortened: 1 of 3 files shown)


api/Dockerfile
```diff
@@ -1,12 +1,12 @@
...
+FROM composer:2 AS composer
 
 WORKDIR /app
 
 # Install composer dependencies
 COPY composer.json .
 COPY composer.lock .
 
-RUN composer install --prefer-dist --no-interaction --no-scripts
+RUN composer install --prefer-dist --no-interaction --no-scripts --ignore-platform-req=php
```
acsauk, line 9: Can we add a comment here to remove `--ignore-platform-req=php` when we're on 8?

acsauk (commented):
One request 🙏

acsauk (approved):
:shipit:

PR #643, opened 2021-05-11
Ddpb 3980

jamesrwarren:
## Purpose
We seem to be reinventing the wheel on each of the summary assertions.

We want a unifying function that is fairly simple to use that still has enough functionality to assert that:

- Items are present
- Items are in the correct section
- Items are in the correct order

If we have an error we should have enough information to easily fix. When developing we should be able to easily see what it brings back on the page so we can easily assert against it.

Fixes DDPB-3980

## Approach
Found that the types of paths that the page takes could be broken down to a selection of if statements. This way we can handle any of the known existing page layouts.

Broke page layouts down to what could be termed rows which are represented by an array of arrays (which are the fields in the rows) and we can compare them against what we are expecting.

There will be extra 'fields' when bringing things back this way so to find things in the right order we do a fun thing slicing the array to make sure were always searching forward through the elements...

## Learning
NA

## Checklist

### Frontend


api/tests/Behat/bootstrap/v2/Common/ExpectedResultsTrait.php
```diff
@@ -0,0 +1,158 @@
...
+                    $listItems = $tableItem->findAll('xpath', $xpath);
+
+                    if (count($listItems) > 0) {
+                        $this->getTextItems($listItems);
+                    } else {
+                        $xpath = '//dt|//dd';
+                        $descriptionDataItems = $tableItem->findAll('xpath', $xpath);
+                        $this->getTextItems($descriptionDataItems);
+                    }
+                }
```
acsauk, line 45: As we're doing what looks like the same actions regardless of if `$listSummaryRowItems` are more than 0 or not, it looks like this can be collapsed down to:

```
            $xpath = '//li';
            $listItems = $tableItem->findAll('xpath', $xpath);
            
            if (count($listItems) > 0) {
                $this->getTextItems($listItems);
            } else {
                $xpath = '//dt|//dd';
                $descriptionDataItems = $tableItem->findAll('xpath', $xpath);
                $this->getTextItems($descriptionDataItems);
            }
```

Unless I'm missing something subtle?

api/tests/Behat/bootstrap/v2/Common/ExpectedResultsTrait.php
```diff
@@ -0,0 +1,158 @@
...
+
+    public function expectedResultsDisplayed($sectionNumber, $expectedResultArrays, $context)
+    {
+        $xpath = '//dl|//tbody';
+        $topLevelTables = $this->getSession()->getPage()->findAll('xpath', $xpath);
+
+        $sections = [];
+        foreach ($topLevelTables as $tableItem) {
+            $this->comparisonItems = [];
+            if ('dl' == $tableItem->getTagName()) {
```
acsauk, line 19: This section may be more readable if we extract the different responsibilities out, something like:

```
        foreach ($topLevelTables as $tableItem) {
            $this->comparisonItems = [];
            
            $this->validateTableItem($tableItem);
            $this->handleDescriptionList($tableItem);
            $this->handleTableBody($tableItem);
            
            $sections[] = $this->comparisonItems;
        }
```

```
    private function validateTableItem(NodeElement $tableItem)
    {
        $htmlTag = $tableItem->getTagName();
        
        if (!in_array($htmlTag, ['tbody', 'dl'])) {
            $this->throwContextualException('The element being asserted on does not contain a dl or tbody element');
        }
    }
```

```
    private function handleDescriptionList(NodeElement $tableItem)
    {
        if ('dl' == $tableItem->getTagName()) {
            $xpath = '//li';
            $listItems = $tableItem->findAll('xpath', $xpath);
            
            if (count($listItems) > 0) {
                $this->getTextItems($listItems);
            } else {
                $xpath = '//dt|//dd';
                $descriptionDataItems = $tableItem->findAll('xpath', $xpath);
                $this->getTextItems($descriptionDataItems);
            }
        }
    }
```

```
    private function handleTableBody(NodeElement $tableItem)
    {
        if ('tbody' == $tableItem->getTagName()) {
            $xpath = '//tr';
            $tableRowItems = $tableItem->findAll('xpath', $xpath);
    
            if (count($tableRowItems) > 0) {
                foreach ($tableRowItems as $tableRowItem) {
                    $xpath = '//td|//th';
                    $tableDataItems = $tableRowItem->findAll('xpath', $xpath);
                    $this->getTextItems($tableDataItems);
                }
            } else {
                $xpath = '//td|//th';
                $tableDataItems = $tableItem->findAll('xpath', $xpath);
                $this->getTextItems($tableDataItems);
            }
        }
    }
```

api/tests/Behat/bootstrap/v2/Common/ExpectedResultsTrait.php
```diff
@@ -0,0 +1,158 @@
...
+                    $found = true;
+                    $foundInElem = $foundItemKey + $foundInElemPrevious;
+                    break;
+                }
+            }
+            if ($found and $foundInElem >= $foundInElemPrevious) {
+                $foundInElemPrevious = $foundInElem;
+            } else {
+                $raiseException = true;
+            }
```
acsauk, line 105: Finding this quite hard to follow - could you take me through it?

acsauk (commented):
Looks like there are some really cool things happening here but I'm finding it a little hard to follow. Could we catch up when you're next free to talk through intent and what this is testing? I'm not sure we need to be tying ourselves to the order that items appear on the screen as this is likely going to make the tests brittle and prone to update when we make small template updates (although happy to be proven wrong if item order affects how the app works or is used)

api/tests/Behat/bootstrap/v2/Common/ExpectedResultsTrait.php
jamesrwarren: To be honest it was hard to implement lol. Broke my brain a bit... I'll talk it through with you

api/tests/Behat/bootstrap/v2/Common/ExpectedResultsTrait.php
```diff
@@ -155,4 +185,72 @@ private function raiseExpectedResultsException($expectedItems, $foundItems, $sec
...
+    }
+
+    private function checkExpectedResultsCorrectFormat($expectedResults)
+    {
+        foreach ($expectedResults as $expectedResult) {
+            if (!is_array($expectedResult)) {
+                $this->throwContextualException(
+                    "\nIncorrect data types - \nexpectedResults must be an array of array of strings to represent rows and fields on summary page"
+                );
+            }
```
acsauk, line 253: Depends on how far we want to go but you could also check all of the elements are strings here too as the error message makes it sounds like we are already checking this. Another `foreach` and `is_string()` would do it.

api/tests/Behat/bootstrap/v2/Common/ExpectedResultsTrait.php
```diff
@@ -6,83 +6,113 @@
...
+                $this->updateSummarySectionItemsFound($tableDataItems);
+            }
+        } else {
+            $xpath = '//td|//th';
+            $tableDataItems = $table->findAll('xpath', $xpath);
+            $this->updateSummarySectionItemsFound($tableDataItems);
+        }
+    }
+
+    private function updateSummarySectionItemsFound($items)
```
acsauk, line 109: We're never changing anything in `summarySectionItemsFound` in this function so maybe `addToSummarySectionItemsFound ` is clearer?

api/tests/Behat/bootstrap/v2/Common/ExpectedResultsTrait.php
```diff
@@ -6,83 +6,113 @@
...
+        $listItems = $descriptionList->findAll('xpath', $xpath);
+
+        if (count($listItems) > 0) {
+            $this->updateSummarySectionItemsFound($listItems);
+        } else {
+            $xpath = '//dt|//dd';
+            $descriptionDataItems = $descriptionList->findAll('xpath', $xpath);
+            $this->updateSummarySectionItemsFound($descriptionDataItems);
+        }
+    }
```
acsauk, line 77: Aren't these doing the same thing?

acsauk (commented):
Looking good - definitely easier to follow and the var naming has made things clearer. A couple of comments on some refactoring and validation checking but looks almost ready to merge 👍

api/tests/Behat/bootstrap/v2/Common/ExpectedResultsTrait.php
jamesrwarren: they weren't when i first set it up and i seem to have made them do the same. it will actually allow more fine grained control if i split them out properly so done that :)

api/tests/Behat/bootstrap/v2/Common/ExpectedResultsTrait.php
jamesrwarren: good point. done

api/tests/Behat/bootstrap/v2/Common/ExpectedResultsTrait.php
jamesrwarren: yep done

api/tests/Behat/bootstrap/v2/Reporting/Sections/ActionsSectionTrait.php
```diff
@@ -80,13 +80,15 @@ public function iChooseYesOnDoYouHaveConcernsActionsSection()
...
+        $this->commentsAndAnswers[] = [$answer];
 
-        $this->commentsAndAnswers[] = $answer;
         if (null != $comment) {
             $this->fillField($commentName, $comment);
-            $this->commentsAndAnswers[] = $comment;
+            $this->commentsAndAnswers[] = [$comment];
         }
 
+//        $this->commentsAndAnswers[] = $commentAndAnswer;
```
acsauk, line 90: Do we still need this?

acsauk (commented):
One potential commented out line to drop but other than that its ready to :shipit: 

Let the experiment begin!

acsauk (approved):
:shipit:

PR #644, opened 2021-05-11
DDPB-3982 fix bug with terraform apply

jamesrwarren:
## Purpose
Because the clamav docker provider didn't previously version their images we ended up using our own image which is now quite out of date and seems to be falling over on initial build with a OOM exception.

Tried with the latest fixed version (yes they seem to now be versioning) of the container and it seems to work without falling over on build. yay.

Also fixed a small path error that was missed with the behat move.

Fixes DDPB-3982

## Approach
NA

## Learning
NA
## Checklist

### Frontend

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -153,7 +153,7 @@ services:
         restart: always
 
     file-scanner-server:
-        image: mkodockx/docker-clamav:alpine
+        image: mkodockx/docker-clamav:1.0.1-alpine
         environment:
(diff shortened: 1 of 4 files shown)


acsauk (changes requested):
Just need to drop the Dockerfile from behat then ✅

acsauk (approved):
And I should read PRs more carefully 😂 

:shipit:

PR #647, opened 2021-05-17
Apply PSR12

acsauk:
## Purpose
FInally got sick of seeing PSR12 based changes in my PRs so I've applied the standard against API and client. This should cut down on some of the noise in our PRs 🙏 

## Approach
Ran `phpcbf --standard=PSR12 ./client/src` and `phpcbf --standard=PSR12 ./api/src` 🤓


Gugandeep (approved):
These should make future PRs a lot more easier to read through.

PR #648, opened 2021-05-18
Ddpb 3485 - lifestyle behat

jamesrwarren:
## Purpose
Added the health and welfare lifestyle section. Checked it doesn't appear on NDR or PFA.

image

Fixes DDPB-3485

## Approach
Used the expected results function again. Seemed to save me a lot of time. Think we could maybe genericise some other bits like the  'I Visit' bits but I didn't do much refactoring in this one. Kept it simple and followed existing patterns.

## Learning
I added health and welfare types in... probably could have done all the others I know we'll need but I was just trying to get this small pr done and dusted.. Soz

## Checklist

### Frontend


acsauk (commented):
All looks good - I think we should add one further scenario for when a user answers both questions and then edits the initial question to no so the form gets cleared. Ready to merge after that 👍

jamesrwarren:
Added the extra test for editing the second answer to no. There isn't actually a yes/no for the first question, just different text which we were changing anyway. I added the extra step to change second from yes to no anyway :)

acsauk (approved):
:shipit:

PR #650, opened 2021-05-24
DDPB-3486: Deputy expenses section (and alternative proposal for asserting on summary pages)

acsauk:
## Purpose
Covers the deputy expenses section of the report.

Fixes DDPB-3486

## Approach
The section itself was pretty straightforward but I wanted to build on what Jim put in place to see if I could make a slightly simplified version. The main difference I went with was tying in filling in the form values with a function to also update the array of form values required to assert against. This feels a bit more user friendly and should hopefully remove some of the learning curve using the assertion code. I also went with allowing the person writing the tests to name the section how they please rather than having to keep track of a section number in their code with the current approach.

Additionally, I decided to drop checking the order of things on summary pages. I know @jamesrwarren was keen on this but I think for the added complexity of the code the payoff wasn't justified as the likelihood of us both mixing up things in templates and with UAT not picking it up seemed pretty low. There are some alternative approaches we could take here to do proper UI testing where we have some golden files of expected responses that we run a quick suite against to give us some extra confidence, but also open to being convinced on this point 👍 .

One part that did become a little more complex than I wanted to was grouping responses together for what we're calling a row in the summary overview. As we usually want to test removing a response a user has provided this presented an issue as generally we'll want to remove all values the user entered for the response rather than one filed, e.g. removing `expenseDetails[0]` vs. removing `expenseDetails[0]['expenses_single[explanation]']`:

```
array(2) {
      │   ["anyExpensesClaimed"]=>
      │   array(1) {
      │     [0]=>
      │     array(1) {
      │       ["yes_no[paidForAnything]"]=>
      │       string(3) "yes"
      │     }
      │   }
      │   ["expenseDetails"]=>
      │   array(2) {
      │     [0]=>
      │     array(2) {
      │       ["expenses_single[explanation]"]=>
      │       string(94) "Fugit sit nemo sit quia aspernatur eligendi soluta cumque perferendis deserunt incidunt autem."
      │       ["expenses_single[amount]"]=>
      │       int(789)
      │     }
      │     [1]=>
      │     array(2) {
      │       ["expenses_single[explanation]"]=>
      │       string(90) "Earum possimus qui ut inventore tempora ratione voluptatem aut perferendis illo vitae eum."
      │       ["expenses_single[amount]"]=>
      │       int(782)
      │     }
      │   }
      │ }
```

As you can see above, I had to group associated responses together for easy alterations in other parts of the code which means the person using the interface needs to know which group of responses they want to add a field response to. Previously this was a flat array that looked like:

```
["expenseDetails"]=>
      │   array(4) {
      │     [0]=>
      │     array(1) {
      │       ["expenses_single[explanation]"]=>
      │       string(85) "Quidem maxime nesciunt dicta blanditiis voluptatem saepe rerum impedit voluptatem et."
      │     }
      │     [1]=>
      │     array(1) {
      │       ["expenses_single[amount]"]=>
      │       int(694)
      │     }
      │     [2]=>
      │     array(1) {
      │       ["expenses_single[explanation]"]=>
      │       string(68) "Consequatur et odio consectetur ut nihil blanditiis inventore vitae."
      │     }
      │     [3]=>
      │     array(1) {
      │       ["expenses_single[amount]"]=>
      │       int(631)
      │     }
      │   }
      │ }
```

But as there was nothing to tie associated answers together it made removing items for the array near impossible (or very ugly with nested loops/6 variable function calls etc).

This seemed like an acceptable tradeoff for me to keep the interface minimal but if anyone else has any other idea to handle it then let me know in the comments below.

## Checklist


api/tests/Behat/bootstrap/v2/Common/FormFillingTrait.php
```diff
@@ -0,0 +1,304 @@
+<?php
```
coderste, line 1: Nice job on this 👏🏻 will definitely save a lot of time when writing these tests now not having to duplicate form fields

coderste (approved):

coderste (dismissed):

jamesrwarren (approved):
After all the discussions in chat, looks good. Let's ship this bad boy :)

PR #653, opened 2021-05-27
Ddpb 3995

jamesrwarren:
## Purpose
Currently our new behat tests are fast but how fast can we make them? Can we parallise with a big 4 core behat container and get to that sweet sub 2 min nirvana? Annoyingly this takes 1m30 locally but it seems a lot slower here. Still under 5 mins is pretty good right!

image

In another PR i'm 3999, i'm going to scrape out the build redeployer to give us another 20 secs off. When we have the admin tests in it might be worth splitting them into a separate container that runs in parallel and have @v2 and @v2-admin or something as tags.

Also let's get rid of separate step for the reset. Let's do a reset in the same process to save on time before we run our parallel processes! 

Fixes DDPB-3995

## Approach
Removed all the setup that blanket creates users for each scenario as we're not going to be resetting the database so we don't want it getting bloated between builds with unused users. This should also be marginally faster.

Then used a php package called fastest that can parallelise anything. It takes files passed into it and runs processes in parallel in the background.

Because this will be processed as separate behat processes, it means we need a separate reset db command that runs before. This is fine though as we were using a shell script anyway so lets just chuck that step in beforehand.

Finally I did spend quite a bit of time trying to get the API tests to go faster but didn't make much headway so giving up on that for now. Wasn't really this ticket anyway. 

## Checklist

### Frontend


api/tests/Behat/features-v2/reporting/sections/accounts/accounts.ndr.feature
```diff
@@ -1,6 +1,7 @@
-@v2 @accounts.ndr
```
jamesrwarren, line 1: there's some inconsistent issue with this so removing it for now

Gugandeep (approved):
Looks good to me.

The package we're using for parallelising seems it would be pretty useful for other tests too. Good find!

PR #654, opened 2021-05-27
DDPB-3999 fix bugs in pipeline

jamesrwarren:
## Purpose
This fixes some minor pipeline bugs.

- Add caching to build to speed up build on subsequent runs. EDIT: this did not work out well... I have added my steps into the ticket and maybe we can come back and look at it but it didn't save any time..
- Remove stabiliser as terraform has this inbuilt now and this should fix the timeouts we're seeing due to clam not stabilising in time
- Add dependencies to stop failing terraform destroy nightly
- Change the build portion of ecs helper to be a separate task as it only needs to get called for shared infra and takes 40 odd seconds which we can shave off the other calls to terraform module.

Fixes DDPB-3999

## Approach
Nothing tricky here. Just a bit time consuming back and forth with the pipeline...

## Learning
NA

## Checklist

### Frontend


acsauk (dismissed):
:shipit: 

Some nice improvements here 👍

jackgoodby (approved):
LGTM

acsauk (approved):
🚢

PR #656, opened 2021-06-01
DDPB-3483 - Assets: write Behat tests in the new suite

Gugandeep:
## Purpose
Provides tests to cover the Assets section of questions on the report.

Fixes DDPB-3483

## Approach
There is some complexity as the Assets section has a slightly different layout to other sections.
This is mainly due to the number of different asset types leading to many sections for each asset type, including property assets which have additional questions.

A brief overview of what the code is attempting to do:

Fill fields and capture data entered as an asset. (Assets are grouped by there asset type)
Sort asset groups to match the order they appear on the summary page.
Calculate totals for each asset section.
Assert expected asset sections matches found asset sections.
Assert calculated total of assets matches found calculated total of assets.

## Learning
Currently the Assets Summary Page has the following intricacies:

- Antiques, artwork & jewellery assets are grouped into one summary list
- Property section of the page does not have a subtotal (Value of properties is still taken into account for the total value of assets)
- Asset types are not displayed on the summary page in the same order they are listed on the question. Specifically 'Assets outside of England & Wales' and 'Other valuable assets section' appear earlier than how they are listed.

## Checklist

### Frontend


api/tests/Behat/features-v2/reporting/sections/assets/assets.feature
```diff
@@ -0,0 +1,51 @@
...
+    Then I should see the expected assets report section responses
+    When I follow link back to report overview page
+    Then I should see "assets" as "1 asset"
+
+@pfa-high-not-started
+  Scenario: A user adds multiple properties
+    Given a Lay Deputy has not started a report
+    When I view and start the assets report section
+    And I confirm the client has assets
+    And I add multiple property assets
```
acsauk, line 38: Hmmm I think if we're explicitly checking for a specific amount as the last line of this feature we should probably be explicit about how many we're adding rather than just using `multiple`

api/tests/Behat/features-v2/reporting/sections/assets/assets.feature
```diff
@@ -0,0 +1,51 @@
...
+    Then I should see the expected assets report section responses
+    When I follow link back to report overview page
+    Then I should see "assets" as "3 assets"
+
+@pfa-high-not-started
+  Scenario: A user adds multiple assets and a property
+    Given a Lay Deputy has not started a report
+    When I view and start the assets report section
+    And I confirm the client has assets
+    And I add multiple assets and a property
```
acsauk, line 48: Same as a above - could be clearer to explicitly say how many we're adding so the last line matches up

acsauk (changes requested):
Looks like a really tricky section - good job on some clever approaches to getting the data collected and asserted on.

Some minor changes that could make the BDD a bit clearer but ready to go after that 👍

acsauk (dismissed):
:shipit: nice nice nice

jamesrwarren (approved):

jamesrwarren:
Alex has already reviewed so just ticking it off as it has now passed all tests :). Looks good.

PR #658, opened 2021-06-07
DDPB-4003: Add cloudwatch queries

acsauk:
## Purpose
Persists a couple of common queries we use when filtering cloudwatch logs in the insights view.

Screenshot 2021-06-07 at 16 42 23

Fixes DDPB-4003


jamesrwarren (approved):
Excellent. Very handy feature!

PR #659, opened 2021-06-09
Tweak behat colours

acsauk:
## Purpose
Increases contrast for both light and dark terminals when outputting behat text.

Before:
Screenshot 2021-06-08 at 12 30 57

Screenshot 2021-06-09 at 11 30 20

After:
Screenshot 2021-06-09 at 11 10 16

Screenshot 2021-06-09 at 11 29 30


jamesrwarren (approved):
Haha, not sure how I feel about my eyeballs here but let's do it! Good find :)

PR #660, opened 2021-06-11
DDPB-3990: Behat rewrite - Deputy Costs

acsauk:
## Purpose
Covers the deputy costs section of the report in new style behat tests.

Fixes DDPB-3990.

## Approach
I added a QOL additional to the assertion code. If you just want to assert on all sections that you have set up during form filling then provide a null value for `$sectionName` to `expectedResultsDisplayedSimplified `. This will loop over every array key in `$this->submittedAnswersByFormSections` to be used as the section name to assert against.

## Checklist


jamesrwarren (approved):
This was a complicated one. Looks great :)

Gugandeep (approved):
Looks good!

PR #667, opened 2021-06-30
DDPB-3979: Update addresses for org named deputies

acsauk:
## Purpose
On CSV upload we determine if a named deputy is newly created if the following fields from the casrec CSV match the equivalent properties of named deputies in the database:

```'email1' => $dto->getDeputyEmail(),
                'deputyNo' => $dto->getDeputyNumber(),
                'firstname' => $dto->getDeputyFirstname(),
                'lastname' => $dto->getDeputyLastname(),
                'address1' => $dto->getDeputyAddress1(),
                'addressPostcode' => $dto->getDeputyPostCode(),
```

If the details match we don't make any updates to the named deputy. If the details don't match we create a new named deputy. Imagine the following scenario:

- Named Deputy is created on CSV upload
- Named deputies address1 or postcode is updated in casrec
- On the next CSV upload a new named deputy is created as their details don't match

This has led to having multiple versions of the same named deputy with slightly different details and all attached to different clients. We should stop doing this and rely on a single UID - in this case Deputy No. We should then update existing named deputies if their address details change.

Fixes DDPB-3979

## Approach
I've added a migration here to drop all existing named deputies so we can start with a fresh list to keep them unique. I've also added a unique constraint on deputy number to stop any duplicates creeping back in in the future.

There is also a migration to add `address4` and `address5` to users (and client contacts - an unintended consequence as address properties are in a trait used by both entities but I cant see any issue with this). The reason behind this is that there is no consistency in casrec on what each address field maps to (e.g. address3 could be town in some instances and county in others) so we need to take all five fields in and display them.

As we're dropping the existing named deputies we will need to release this out of hours and put the app in maintenance mode while we upload the new prof and pa csvs to repopulate the named deputy table.

This change also removes the original tests from the pipeline as changes to CSV upload would have required a rewrite of large sections of the tests which seems like wasted time seeing how close we are to finishing off behat in the new style.

## Checklist


.circleci/config.yml
```diff
@@ -235,9 +219,9 @@ workflows:
           name: integration test
           requires: [ reset integration ]
           filters: { branches: { only: [ main ] } }
-          task_name: integration_test
+          task_name: integration_test_v2
```
jamesrwarren, line 222: Yay

api/src/Repository/ClientRepository.php
```diff
@@ -21,13 +21,13 @@ public function __construct(ManagerRegistry $registry, ClientSearchFilter $filte
...
     /**
-     * Search Clients
+     * Search Clients.
      *
      * @param string $query     Search query
      * @param string $orderBy   field to order by
      * @param string $sortOrder order of field order ASC|DESC
      * @param int    $limit     number of results to return
-     * @param int $offset
+     * @param int    $offset
```
jamesrwarren, line 30: bit wierd ;)

jamesrwarren (dismissed):
Such a beast of a PR. I've looked thorugh as best I can and can't see anything wrong with it and tests seem sensible so tickety tick. Awesome stuff

api/src/Repository/ClientRepository.php
acsauk: I think PSR2 likes things lined up!

jamesrwarren (dismissed):
🍺

acsauk:
Closing in favour of #695

PR #668, opened 2021-07-02
[DDPB-3948] reporting checklist admin behat tests

coderste:
Fixes DDPB-3948

### DIFF
--- a/api/tests/Behat/features-v2/reporting/admin/checklist/checklist.feature
+++ b/api/tests/Behat/features-v2/reporting/admin/checklist/checklist.feature
@@ -5,7 +5,7 @@ Feature: Reporting Checklists - Reporting checklist
     Scenario: An admin submits the checklist form - applies to all admin roles
         Given an admin user accesses the admin app
         When I navigate to the clients search page
-        And I search for the client
+        And I search for the client 'lay'
         And I click the clients details page link
         When I navigate to the clients report checklist page
         And I submit the checklist with the form filled in
@@ -15,7 +15,7 @@ Feature: Reporting Checklists - Reporting checklist
     Scenario: An admin submits the checklist form with errors - applies to all admin roles
         Given an admin user accesses the admin app
         When I navigate to the clients search page
-        And I search for the client
+        And I search for the client 'lay'
         And I click the clients details page link
         When I navigate to the clients report checklist page
         And I submit the checklist without filling it in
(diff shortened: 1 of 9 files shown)


acsauk (commented):
So I'm open to being convinced otherwise on this but there's a variable we pass into the template called `submittedEstimateCosts` that only applies to profs with costs that are not fixed. This shows an extra section of the checklist that sums up the costs.

I guess its not really a feature as such but is a variation. The argument could be made that if we do it for this we should also test other conditional sections of the checklist but yeah, looking to see what you/anyone else reviewing this PR has to say ( cc: @Gugandeep @jamesrwarren)

Gugandeep:
> So I'm open to being convinced otherwise on this but there's a variable we pass into the template called `submittedEstimateCosts` that only applies to profs with costs that are not fixed. This shows an extra section of the checklist that sums up the costs.
> 
> I guess its not really a feature as such but is a variation. The argument could be made that if we do it for this we should also test other conditional sections of the checklist but yeah, looking to see what you/anyone else reviewing this PR has to say ( cc: @Gugandeep @jamesrwarren)

I think variations of the checklist should be tested since part of the checklists core functionality is having the right questions appearing. Although doesn't necessarily have to be done through behat tests. 
If it's easier (and possible) to do in unit tests by testing `submittedEstimateCosts`  then I'd be happy that would give us enough validity that the code is working fine.

jamesrwarren (commented):
Yep I think conditional parts should be checked using the required user to access them. Do we need to do negative tests as well? That i'm not sure about that but it would be nice to (for example that user X can see section A but user Y can't see section A). As for Gug's suggestion of doing it as part of unit tests... if it's possible and easier then maybe but I feel like behat is more the place for them to be tested as it is part of the user journey. Also happy to be challenged on that! :)

coderste:
I feel like I agree an oversight on my part about the the checklist form being different depending on the type of client it is. I feel like a single negative test may also be plausible around `client health welfare has section Y but Pfa high assets does not have section Y` just to make sure we are not seeing an incorrect checklist for certain clients.

api/tests/Behat/features-v2/reporting/admin/checklist/checklist.feature
```diff
@@ -0,0 +1,49 @@
...
+        Given an admin user accesses the admin app
+        When I navigate to the clients search page
+        And I search for the client 'lay'
+        And I click the clients details page link
+        When I navigate to the clients report checklist page
+        And I submit the checklist without filling it in
+        Then I should see all the validation errors
+
+    @admin @lay-health-welfare-submitted
+    Scenario: An lay hw checklist cannot see the pub auth specific sections - applies to all admin roles
```
acsauk, line 25: The wording on the three new scenarios could be a little clearer here (if I'm understanding it right!) 

Something like:

```
A Lay hw checklist does not contain public authority specific sections - applies to all admin roles
```

What do you think?

acsauk (changes requested):
Couple of small changes but otherwise good to go now it's covering the variations 🙌

jamesrwarren (approved):
Seems you've sorted all Alex's comments :)

acsauk (approved):
Great stuff, LGTM! :shipit:

PR #669, opened 2021-07-07
DDPB-3944: Behat Rewrite: Admin - Ingest Users (CSV Upload)

acsauk:
## Purpose
Covers the remaining lay CSV upload features for casrec and sirius CSVs

Fixes DDPB-3944

## Approach
I've had to add a step to drop the casrec table at the start of a couple of tests to help asserting on the number of casrec entities in tests...this may trip us up when we come to parrellise the test suite but we can cross that bridge when/if it happens.

This is pointing to an open PR for now so the changes were kept smaller in that branch. If this is approved before that branch is merged I'll merge this one into it instead.


jamesrwarren (approved):
Looks good. For the counts, anything we do like count before and count after could still trip us up in parallel, same with using a date column. I guess one way would be to check that each row exists rather than doing a count (eg based on some ID). 

Alternatively, we're hitting bigger numbers of minutes now anyway even in parallel so I suggest we split out into separate containers. Could have one for admin (parallel), one for client (parallel), one for all serial required tests marked @serial maybe or something.

acsauk:
No you're right - we shouldn't fall back into the trap of having tests leak out into others or state impact an assertion. I've updated the tests to extract UIDs from the CSV being uploaded to search the database by instead.

PR #671, opened 2021-07-07
DDPB-4015 - Deputy details accordion to signpost deputy users to address change

Gugandeep:
## Purpose
Add a Deputy details accordion to overview page for Org users to show the details of the named deputy attached to the client.

Fixes DDPB-4015
Fixes DDPB-4037

## Approach
Added an accordion to report overview detailing named deputy details for professional and public authority users.
Added information banner for contact email (email changes dependant on which type of user is logged in).

## Learning
Attempted using an existing macro (`informationBanner`) to display the contact email for deputy detail changes.
However, it seems you cannot pass it any html tags like `<a>` as it just displays anything you pass in as plain text.

For Behat tests, the accordions have no unique identifying attributes to distinguish between them.
The heading for each accordion is in a `govuk-accordion__section-header` tag, whilst the content of the accordion is in a `govuk-accordion__section-content`. The content is not nested under the header tag, so there wasn't an easy way to extract the necessary accordion as far as I was aware.

I did notice that depending on if the accordion was expanded or not there was an `expanded` attribute set.
I attempted to make use of the buttons on the page to close all the accordions and then press the heading for the accordions (buttons) to open the accordion. Any open accordions would have the necessary attribute to allow for easy extraction.

Unfortunately, no matter what I tried it didn't seem to work. From what I could find on the web, it seems with the Goutte driver you cannot press buttons which are not part of a form. =(

After talking with other devs, decided to extract headings and contents separately and match them manually.
The `findAll()` function returns an array with the different elements sorted in order of appearance.

## Checklist

### Frontend


acsauk (dismissed):
Great job, ready to merge! :shipit:

acsauk (approved):
:shipit:

PR #673, opened 2021-07-13
DDPB-4001 - rework

jamesrwarren:
So in the older version the env vars don't propagate to the conf file and in newer version the old health check doesn't work.


acsauk (approved):
:shipit:

PR #674, opened 2021-07-19
DDPB-3952 add analytics v2 tests

jamesrwarren:
## Purpose

Create analytics page tests for behat v2

Fixes DDPB-3952

## Approach
Bit of a different approach to this one as it was one that needed different report submissions, clients, deputies and satisfaction ratings creating on different dates so that we could check all the filters were working correctly.

Beefed out our fixtures helper to enable us to submit the correct data. 

This has to be run standalone as there is no way of running it alongside anything else as the analytics page is based off reports and clients etc which would be getting created all the time in parallel. As such I've added a new tag called v2_standalone. 

We will run anything with this tag in order before the main parallel tests. This is also the reason I didn't split these tests into two feature files. As the results of one would affect the other.

Didn't check file contents as part of this PR as that is more of a unit test functionality. Just made sure we had the right menu options and could download the relevant files.

## Learning
Working out which fields everything was being filtered on and working out how to check that counts were appearing in the right date spans was a pain. I think the overall solution is fairly good but I hope it's readable.

A lot of time was spent getting the fixture data in order so we could filter it. Hopefully this can be used elsewhere.

## Checklist

### Frontend


api/tests/Behat/bootstrap/v2/Helpers/FixtureHelper.php
```diff
@@ -744,6 +822,45 @@ public function createSuperAdmin(string $testRunId)
...
+            true,
+            $startDate,
+            $satisfactionScore
+        );
+
+        $this->createOrgUserClientNamedDeputyAndReport(
+            $testRunId.'_2',
+            User::ROLE_PA_NAMED,
+            'analytics-pa-submitted',
+            Report::TYPE_104_6,
```
jamesrwarren, line 844: just noticed this is wrong type. will let someone review first anyway to save a build ;)

coderste: This been resolved?

coderste (approved):
LGTM 👍🏻

Makefile
```diff
@@ -77,6 +77,9 @@ else
 	docker-compose -f docker-compose.yml -f docker-compose.dev.yml run --rm test sh ./tests/Behat/run-tests.sh --profile v2-tests-goutte --tags @v2
 endif
 
+behat-tests-v2-goutte: up-app-integration-tests reset-fixtures disable-debug
+	docker-compose -f docker-compose.yml -f docker-compose.dev.yml run --rm test sh ./tests/Behat/run-tests-v2.sh
+
```
acsauk, line 82: We probably want to give this a different name or drop the old existing version above to save some make confusion

api/tests/Behat/bootstrap/v2/Analytics/AnalyticsTrait.php
```diff
@@ -0,0 +1,393 @@
...
+        'layClients' => '//span[@aria-labelledby=\'metric-clients-deputyType-lay-label\']',
+        'profSatisfaction' => '//span[@aria-labelledby=\'metric-satisfaction-deputyType-prof-label\']',
+        'profSubmitted' => '//span[@aria-labelledby=\'metric-reportsSubmitted-deputyType-prof-label\']',
+        'profDeputies' => '//span[@aria-labelledby=\'metric-registeredDeputies-deputyType-prof-label\']',
+        'profClients' => '//span[@aria-labelledby=\'metric-clients-deputyType-prof-label\']',
+        'paSatisfaction' => '//span[@aria-labelledby=\'metric-satisfaction-deputyType-pa-label\']',
+        'paSubmitted' => '//span[@aria-labelledby=\'metric-reportsSubmitted-deputyType-pa-label\']',
+        'paDeputies' => '//span[@aria-labelledby=\'metric-registeredDeputies-deputyType-pa-label\']',
+        'paClients' => '//span[@aria-labelledby=\'metric-clients-deputyType-pa-label\']',
+    ];
```
acsauk, line 29: Nice - I've slowly grown to ~love~ like xpath through this whole behat shizzle

api/tests/Behat/bootstrap/v2/Analytics/AnalyticsTrait.php
```diff
@@ -0,0 +1,393 @@
...
+     * @When reports exist that were submitted at different times
+     */
+    public function reportsExistThatWereSubmittedDifferentTimes()
+    {
+        ++$this->runNumber;
+        $this->createAdditionalDataForAnalytics('14 months ago', $this->runNumber, 5);
+    }
+
+    /**
+     * @When I add more clients, deputies are reports
```
acsauk, line 41:
```suggestion
     * @When I add more clients, deputies and reports
```

api/tests/Behat/features-v2/analytics/analytics.feature
```diff
@@ -0,0 +1,63 @@
...
+        When I visit the admin analytics page
+        And I change reporting period to be all time
+        Then I should see the correct metric values displayed
+        When I change reporting period to be last 30 days
+        Then I should see the correct metric values displayed
+        When I change reporting period to be this year
+        Then I should see the correct metric values displayed
+        When I change reporting period to apply only to our generated data
+        Then I should see the correct metric values displayed
+        When I add more clients, deputies are reports
```
acsauk, line 17:
```suggestion
        When I add more clients, deputies and reports
```

api/tests/Behat/run-tests-v2.sh
```diff
@@ -10,6 +10,7 @@ confd -onetime -backend env
 
 su-exec www-data php app/console doctrine:fixtures:load --no-interaction
 
+./vendor/bin/behat --config=./tests/Behat/behat.yml --stop-on-failure --profile v2-tests-goutte --tags @v2_standalone
 ./vendor/bin/behat --config=./tests/Behat/behat.yml --stop-on-failure --profile v2-tests-goutte --tags @v2 --list-features | ./vendor/liuggio/fastest/fastest "./vendor/bin/behat --profile v2-tests-goutte --config=./tests/Behat/behat.yml {}"
```
acsauk, line 14: Hmmmm - if we end up using the shell script in the makefile is there a way we can still enable passing in the suite tag like we're going with the existing `behat-tests-v2-goutte` command? Just saves having to mess around with different make commands.

docker-compose.dev.yml
```diff
@@ -8,6 +8,7 @@ services:
             - mock-sirius-integration
         volumes:
             - ./api/tests/Behat:/var/www/tests/Behat
+            - ./api/src:/var/www/src
```
acsauk, line 11: Has this impacted run time at all? I've been finding it super annoying not having the volume and doing manual builds so very happy if performance is still good!

api/tests/Behat/bootstrap/v2/Helpers/FixtureHelper.php
jamesrwarren: This one was right. The prof one was wrong. Fixed now ;)

Makefile
jamesrwarren: Oh yes! Nice spot! I didn't use this make command in the end and had forgot to rename.. made everything called parallel now to be a bit more descriptive especially as we are going to be ditching v2 name when it's the only one at some point :)

api/tests/Behat/bootstrap/v2/Analytics/AnalyticsTrait.php
jamesrwarren: Haha love is a strong word but it's actually quite good. Was experimenting using ancestor before which might have some use case... just turns out i didn't need it :)

api/tests/Behat/run-tests-v2.sh
jamesrwarren: Er pass.. not sure how we can as this one runs in parallel. When we're wanting to run on individual tags we don't really want liuggio bundling it all up anyway. This for now keep em separate and if we work out a way then we can change it but can't think how now... actually we can do it with bash args but I'll change that in the general fix up and split out parallel builds ticket.

docker-compose.dev.yml
jamesrwarren: Didn't seem to make any difference when running individual tags. I can do a timing test tomorrow if you're uncomfortable about merging before checking that but yeah I didn't realise it was a performance thing. I thought we'd just forgotten it ;)

jamesrwarren: Ok I did have a quick check this morning and it's 251 secs with it and 229 secs without it for a full run. As it's not much difference and locally we would normally be devving and debugging that seems totally worth it for me to not have to rebuild and scratch your head when stuff hasn't changed which you believe should have! ;)

coderste (approved):

docker-compose.dev.yml
acsauk: Yep agree, lets keep it.

acsauk (approved):
:shipit:

PR #675, opened 2021-08-02
DDPB-3943 behat tests for user administration

jamesrwarren:
## Purpose
Create user administration behat tests (not file upload of users though)

Fixes DDPB-3942

## Approach
Fairly standard approach. A few caveats though:
- When we edit users, the super user edits each type but for admin and manager I decided it would take too long so just checked they have the edit option available.
- Same for checking the fields. I only go into the view on a lay rather than each as it's the same mechanism (though i do check all of them in the search window for updates).

Picked up a few mistakes in the templates as I was doing this. Fixed them. 

## Learning
NA

## Checklist

### Frontend

### DIFF
--- a/api/tests/Behat/bootstrap/v2/Common/AlertsTrait.php
+++ b/api/tests/Behat/bootstrap/v2/Common/AlertsTrait.php
@@ -8,16 +8,8 @@ trait AlertsTrait
 {
     public function assertOnAlertMessage(string $alertMessage)
     {
-        $alertDiv = $this->getSession()->getPage()->find('css', 'div.opg-alert--info');
-        if (is_null($alertDiv)) {
-            // fall back to error div
-            $alertDiv = $this->getSession()->getPage()->find('css', 'div.govuk-error-summary');
-            if (is_null($alertDiv)) {
-                $this->throwContextualException(
-                    'A div with the class opg-alert--info was not found. This suggests the page is not what was expected or a condition to display an alert has not been met'
-                );
-            }
-        }
+        $xpath = '//div[contains(@class, \'opg-alert__message\')]|//div[contains(@class, \'opg-alert--info\')]|//div[contains(@class, \'govuk-error-summary\')]';
+        $alertDiv = $this->getSession()->getPage()->find('xpath', $xpath);
 
         $alertHtml = $alertDiv->getHtml();
         $alertMessageFound = str_contains($alertHtml, $alertMessage);
--- a/api/tests/Behat/bootstrap/v2/Common/AlertsTrait.php
+++ b/api/tests/Behat/bootstrap/v2/Common/AlertsTrait.php
@@ -8,7 +8,7 @@ trait AlertsTrait
 {
     public function assertOnAlertMessage(string $alertMessage)
     {
-        $xpath = '//div[contains(@class, \'opg-alert__message\')]|//div[contains(@class, \'opg-alert--info\')]|//div[contains(@class, \'govuk-error-summary\')]';
+        $xpath = '//div[contains(@class, "opg-alert__message")]|//div[contains(@class, "opg-alert--info")]|//div[contains(@class, "govuk-error-summary")]';
         $alertDiv = $this->getSession()->getPage()->find('xpath', $xpath);
 
         $alertHtml = $alertDiv->getHtml();
--- a/api/tests/Behat/bootstrap/v2/Common/AssertTrait.php
+++ b/api/tests/Behat/bootstrap/v2/Common/AssertTrait.php
@@ -43,6 +43,29 @@ trait AssertTrait
         );
     }
 
+    public function assertBoolIsTrue(
+        $expected,
+        string $comparisonSubject
+    ) {
+        assert(
+            true == $expected,
+            $this->getAssertMessage($expected, 'false', $comparisonSubject)
+        );
+    }
+
+    public function assertStringDoesNotContainString(
+        $notExpected,
+        $found,
+        string $comparisonSubject
+    ) {
+        $foundFormatted = strval(trim(strtolower($found)));
+        $notExpectedFormatted = strval(trim(strtolower($notExpected)));
+        assert(
+            !str_contains($foundFormatted, $notExpectedFormatted),
+            $this->getAssertMessage('\''.$notExpectedFormatted.'\' should not exist in searched element!', $foundFormatted, $comparisonSubject)
+        );
+    }
+
     private function getAssertMessage(
         $expected,
         $found,
--- a/api/tests/Behat/bootstrap/v2/Common/AssertTrait.php
+++ b/api/tests/Behat/bootstrap/v2/Common/AssertTrait.php
@@ -48,7 +48,7 @@ trait AssertTrait
         string $comparisonSubject
     ) {
         assert(
-            true == $expected,
+            true === $expected,
             $this->getAssertMessage($expected, 'false', $comparisonSubject)
         );
     }
--- a/api/tests/Behat/bootstrap/v2/Common/ElementSelectionTrait.php
+++ b/api/tests/Behat/bootstrap/v2/Common/ElementSelectionTrait.php
@@ -64,6 +64,13 @@ trait ElementSelectionTrait
         $element->click();
     }
 
+    public function clickBasedOnText($text)
+    {
+        $xpath = sprintf('//a[text()[contains(., \'%s\')]]', $text);
+        $link = $this->getSession()->getPage()->find('xpath', $xpath);
+        $link->click();
+    }
+
     /**
      * @param string $sectionText Text that appears in the row/container element with the link you want to select
      * @param string $linkText    Text of the link you want to select
--- a/api/tests/Behat/bootstrap/v2/Common/ElementSelectionTrait.php
+++ b/api/tests/Behat/bootstrap/v2/Common/ElementSelectionTrait.php
@@ -102,10 +102,10 @@ trait ElementSelectionTrait
     public function iClickBasedOnAttributeTypeAndValue(string $elementType, string $attributeType, string $attributeValue)
     {
         $xpath = sprintf("//%s[@%s='%s']", $elementType, $attributeType, $attributeValue);
-        $session = $this->getSession();
-        $element = $session->getPage()->find(
+
+        $element = $this->getSession()->getPage()->find(
             'xpath',
-            $session->getSelectorsHandler()->selectorToXpath('xpath', $xpath)
+            $xpath
         );
 
         if (null === $element) {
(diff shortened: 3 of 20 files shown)


api/tests/Behat/bootstrap/v2/Common/AlertsTrait.php
```diff
@@ -8,16 +8,8 @@ trait AlertsTrait
...
-        if (is_null($alertDiv)) {
-            // fall back to error div
-            $alertDiv = $this->getSession()->getPage()->find('css', 'div.govuk-error-summary');
-            if (is_null($alertDiv)) {
-                $this->throwContextualException(
-                    'A div with the class opg-alert--info was not found. This suggests the page is not what was expected or a condition to display an alert has not been met'
-                );
-            }
-        }
+        $xpath = '//div[contains(@class, \'opg-alert__message\')]|//div[contains(@class, \'opg-alert--info\')]|//div[contains(@class, \'govuk-error-summary\')]';
```
acsauk, line 11: 😍

acsauk: Although one suggestion - you can lose the `\` if you use `"` instead

api/tests/Behat/bootstrap/v2/Common/AssertTrait.php
```diff
@@ -43,6 +43,29 @@ public function assertIntEqualsInt(
         );
     }
 
+    public function assertBoolIsTrue(
+        $expected,
+        string $comparisonSubject
+    ) {
+        assert(
+            true == $expected,
```
acsauk, line 51: Not sure what level of weirdness PHP goes to when using `==` vs. `===` but I always tend to err on the side of caution and default to `===` to be sure the strictest assertion is used.

api/tests/Behat/bootstrap/v2/Common/AssertTrait.php
```diff
@@ -43,6 +43,29 @@ public function assertIntEqualsInt(
...
+
+    public function assertStringDoesNotContainString(
+        $notExpected,
+        $found,
+        string $comparisonSubject
+    ) {
+        $foundFormatted = strval(trim(strtolower($found)));
+        $notExpectedFormatted = strval(trim(strtolower($notExpected)));
+        assert(
+            !str_contains($foundFormatted, $notExpectedFormatted),
```
acsauk, line 64: I think this is a PHP 8 only function, is it working? I do remember seeing some Symfony polyfill code for this a while ago so maybe that's helping us out here? https://www.php.net/manual/en/function.str-contains.php

api/tests/Behat/bootstrap/v2/UserManagement/UserManagementTrait.php
```diff
@@ -0,0 +1,677 @@
...
+    }
+
+    /**
+     * @When I should see the correct search results
+     */
+    public function iShouldSeeTheCorrectSearchResult()
+    {
+        $xpath = '//h2[@id="users-list-title"]/parent::div/p[@class="govuk-body"]';
+        $userText = $this->getSession()->getPage()->find('xpath', $xpath)->getHtml();
+        $pluralUsers = 1 == $this->userCount ? 'user' : 'users';
```
acsauk, line 55: SImilar to above, could we update the equality check to be strictest `===`?

api/tests/Behat/bootstrap/v2/UserManagement/UserManagementTrait.php
```diff
@@ -0,0 +1,677 @@
...
+    /**
+     * @When I search for the newly created user
+     */
+    public function iSearchForNewlyCreatedUser()
+    {
+        $this->searchUserWithFilter('', $this->expectedUsers[0]['email']);
+    }
+
+    /**
+     * @When I can see the user as non active in search results
```
acsauk, line 225: Should this be using `Then` if it's an assertion?

api/tests/Behat/bootstrap/v2/UserManagement/UserManagementTrait.php
```diff
@@ -0,0 +1,677 @@
...
+
+    /**
+     * @When I navigate to the add a new user page
+     */
+    public function iNavigateToAddNewUser()
+    {
+        $this->iAmOnAdminUsersSearchPage();
+        $this->clickBasedOnText('Add new user');
+    }
+
```
acsauk, line 297: This should probably live in `INavigateToAdminTrait` with its other navigations sibling functions.

api/tests/Behat/features-v2/user-management/create-user.feature
```diff
@@ -0,0 +1,37 @@
...
+    Then I see that an activation email has been sent to the user
+    When I search for the newly created user
+    Then I can see the user as non active in search results
+    When I resend activation email
+    Then I see that activation link has been sent
+
+@super-admin
+Scenario: A super admin creates admin users
+    Given a super admin user accesses the admin app
+    When I add each of the appropriate user types for a super admin
```
acsauk, line 17:
```suggestion
    When I add each of the available user types for a super admin
```

Feels like this is a bit more explicit about the fact adding users with specific roles are limited to different admin type? What do you think?

api/tests/Behat/features-v2/user-management/edit-user.feature
```diff
@@ -0,0 +1,43 @@
...
+
+    @admin-manager
+    Scenario: A admin manager user edits other admins
+        Given an admin manager user accesses the admin app
+        And I have created the appropriate test users to edit
+        When I view the super admin user
+        Then I should not be able to edit that user
+        When I view the admin manager user
+        Then I should not be able to edit that user
+        But I should be able to delete that user
```
acsauk, line 30: Oooo I've never used `But` before - I like!

api/tests/Behat/features-v2/user-management/search-user.feature
```diff
@@ -0,0 +1,29 @@
+@v2 @admin-user-search @admin-user
+Feature: Admin - User Search
```
acsauk, line 2: Could we get a clarification in the feature title that this applies to all admin roles?

client/templates/Admin/Index/index.html.twig
```diff
@@ -67,7 +67,7 @@
                     </p>
                 {% else %}
                     <p class="govuk-body">
-                        Found {{ users | length }} users
+                        {% if (users | length) == 1 %}Found 1 user{% else %}Found {{ users | length }} users{% endif %}
```
acsauk, line 70: 👍

client/translations/admin-clients.en.yml
```diff
@@ -4,7 +4,7 @@ clientSearch:
   searchBy: Search by client surname or case number
   clientTable:
     heading: Client list
-    results: Found %numberOfClients% clients jim
```
acsauk, line 7: Hahahaha how long has that been in there? 😆

acsauk (changes requested):
Few minor points but looking great 👍

api/tests/Behat/bootstrap/v2/Common/AlertsTrait.php
jamesrwarren: Yeah I did have it like that and for some reason I preferred the slashes but acually slashes where we don't need them and probably bad.. I dunno. I can change that :)

api/tests/Behat/bootstrap/v2/Common/AssertTrait.php
jamesrwarren: Ah thanks!

api/tests/Behat/bootstrap/v2/Common/AssertTrait.php
jamesrwarren: Doesn't complain and I've used it a few times before. Didn't realise it's php8!

api/tests/Behat/bootstrap/v2/UserManagement/UserManagementTrait.php
jamesrwarren: Yes! Good spot

api/tests/Behat/features-v2/user-management/create-user.feature
jamesrwarren: Sure. I'm down with that!

api/tests/Behat/features-v2/user-management/search-user.feature
jamesrwarren: Yes good point

client/translations/admin-clients.en.yml
jamesrwarren: The shame!!!! It was probably a different Jim... not me, nothing to see here... Since I did that massive `raw` removal PR. So er... quite a while ;)

api/tests/Behat/bootstrap/v2/Common/AssertTrait.php
jamesrwarren: If you want me to use a replacement I can but have left in as it's doing it's thang (and we're going to upgrade to php8 soonish anyway)

acsauk: Naw its all good if its working. Like you said we'll be there soon so makes sense to use the shiny things when we can.

api/tests/Behat/bootstrap/v2/Common/AlertsTrait.php
acsauk: Usually I'm not fussed but when it's stuff like regex or xpath and a slash could actually mean something I tend to try to avoid where possible.

api/tests/Behat/bootstrap/v2/UserManagement/UserManagementTrait.php
jamesrwarren: sneaky comment hiding away :)

api/tests/Behat/bootstrap/v2/Common/ElementSelectionTrait.php
```diff
@@ -64,6 +64,13 @@ public function iClickOnNthElementBasedOnRegex(string $regex, int $elementIndex)
         $element->click();
     }
 
+    public function clickBasedOnText($text)
+    {
+        $xpath = sprintf('//a[text()[contains(., \'%s\')]]', $text);
+        $link = $this->getSession()->getPage()->find('xpath', $xpath);
+        $link->click();
+    }
+
```
acsauk, line 73: Ah missed what this was doing in the original code. There's the built-in mink function that does this in one line:

```
$this->clickLink('Click me!');
```

acsauk (dismissed):
:shipit:

acsauk (approved):
:shipit:

PR #677, opened 2021-08-03
tweak to sirius request timeout

jamesrwarren:
## Purpose

image

believe timeouts are causing healthcheck to fail even though we're seeing 200s despite sirius being down (as we had expected and fixed in a previous ticket months ago)

image

Pretty sure that going for timeouts under 2 secs is the way to go here. Doesn't really matter if occasionally says it's offline when it isn't as this doesnt affect our availability but it will stop us getting alerts when sirius is down and we're not (I hope).


acsauk (approved):
Good detective work - never seemed to make sense with the alters :shipit:

PR #678, opened 2021-08-04
Added date feature flag for additional benefits questions

Gugandeep:
## Purpose
Adding a date-based feature flag to allow developers to show/hide certain features and functionality behind the flag.
Using a date as to accommodate certain scenarios such as hiding questions on reports which have submitted before said date.

Fixes DDPB-4038

## Approach
Update terraform to create additional feature flag in AWS Parameter Store.
Update AWS IAM Policy to allow reading of variable to role.
Update scripts to create flag for localstack.
Add feature flag to Parameter Store Service.
Update Report and NDR controllers to pull in feature flag.

## Learning
Not only do you need to create a feature flag in `parameters.tf`, but you will also need to ensure you update the policy for the aws role to allow for reading in `front.tf`.

## Checklist

### Frontend

### DIFF
--- a/api/scripts/localstack-init.php
+++ b/api/scripts/localstack-init.php
@@ -72,7 +72,7 @@ $ssmClient->putParameter([
 $ssmClient->putParameter([
      'Name' => '/default/parameter/benefits-questions',
      'Type' => 'String',
-     'Value' => '03-08-2021 14:00:00',
+     'Value' => '31-12-2030 00:00:00',
      'Overwrite' => true,
 ]);
 
--- a/api/scripts/localstack-init.php
+++ b/api/scripts/localstack-init.php
@@ -5,14 +5,15 @@ use Aws\S3\S3Client;
 use Aws\Ssm\SsmClient;
 use GuzzleHttp\Client;
 
-require __DIR__ . '/../vendor/autoload.php';
+require __DIR__.'/../vendor/autoload.php';
 
 function isLocalstackAvailable()
 {
     try {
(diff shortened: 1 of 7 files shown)


api/scripts/localstack-init.php
```diff
@@ -37,73 +38,80 @@ function isLocalstackAvailable()
...
     'Value' => '100',
-    'Overwrite' => true
+    'Overwrite' => true,
+]);
+
+$ssmClient->putParameter([
+     'Name' => '/default/parameter/benefits-questions',
+     'Type' => 'String',
+     'Value' => '31-12-2030 00:00:00',
+     'Overwrite' => true,
```
acsauk, line 76: Do we need this in API if the feature flag is only on the client side?

acsauk (commented):
Great work - one Q on the API localstack init script but otherwise good to go 👍

api/scripts/localstack-init.php
Gugandeep: Good question.

I copied the pattern for the other flags and added it to both scripts. 
However since we're not touching the API side here, it's not really needed in this PR.

I'm not sure if the feature flag will be required on the API side of the app, so I can remove it for now and we can add it back in when it's needed.

acsauk: Yeah, I think if we're not using it yet we should drop it out. If we never need it on the API side, it may be confusing for future selves/new devs.

acsauk (approved):
:shipit:

PR #679, opened 2021-08-09
DDPB-3950: Behat rewrite - Report submissions

acsauk:
## Purpose
Covers the report submissions section of the app (including document synchronisation)

Fixes DDPB-3950

## Approach
I've done a few more tidying up steps in this PR as we're coming to the end of behat work now:

* Moved to 1 and 0 rather than true and false for activating xdebug in tests
* Added a document and valid report submission to submitReport() in test helpers
* Added report types to behat fixture tags that were missing
* Set user logging in to the front end of the app to interactingWithUserDetails so we can use it in other steps


api/tests/Behat/bootstrap/v2/Common/INavigateToAdminTrait.php
```diff
@@ -73,4 +73,12 @@ public function iNavigateToTheReportChecklistPage()
         $this->clickLink('Checklist');
         $this->iAmOnAdminReportChecklistPage();
     }
+
+    /**
+     * @When I navigate to the admin report submissions page
+     */
+    public function iNavigateToAdminReportSubmissionsPage()
+    {
+        $this->clickLink('Submissions');
```
coderste, line 82: It seems all the other functions have the `$this->iAmOnPage();` function to evaluate that you are indeed on the correct page after link click. Maybe want to do the same here?

coderste (changes requested):

api/tests/Behat/bootstrap/v2/Common/INavigateToAdminTrait.php
acsauk: Good catch - will add in 👍

coderste (approved):

PR #681, opened 2021-08-13
Ddpb 4044

jamesrwarren:
## Purpose

To fix security vulnerabilities in container we need to upgrade to something that works well with alpine (because other containers always seem to unearth sec vulnerabilities)

Co-authored-by @Gugandeep 

Fixes DDPB-4044

## Approach

Used weasyprint. Started with the command line standalone but decided to bring in as a package for more granular logging. Renamed everything to generic htmltopdf as the **function** of the service rather than the product. This means if we want to swap out for opg pdf generator then we don't need to rename all the stuff again.

Other considerations:

There is an opg html to pdf converter but because it wasn't on an alpine image and looked like it was going to need a bit of tweaking which I felt a bit out of my depth to do (not familiar with language or any of the packages). Felt this would have same issues as current image and really not being alpine was a deal breaker that I didn't feel it was worth the effort to put in. Happy to move to this when it becomes alpine.

There was an alpine binary for wkhtmltopdf so did try and stick with this. Unfortunately it kept complaining about memory leaks which I couldn't work out and seemed dog slow on alpine. Even on the documentation on the site, it suggests using weasyprint instead as the future of that project seems uncertain and one of the packages it relies on, is not being updated: https://wkhtmltopdf.org/status.html
For all those reasons I gave up on wkhtmltopdf and opted for weasyprint.  

weasyprint has the advantage (and disadvantage) of being in python. I say advantage because we know python and can debug it and we can use the wrapper for wkhtmltopdf with it. I say disadvantage as it's slower than wkhtmltopdf. For the size of PDFs we use this seems to be fine. Would be a different story if it was 50 page pdfs. This is basically because python is a bit slow. The advantages though of using weasyprint are that it's super configurable, very well supported with a small team that respond to issues and regular releases and they seem to be trying to make it less reliant on outside packages for the text formatting and layout etc (pango and cairo can go bye bye). So I feel this will work very well as a stop gap and possibly long term.

## Learning

Had to learn how CSS worked. To get to the point of having those two arguments to the package... there was extensive pain. 

Still learned a lot about everything from media tags to different font formats so was fairly interesting and means we feel comfortable(ish) debugging this kind of thing in the future.

## Checklist

### Frontend


client/src/Service/Availability/HtmlToPdfAvailability.php
```diff
@@ -0,0 +1,36 @@
+<?php
+
```
acsauk, line 2:
```suggestion
declare(strict_types=1);
```

htmltopdf/app.py
```diff
@@ -30,34 +40,21 @@ def application(request):
...
-        # Add source file name and output file name
         file_name = source_file.name
-        args += [file_name, file_name + ".pdf"]
 
-        # Execute the command using executor
-        execute(' '.join(args))
+        try:
+            HTML(file_name, media_type='screen', encoding="utf-8").write_pdf(file_name + ".pdf",
+               stylesheets=[CSS(
+               string='@page {size: Letter;margin: 0.2in 0.44in 0.2in 0.44in;}')])
```
acsauk, line 53: Is there a reason this needs to be inline here rather than in a CSS file?

acsauk (commented):
One tiny change and a question but otherwise good to go 🚢

acsauk (approved):
:shipit:

acsauk (approved):
:shipit:

PR #686, opened 2021-09-01
DDPB-4047 Container level parallelisation of tasks

jamesrwarren:
## Purpose

As we get more tests, there is a the likelihood that times will creep up again. Ideally, we would like to be able to through more resources at the tests and split them up further as they should now all be able to be run independently.

This will keep times to a reasonable level.

Fixes DDPB-4047

## Approach

Whilst it would have been easy to recreate different tasks for each part of the parallel running, it would have been a bit messy and hard to update in the future.

As such, had to delve into a bit of golang again and add a flag that controls the override parameter. We can then override the entrypoint with a command. Also removed the entry point to make this more flexible for this task. This whole bit took a long time to get working. Go structures can be sooo tricky.

Then it was simply a case of creating an entry for each container and balancing the tasks over them. 

Next came the issue that it started throttling connections.. as such I had to spin up extra resources and then spin them down after. There is literally no way of doing a `when: always` in circleci at the job level which is super annoying as this would be a really nice feature. As such, in case of a failure we rely on the scaling policy to bring the containers back down to 1. 

I think the spin up time of the standalone really eats into the gains we make with this approach. If we could rewrite the standalone test (which i wrote so probably should have thought of better way to do it) then we would be properly parallel and this approach would be better. It's still a tiny bit faster and should be faster the more tests we add... 

We should monitor the situation.

## Learning

Golang erg... In future i should try working backwards. I'm not sure i would have invested the time if I had seen the issues with the throttling.. or maybe i would. It's one of those where it's quite hard to work out how it's going to work out without doing the work as it needs real infra and real changes to deploys..

## Checklist

### Frontend


api/tests/Behat/run-tests-parallel.sh
```diff
@@ -7,14 +7,17 @@ export APP_ENV=dev
...
 
 runtime=$((end-start))
 
 echo "Time take: ${runtime} secs"
+
+if [ runtime -gt 300 ]
+then
+    echo "Stage taking too long. Failing the build!"
+    echo "Please split out your tests to a new container"
+    exit 1
```
acsauk, line 22: 🙌

docs/TESTING.md
```diff
@@ -150,6 +150,17 @@ Some parts of our forms have values that are directly related to each other. In
 
 To remove an item use the `removeAnswerFromSection()` function providing a single field name contained in answer group to be removed, and they will both be removed.
 
+#### Parallelisation
+There are two levels of parallelisation involved with the behat tests. Firstly we have the processor level and then the container level.
+Each container is a high specced container that has 4 processors (this is the maximum). We use `liuggio/fastest/` to parallelise our tests across processors.
+
+The next level is at the container level. As we can get further speed improvements, we split out similar themed tests across containers with tags.
+
+There is a a built in timer that ensures a tagged group of tests do not exceed 240 seconds total. If they do then you should split them into a new container.
```
acsauk, line 159: Should this be 300 seconds?

acsauk (approved):
:shipit:

PR #690, opened 2021-09-08
DDPB-4062: Create the entities and migrations for the new report section

acsauk:
## Purpose
Creates the entities and migration for the new form section 'Benefits check'.

Fixes DDPB-4062

## Approach
Following on from comments made in various retros and health checks I've ensured the ids of the two entities are UUIDs to protect against users trying to access resources incrementally. 

There may be some tweaks to the entities when we come to implement the forms - especially `ClientBenefitsCheck::whenLastCheckedEntitlement` as this can be either a string or DateTime so some type of conversion will need to happen after checking if the value is a DateTime (e.g. `DateTime::createFromFormat('m/Y', '01/2021')`).

The readme I added is a first stab and will evolve as we implement the rest of this form section.

## Checklist

### DIFF
--- a/api/src/Entity/Report/ClientBenefitsCheck.php
+++ b/api/src/Entity/Report/ClientBenefitsCheck.php
@@ -1,0 +1,141 @@
+<?php
+
+ declare(strict_types=1);
+
+namespace App\Entity\Report;
+
+use Doctrine\Common\Collections\ArrayCollection;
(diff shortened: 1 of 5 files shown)


Gugandeep (approved):
Looks good to me.

Great idea to add a readme file to make adding future sections are a lot easier!

PR #691, opened 2021-09-09
DDPB-4092c specify environment for main build

jamesrwarren:
## Purpose
Mistakes were made.. we need to specify environments for all the tasks running against main. It uses the WORKSPACE env var which gets set as tf_workspace to decide what environment to use. It does this with sh script that checks if it's been set so we can have dynamic envs. In main it needs to be set.


acsauk (dismissed):
:shipit:

acsauk (approved):
:shipit:

PR #692, opened 2021-09-10
Adds javascript disabled tracking for logged in users

acsauk:
## Purpose
Following on from recent conversations we needed a way to track how many of our users have JS disabled in their browsers so we can make more informed decisions on how we design the app. This adds tracking for for lays and org users when they log in so we can breakdown how many users there are in Google Analytics.


Gugandeep (approved):
Looks good!

PR #701, opened 2021-10-05
[DDPB-4058] Make our cookie opt in/opt out options fully compliant with GDS standards

coderste:
Make our cookie opt in/opt out options fully compliant with GDS standards


acsauk (commented):
All looks good to me - could we get a few tests around the cookieBanner.js code too? 🙏 We use jest for testing JS code. I don't think we're running them in the pipeline yet but we should definitely look at implementing that soon.

acsauk (approved):
As long as we get round to writing the tests I'm fine with that approach. Good job and thanks for pushing modernisation/tech debt tickets 💪

coderste:
> As long as we get round to writing the tests I'm fine with that approach. Good job and thanks for pushing modernisation/tech debt tickets 💪 

Awesome! Yeah I'm all for tests across the board both backend/frontend and hopefully we can get closer to that goal the more we modernise our stack as it should allow us to write tests in an easier more readable format :)

PR #708, opened 2021-10-20
DDPB-4126: NDR client benefits check questions

acsauk:
## Purpose
Adds the new client benefits check section questions to NDR reports. All NDR reports will receive this question.

Fixes DDPB-4126.

## Approach
I really strongly dislike the fact that `NDR`s are their own entities rather than just a type of `Report`. Its lead to so much duplication here but it seemed like the lesser of two evils as if I had started to try to handle the entities differently then it would have confused our future selves. I saved a little bit of repetition in some controller code where I felt it wasn't too confusing but am open to any other suggestions or re-work requests it makes sense to separate them out to their own controllers too.

## Checklist


coderste (dismissed):
LGTM 👍🏻

coderste (approved):

PR #711, opened 2021-10-29
DDPB-4135: Update Report entity type constants to meaningful names

acsauk:
## Purpose
We refer to the type of report in the codebase using its numerical format. While it's possible to refer to confluence pages and code comments it's not always immediately clear if the report type you are using is correct for the deputy type and client circumstances. This update gives each report type const in `Report` a deputy and form type name rather than just reflecting the numerical representation of this. E.g. :

```php
// Not helpful
const TYPE_104 = '104';

// More helpful
const LAY_HW_REPORT_TYPE = '104';
```

I've also grouped various common report types together so we can refer to them in code. For example, you can now check if a report type belongs to a certain user group or has certain sections by checking if it appears in one of the grouping arrays:

```php
in_array('104', Report::allRolesHwAndCombinedReportTypes()) // true
```

Fixes DDPB-4135

## Checklist

### DIFF
--- a/api/src/Entity/Report/Report.php
+++ b/api/src/Entity/Report/Report.php
@@ -88,12 +88,6 @@ class Report implements ReportInterface
 
     const ENABLE_FEE_SECTIONS = false;
 
-    private static $reportTypes = [
-        self::LAY_PFA_LOW_ASSETS_TYPE, self::LAY_PFA_HIGH_ASSETS_TYPE, self::LAY_HW_TYPE, self::LAY_COMBINED_LOW_ASSETS_TYPE, self::LAY_COMBINED_HIGH_ASSETS_TYPE,
-        self::PA_PFA_LOW_ASSETS_TYPE, self::PA_PFA_HIGH_ASSETS_TYPE, self::PA_HW_TYPE, self::PA_COMBINED_LOW_ASSETS_TYPE, self::PA_COMBINED_HIGH_ASSETS_TYPE,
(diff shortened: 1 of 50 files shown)


api/src/Entity/Report/Report.php
```diff
@@ -1431,4 +1400,81 @@ public function setBenefitsSectionReleaseDate(?DateTime $benefitsSectionReleaseD
...
+    }
+
+    public static function profPfaAndCombinedReportTypes(): array
+    {
+        return [
+            self::PROF_PFA_LOW_ASSETS_TYPE, self::PROF_PFA_HIGH_ASSETS_TYPE, self::PROF_COMBINED_LOW_ASSETS, self::PROF_COMBINED_HIGH_ASSETS,
+        ];
+    }
+
+    public static function allRolesLowAssetsReportTypes(): array
```
Gugandeep, line 1477: Is this function needed/used for anything?

Gugandeep (commented):
Just a minor comment, looks like an unused/unfinished function.

But besides that, looks good.
Really nice now that you can see a report type and know what's going to be in it!

coderste (approved):
LGTM 👍🏻

api/src/Entity/Report/Report.php
acsauk: Good spot - think I started this and decided we didn't need it. I'll remove.

coderste (dismissed):

coderste (approved):

PR #715, opened 2021-11-04
DDPB-4147: Upgrade to xdebug v3

acsauk:
## Purpose
Upgrades to the latest version of xdebug so we can turn it on without the app acting like its stuck in tar.

I've based this on everyone using PHPStorm to debug but if people are using other editors (VSCode etc.) feel free to make a follow-up PR to set the IDE key value based on an env var (or some other extensible method). If port 9003 turns out to be a problem for anyone too we can also whack that in an env var. Happy to say though the app acts almost like normal when debugging now 💪 .

Fixes DDPB-4147.


coderste (approved):
All looks good to me 👍🏻

PR #718, opened 2021-11-10
DDPB-4143: Add client benefits check sections to report and checklist PDF

acsauk:
## Purpose
Adds the client benefits check responses to the report PDF and the casework confirmation that they are satisfied with the response to the checklist PDF.

Fixes DDPB-4143

## Approach
There wasn't a lot to do here in the end as the CSS classes were all set up for the other sections. There is a bug when a user adds 22+ entries to any section that pushes the items down onto a new page leaving just the question they relate to and a shaded box. We've decided to handle this in a follow up ticket seeing as its an existing bug affecting other report sections on the PDF.

## Learning
I've added a quick update to the behat for money out as it seemed to be causing issues in CI. It didn't happen after I added it but as its not a consistent failure that's not proof this has had an impact. We can revert to the original setup if the issue keeps on happening.

## Checklist

### Frontend


client/templates/Report/Formatted/_client_benefits_check.html.twig
```diff
@@ -0,0 +1,64 @@
+{% set translationDomain = "report-client-benefits-check" %}
+{% trans_default_domain translationDomain %}
+
+{% set transOptions = {
+    '%client%': report.client.firstname | e,
+} %}
+
+<div class="section break-before" id="client-benefits-check-section">
+    <h2 class="section-heading">{{ 'common.pageTitle' | trans(transOptions) }}</h2>
+
```
coderste, line 10: Not that it matters much but are we able to sort out the whitespacing here?

coderste (approved):

client/templates/Admin/Client/Report/Formatted/checklist_formatted_body.html.twig
```diff
@@ -164,6 +164,27 @@
             </div>
         {% endif %}
 
+        {% if report.hasSection('clientBenefitsCheck') %}
+            <div class="section">
+                <h2 class="section-heading">{{ (page ~ '.heading.lodging.clientBenefitsCheck') | trans({}, 'admin-checklist') }}</h2>
+
+                <h3 class="label question bold">{{ (page ~ '.form.clientBenefitsChecked.label') | trans({}, 'admin-checklist') }}</h3>
```
coderste, line 171: Formatting here as well

client/templates/Report/Formatted/_client_benefits_check.html.twig
acsauk: I've dropped the line break between `box` and `div` but it looks like the markup style for other templates is to leave a break between `h` tags and content. If you're having major eye twitches we could look at changing that up for all templates?

coderste: Yeah I noticed the `h` tags line break from another file I was looking at I did originally had a comment about that but then removed it :)

coderste (approved):

PR #722, opened 2021-11-15
[DDPB-3884] Remove password hashes from api responses

coderste:
## Purpose
Remove password hashes from being returned in an API response that holds user data. There was 1 issue with this where when a user changes their password it would no longer log them out afterwards. To solve this we just redirect the user to the logout page after submitting a password change and the form is valid

Fixes DDPB-3884

### DIFF
--- a/client/src/Controller/SettingsController.php
+++ b/client/src/Controller/SettingsController.php
@@ -121,9 +121,7 @@ class SettingsController extends AbstractController
             ]));
             $request->getSession()->set('login-context', 'password-update');
 
-            $successRoute = $this->getUser()->isDeputyOrg() ? 'org_settings' : 'account_settings';
-
-            return $this->redirect($this->generateUrl($successRoute));
+            return $this->redirect('/logout');
         }
 
         return [
--- a/Makefile
+++ b/Makefile
@@ -96,7 +96,7 @@ else
 	docker-compose -f docker-compose.yml -f docker-compose.dev.yml run --rm test sh ./tests/Behat/run-tests.sh --profile v2-tests-goutte --tags @v2
 endif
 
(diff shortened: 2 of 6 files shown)


acsauk (commented):
Code changes look good but could we get a new/updated regression test in `AuthControllerTest` proving we no longer get the hashed password back?

client/src/Controller/SettingsController.php
```diff
@@ -110,19 +110,20 @@ public function passwordEditAction(Request $request)
...
-            $this->restClient->put('user/' . $user->getId() . '/set-password', json_encode([
+            $this->restClient->put('user/'.$user->getId().'/set-password', json_encode([
                 'password_plain' => $plainPassword,
             ]));
             $request->getSession()->set('login-context', 'password-update');
 
             $successRoute = $this->getUser()->isDeputyOrg() ? 'org_settings' : 'account_settings';
-            return $this->redirect($this->generateUrl($successRoute));
+
+            return $this->redirect('/logout');
```
mattmachell, line 126: Presumably this changes the user flow for the redirect back? So may want to check that with product side.

Can you avoid needing to do this by pushing the isDeputyOrg and generate url up before the password reset or is there some state required? 

Also, presumably the successRoute is no longer used so doesn't need to be set at all.

coderste: As far as I know this doesn't actually change the user flow. Whenever a password was edited it would always log them out as that's just a security setting of Symfony. The reason we had to back the change sot he redirect was to the `/logout` page was because removing the `password hash` from the JSON response made it so once a password was edited it would no longer log them out.

So this redirect change just keeps that user flow the same :)

acsauk (approved):
:shipit:

PR #723, opened 2021-11-15
DDPB-3851: Allow discharge of org deputies without associated named deputies

acsauk:
## Purpose
In the templates for the org client details page we hide the button for discharging a deputy behind a check on if a named deputy is associated with the client. While this makes logical sense we have over 7000 deputies without a named deputy associated with them since we dropped the named deputy table and rebuilt via the CSV. The reason behind this high number is due to the fact we don't remove a client/deputy from digideps when they drop off the CSV. This means in the rare cases where auto discharge doesn't happen on CSV upload we are unable to discharge a deputy and allow a new deputy to be associated with a client.

This change moves the discharge deputy button outside of the check on named deputy so we can discharge when needed. This may be a little confusing for users so we should link in with the teams responsible for this for feedback. There may be no end users actively using the button as currently only super admins and admin managers have access to the button.

Fixes DDPB-3851.

## Checklist

### Frontend

### DIFF
--- a/api/tests/Behat/features-v2/client-management/admin/discharge.org.feature
+++ b/api/tests/Behat/features-v2/client-management/admin/discharge.org.feature
@@ -2,22 +2,31 @@
 Feature: Admin - Client Discharge
 
 @super-admin @prof-admin-health-welfare-submitted
-  Scenario: A super admin user discharges a client
+  Scenario: A super admin user discharges an org client
     Given a super admin user accesses the admin app
     When I visit the admin client details page for an existing client linked to a deputy in an Organisation
     And I attempt to discharge the client
     Then the client should be discharged
 
 @admin-manager @prof-admin-health-welfare-submitted
-  Scenario: An admin manager user can discharge a client
+  Scenario: An admin manager user discharges an org client
     Given an admin manager user accesses the admin app
     When I visit the admin client details page for an existing client linked to a deputy in an Organisation
     And I attempt to discharge the client
     Then the client should be discharged
 
 @admin @prof-admin-health-welfare-submitted
-  Scenario: An admin user cannot discharge a client
+  Scenario: An admin user discharges an org client
     Given an admin user accesses the admin app
     When I visit the admin client details page for an existing client linked to a deputy in an Organisation
     And I attempt to discharge the client
     Then the client should not be discharged
+
+@admin-manager @prof-admin-health-welfare-submitted @acs
+Scenario: An admin manager user discharges an org client without a named deputy
+    Given a Professional Deputy has submitted a Health and Welfare report
+    And an admin manager user accesses the admin app
+    And the client does not have a named deputy associated with them
+    When I visit the admin client details page for an existing client linked to a deputy in an Organisation
+    And I attempt to discharge the client
+    Then the client should be discharged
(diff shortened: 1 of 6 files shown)


api/tests/Behat/features-v2/client-management/admin/discharge.org.feature
```diff
@@ -0,0 +1,32 @@
...
+
+@admin-manager @prof-admin-health-welfare-submitted
+  Scenario: An admin manager user discharges an org client
+    Given an admin manager user accesses the admin app
+    When I visit the admin client details page for an existing client linked to a deputy in an Organisation
+    And I attempt to discharge the client
+    Then the client should be discharged
+
+@admin @prof-admin-health-welfare-submitted
+  Scenario: An admin user discharges an org client
```
Gugandeep, line 19: Really minor, but is this meant to be?

`An admin user cannot discharge an org client`

Gugandeep (commented):
Looks good to me.

Left a really minor comment on the name of a scenario.

api/tests/Behat/features-v2/client-management/admin/discharge.org.feature
acsauk: I've been having this debate in my head around naming scenarios based on the feature itself of the result of using the feature. I'm not sure its been totally consistent between scenarios in the codecase but worth having a discussion at some point to see what we find most useful/clear. What's your preference @Gugandeep?

@coderste - any thoughts?

Gugandeep: Yeah it's a tough one to decide on.
I think it's nice to have the result of the feature in the scenario just so that when you read it you know kinda know what to expect to happen. (Although this is possible from reading the steps too I suppose).

Having just the action in the scenario is a bit more concise, and I guess more future proof if say the result of the test is likely to change. For example in this case if we made a change to allowed admin users to discharge clients, the description `An admin user discharges an org client` fits just as well.

In terms of preference, I might go with having the result in the test, just as long it's not comprising the action part of the description.

mattmachell: Having the scenario be what _should_ happen at during the interaction will help your future self when the context is lost and you get a related ticket. If you leave it ambiguous that becomes harder as you need to dig to work out the real desired behaviour, not just the steps to reach it.

coderste: Yeah I'm sort of torn between the 2 if I'm honest, I can understand both perspectives. I think the point Gugs raised around future changes is making me lean towards the current naming you have as like he said if it changes in the future so admin can discharge an org client then the current scenario name is still valid whereas you add the word cannot and then you would have to update that.

I think mainly for that reason I'm going to vote for the `An admin user discharges an org client` which is the current scenario name

acsauk: Hmmmm I'm now actually reconsidering my position after Matts comment! I'm thinking of occasions where future selves are trying to work out why an integration test has failed. If we haven't got it 100% clear in our steps then the intent may not be as clear as a definite scenario that's describing the feature outcome rather than just the feature itself.

I do get what you're both saying about having to update when changing features in the future @coderste and @Gugandeep but in those instances, we'd be updating the steps anyway so editing the scenario name isn't a huge extra job to do.

Gugandeep (approved):

PR #724, opened 2021-11-15
DDPB-4167: Add audit logging for failed registration

acsauk:
## Purpose
Moving the logs we emit for failed self-registration to audit logs. This gives us a single place to search for these when we need to debug a users registration attempts and opens up some flexibility in the future to add a front end if we feel we need to.

Fixes DDPB-4176

## Learning
I got sick of waiting for actual AWS audit logs to appear in AWS on the branch env when developing this ticket so I added a make command to get them locally.

## Checklist


api/tests/Unit/Controller/SelfRegisterControllerTest.php
```diff
@@ -180,4 +190,66 @@ public function throwErrorForDuplicate()
...
+
+        $token = $this->login('deputy@example.org', 'DigidepsPass1234', API_TOKEN_DEPUTY);
+
+        $responseArray = $this->assertJsonRequest('POST', '/selfregister', [
+            'mustFail' => true,
+            'AuthToken' => $token,
+            'data' => [
+                'firstname' => 'Zac',
+                'lastname' => 'Tolley',
+                'email' => 'wrong@email.com',
```
mattmachell, line 221: Worth using example.org/com here as it's reserved, rather than "fake" emails like this.

coderste (dismissed):

coderste (approved):

PR #725, opened 2021-11-17
DDPB-4094: Upgrade client to PHP 8

acsauk:
## Purpose
Upgrades the frontend and admin apps to PHP 8.0.12.

Fixes DDPB-4094

## Approach
Most of this was working out the dependency trees and which packages to update. The vast majority of changes here are updating files for PHP 8 syntax. The main code changes I had to make as a result of PHP 8 was checking array indexes exist before accessing them as these are now thrown as warnings rather than notices. 

## Learning
It looked like some of our client unit tests weren't being run as there were a number of tests that were broken (e.g. variables not defined, prophecy expectations not set up properly). Everything seems to be running now but with a bunch of deprecation warnings so we should tackle them in a future ticket.

I also attempted to run rector on the project which had some amazing refactoring results but it seems to be conflicting with our pre-commit hooks so I reverted the changes for now. We should definitely run it in the future as some of the code changes were great and converted old-style PHP to shiny new style PHP 8 (particularly replacing switch statements to match statements which are so much easier to read).

## Checklist


Gugandeep (approved):
A lot of clean up to the formatting of code.

Looked at rector briefly - looks like a really good way to do upgrades!

mattmachell:
> Looked at rector briefly - looks like a really good way to do upgrades!

We've been using it in Sirius a bit and it does help a lot.

PR #726, opened 2021-11-18
DDPB-4171: Fix redis and add service health integration tests

acsauk:
## Purpose
Fixes the Redis health check on /manage/availability along with adding a missing integration test for overall service health.

Fixes DDPB-4171

## Approach
It turns out Redis wasn't broken - the RedisAvailability class was still using the old skool approach of getting Redis directly from the container after it was built rather than via dependency injection.

## Learning
As a side adventure (and in order to make the integration tests pass locally) I drafted up a super quick open API spec for the two Notify endpoints we use in the app so we could have a real mock Notify to call out to. It could be a little more dynamic but for the purpose of getting the integration tests to pass it's serviceable! The logic is a little backwards in that we only ever provide the base URI locally via frontend.env. If that env var is missing (which it will be in branch envs, preprod and prod) then it defaults to false which uses the hardcoded real API endpoint that's in the Notify PHP client:

```
Local:
    API key: fake (although doesn't matter as the API is mocked)
    URI: http://mock-notify-api:4010
Branch env:
    API key: dev
    URI: false
Integration/Preprod env:
    API key: dev
    URI: false
Live env:
    API key: live
    URI: false
```

I also saw that when we send an email to notify the response from them contains the full email body which could mean we could fully test our email sending on branches. This is currently not possible as templates are stored only within Notify itself so we can never be sure we're passing all the required personalisation values. A future ticket...

## Checklist

### DIFF
--- a/api/tests/Behat/behat.yml
+++ b/api/tests/Behat/behat.yml
@@ -363,6 +363,12 @@ v2-tests-goutte:
             contexts:
                 - App\Tests\Behat\v2\Reporting\Sections\ReportingSectionsFeatureContext
 
+        service-helath:
+            description: Covering service health monitoring features of the app
+            paths: [ "%paths.base%/features-v2/service-health" ]
+            contexts:
+                - App\Tests\Behat\v2\ServiceHealth\ServiceHealthFeatureContext
+
         user-management:
             description: Covering the user management features of the app (admin)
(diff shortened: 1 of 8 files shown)


Gugandeep (dismissed):
Great stuff!
Looks good to go.

Gugandeep (approved):

PR #728, opened 2021-11-22
DDPB-4139 - Create events for the different CSV uploads

Gugandeep:
## Purpose
The first part of work to allow us to create alerts when CSV uploads are not carried out on time.
Events are created and dispatched when CSVs are uploaded. These events flag the different sources the CSVs are from, and the role types the CSVs are for.
Additionally, we are also logging these events.

Fixes DDPB-4139

## Approach
Created a new `CSVUploadedEvent`.
Created a new `CSVUploadedSubscriber` to listen for `CSVUploadedEvents`.
Modified CSV processing methods to create and dispatch `CSVUploadedEvents`.
Added a function to `CSVUploadedSubscriber` to audit log the events.

Slight difficulty in figuring out the best place to create and dispatch the event for each CSV processing method.
Tried to place as near to the end of the process as possible, whilst still having enough data to figure out the source of the CSV for Lays, and the role types for Org CSVs.

## Learning

## Checklist

### Frontend


acsauk (dismissed):
Looks good to me 👍 

We've never written any integration tests for any of these custom app events which is a bit of an oversight and has bitten us before (e.g. registration emails not being sent). With the next phase of the ticket we should work out the simplest way to test this. I'm thinking Behat may be a little tricky so we could always look at Symfonys built-in [integration tests](https://symfony.com/doc/current/testing.html#integration-tests) which could allow us to listen in to events etc.

client/src/Entity/CasRec.php
```diff
@@ -0,0 +1,290 @@
...
+
+    public function getDeputyPostCode()
+    {
+        return $this->deputyPostCode;
+    }
+
+    /**
+     * @return string
+     */
+    public function getTypeOfReport()
```
coderste, line 184: Should we be type hinting the end of the function with `: string` if we are adding the DocBlock for it as well?

coderste: Considering we type hint it inside the `CSVUploadedEvent` then for consistency I think we should type hint in this file also for the functions

client/src/Event/CSVUploadedEvent.php
```diff
@@ -0,0 +1,76 @@
...
+    public function setSource(string $source): CSVUploadedEvent
+    {
+        if (in_array($source, $this->validSources)) {
+            $this->source = $source;
+        }
+
+        return $this;
+    }
+
+    public function getRoleType()
```
coderste, line 63: We know this is a string so maybe we should type hint it

client/src/EventSubscriber/CSVUploadedSubscriber.php
```diff
@@ -0,0 +1,47 @@
...
+
+    public function __construct(
+        DateTimeProvider $dateTimeProvider,
+        LoggerInterface $logger
+    ) {
+        $this->dateTimeProvider = $dateTimeProvider;
+        $this->logger = $logger;
+    }
+
+    public static function getSubscribedEvents()
```
coderste, line 29: Again I think we should be type hinting the functions

client/src/Service/OrgService.php
```diff
@@ -239,4 +250,15 @@ public function process($data, $redirectUrl)
 
         return $response;
     }
+
+    private function dispatchCSVUploadEvent($source, $roleType)
```
coderste, line 254: Here also

coderste (dismissed):
Adding comments around type hinting the functions

client/src/Entity/CasRec.php
Gugandeep: Yeah I don't think I've been consistent with the way in which I type hint.
I think the PSR-12 removes the DocBlock if we type hint in the function declaration.

Before I merge, I'll try to make the type hinting consistent, and experiment with what the formatting pre-hook does.

Gugandeep (commented):
Hopefully I've typed hinted everything appropriately.

I found that the PSR-12 Formatter that we use as a pre-hook will remove the DocBlock comments if you have type hinted in the function declaration i.e.
 `function foo(string $bar): string`
 
 It will keep the DocBlock comments however if you have not type hinted in the function declaration.
 
 Going forward, we should probably look at being consistent with our approach, either using the DocBlock or type hinting during the function declaration.

mattmachell:
> Going forward, we should probably look at being consistent with our approach, either using the DocBlock or type hinting during the function declaration.

Yeah, DocBlock existed before PHP had good type support, but we're past that now so it's better to standardise on type hints from now on.

coderste (approved):

PR #730, opened 2021-11-24
DDPB-4173: Fix report status bug

acsauk:
## Purpose
We have had reports from PA orgs that their reports are not appearing in the correct columns of their dashboards. After some investigation, it turned out when a user adds only 1 document the report section status cache does not get persisted to the database. As the dashboard view relies on the cached report status rather than calculated based on reports on the fly this was causing reports that had all required sections filled in and one document added to appear as not finished.

Fixes DDPB-4173.

## Approach
When mobbing on this we tried to see if we could update the existing flow to include this but as the issue was only affecting one section we decided to add the call to refresh the cache in the document add page rather than attempt to modify all other flows.

## Learning
The report status flow is pretty complex. I'm not sure if there is any appetite to refactor that part of the app but it would make debugging these types of bugs far easier in the future if so.

## Checklist


Gugandeep (approved):
Looks good to me.

Nice to wrap the report cache refresh in a function should we ever need it again.

PR #735, opened 2021-12-01
[DDPB-4034] Update from node-sass to sass

coderste:
This solves the dependabot alert where the `json-schema` package needed updating but because `node-sass` is not actively maintained it's easier to change packages then waiting for the update

Fixes DDPB-4034


acsauk (approved):
:shipit:

PR #739, opened 2021-12-06
DDPB-4078: Update API app to PHP 8 and Doctrine/DBAL to v3

acsauk:
## Purpose
Updates the API app to PHP 8 and bumps associated packages (most notably, Doctrine/DBAL -> v3).

Fixes DDPB-4078

## Approach
As with #725, the majority of updates were relating to non-existent indexes of arrays and some changes to the Doctrine api. There were a couple of unit test quirks to do with how the container is accessed but other than that not too painful (excluding the two packages in learning below 🤦 )

## Learning
We're seeing a bunch of output from the Symfony Process component when running tests locally in parallel. This is not something we can fix right now - see https://github.com/liuggio/fastest/issues/176.

There is also an issue with MinkExtension + PHP 8 which is causing our tests to fail when run in parallel. As the issue has been open since June we've forked the repo for a temporary fix in the hope the upstream repo will be updated. We should set some reminders to check the status of https://github.com/FriendsOfBehat/MinkExtension/pull/9 or https://github.com/FriendsOfBehat/MinkExtension/pull/8 so we can drop the reference to the fork and move back to the original package.

## Checklist

### Frontend


coderste (approved):
LGTM 👍🏻

api/composer.json
```diff
@@ -74,5 +74,11 @@
     "incenteev-parameters": {
       "file": "config/parameters.yml"
     }
-  }
+  },
+  "repositories": [
+      {
+        "url": "https://github.com/acsauk/MinkExtension.git",
```
mattmachell, line 80: Might be worth updating this to be in the MOJ account so that we have ownership if you move on at any point. Can be done as another ticket though.

acsauk: Yeah good point - will drop a ticket in now.

acsauk: https://opgtransform.atlassian.net/browse/DDPB-4184

PR #740, opened 2021-12-07
DDPB-4185: Reorganise, remove gd extension and bump client dependencies

acsauk:
## Purpose
During the PHP upgrades, I noticed we were requiring some dependencies in the main require stanza when they should be in the dev-dependencies stanza. I also took the opportunity to apply the latest package updates which created a breaking change for some old twig functions and syntax we were using.

I also experimented with removing the `gd` extension as I've never understood what we use it for (image processing seems to be encapsulated in the PDF service we use). Tests are all passing so I'm assuming it was an old dependency that was never removed once we stopped using it.

Fixes DDPB-4185.

### DIFF
--- a/api/tests/Behat/run-tests-parallel.sh
+++ b/api/tests/Behat/run-tests-parallel.sh
@@ -7,7 +7,7 @@ export APP_ENV=dev
 start=`date +%s`
 
 confd -onetime -backend env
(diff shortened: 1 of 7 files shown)


coderste (approved):
LGTM 👍🏻

PR #741, opened 2021-12-08
DDPB-3672: Remove inlined scripts to allow for easier Content Security Policy building

coderste:
## Purpose
A previous IT Health Check identified a potential weakness with the amount of inlined scripts we have in our codebase. This PR removes as many as possible (excluding Google Analytics) and covers some of the newly written JS in tests.

Fixes DDPB-3672

## Learning
@coderste and I agreed we should make hot reloading a priority piece of work before picking up any further JS tickets as currently, the app requires a full rebuild for any changes to propagate through the system. This can lead to a feedback loop of 30 seconds - 3 minutes depending on how quickly docker rebuilds the app.

We also agreed we should add some context to our JS directory structure to make it clear for our future selves what the code relates to. We made a small start with this by adding an extra entry point for some page-specific JS and explicitly requiring the script on just that page.


acsauk (approved):
:shipit:

PR #750, opened 2022-01-10
DDPB-4202: Fix client PHPUnit Deprecations

acsauk:
## Purpose
Since upgrading PHP and PHPUnit we've had very noisy unit tests with lots of deprecation warnings. Leaving these as is would mean when we come to update either PHP or PHPUnit we would have work to do to avoid breaking changes. This PR fixes the client deprecation warnings - API to follow.

Fixes DDPB-####

## Approach
It was a case of working through each deprecation to see what the issue was. The main change here is PHPUnit 9 deprecates having PHPSpec Prophecy as a bundled dependency. The recommended fix is to import a trait and explicitly call it at the start of the test.

There was also a deprecation for the PHPUnit `at()` method which required a full rewrite of the ChecklistSyncCommand test. Its not as readable but trying to chain multiple consecutive calls didn't seem possible so putting all expected API calls and returns together seemed like an OK compromise. Happy for any suggestions of other ways of doing this.

## Checklist

### DIFF
--- a/api/src/TestHelpers/ClientTestHelper.php
+++ b/api/src/TestHelpers/ClientTestHelper.php
@@ -8,12 +8,16 @@ use App\Entity\Client;
 use App\Entity\Organisation;
 use App\Entity\Report\Report;
 use App\Entity\User;
+use DateTime;
 use Doctrine\ORM\EntityManager;
 use Faker\Factory;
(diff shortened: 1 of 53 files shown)


Gugandeep (approved):
This one looks good to go

PR #751, opened 2022-01-12
DDPB-3880: Filename sanitisation

MiaGordon91:
## Purpose
An output of the IT health check identified a vulnerability with our file upload service, where users could potentially provide malicious filenames, that could be executed by our app. This PR will sanitise filenames to ensure this cannot happen.

Fixes DDPB-3880

## Approach
We were already manipulating some filenames as part of the S3 uploader, so we extended the functionality in this class to remove and replace suspicious characters.

## Checklist

### DIFF
--- a/client/src/Service/File/S3FileUploader.php
+++ b/client/src/Service/File/S3FileUploader.php
@@ -10,6 +10,7 @@ use App\Entity\ReportInterface;
 use App\Service\Client\RestClient;
 use App\Service\File\Storage\StorageInterface;
 use App\Service\Time\DateTimeProvider;
+use Exception;
 use Symfony\Component\HttpFoundation\File\UploadedFile;
 
 class S3FileUploader
@@ -53,6 +54,7 @@ class S3FileUploader
 
         $fileName = $this->fileNameFixer->addMissingFileExtension($file, $body);
         $fileName = $this->fileNameFixer->removeWhiteSpaceBeforeFileExtension($fileName);
+        $fileName = $this->fileNameFixer->removeUnusualCharacters($fileName);
 
         return [$body, $fileName];
     }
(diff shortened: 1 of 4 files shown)


client/src/Service/File/S3FileUploader.php
```diff
@@ -53,6 +54,7 @@ private function getFileBodyAndFileName(UploadedFile $file): array
 
         $fileName = $this->fileNameFixer->addMissingFileExtension($file, $body);
         $fileName = $this->fileNameFixer->removeWhiteSpaceBeforeFileExtension($fileName);
```
mattmachell, line 56: `removeWhiteSpaceBeforeFileExtension` is now potentially redundant given your new function replaces it with underscores.

client/tests/phpunit/Service/File/FileNameFixerTest.php
```diff
@@ -61,4 +62,26 @@ public function missingExtensionFilesProvider()
...
+    {
+        $actualFileName = $this->sut->removeUnusualCharacters($fileName);
+
+        self::assertEquals($expectedFileName, $actualFileName);
+    }
+
+    public function unusualCharactersProvider()
+    {
+        return [
+            'white space is transformed to underscores' => ['My File 1 2nd revision.pdf', 'My_File_1_2nd_revision.pdf'],
```
mattmachell, line 80: Good clear set of test cases! :)

client/src/Service/File/S3FileUploader.php
acsauk: Yeah we were debating between ourselves on this one and decided the resulting filename readability gains of potentially having a bunch of `_` at the end of the filename before the extension was worth leaving this in.

PR #752, opened 2022-01-12
DDPB-4205: Fix API PHPUnit deprecations

acsauk:
## Purpose
A continuation of #750 to remove deprecations from test outputs.

Fixes DDPB-4205

## Approach
API unit tests were pretty much unreadable due to there being deprecation outputs between each test case - now it's a see of `.` instead.

There was a final deprecation related to the auth component we're using in Symfony that will be removed in v5 that required a rewrite of the security config and implementation. We should switch to using the new [Symfony Security component](https://symfony.com/doc/current/security.html) when we upgrade to v6 - I'm not sure if this is possible while on 4 but its worth exploring so we can minimise upgrade PRs and potential risk.

## Checklist

### DIFF
--- a/api/composer.json
+++ b/api/composer.json
@@ -25,6 +25,7 @@
     "jms/serializer-bundle": "^4.0.1",
     "league/csv": "^9.5",
     "monolog/monolog": "^1.25.1",
+    "phpspec/prophecy-phpunit": "^2.0",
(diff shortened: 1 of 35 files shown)


acsauk:
Closing in favour of #753 due to branch env name conflicts with #750

PR #754, opened 2022-01-13
DDPB-4190 - Fix Scenario: A user adds one of each payment type

Gugandeep:
## Purpose
Occasionally when running a build, this particular test can sometimes fail.
It seems this is as a result of the steps falling out of sync with the browser, as the usual error message refers to the test step expecting the browser to be on a certain page and the browser actually being on a different one.

This ticket is to try and solve this issue.

Fixes DDPB-4190

## Approach
Ran this test a few times with various number of steps, and it did seem at a higher number of steps the error was more likely to occur.

Although it would be nice to test every type of money out payment as we currently do, all the types use the same layout and so just testing a few should prove the pages work as expected.

## Learning

## Checklist

### Frontend

### DIFF
--- a/api/tests/Behat/bootstrap/v2/Reporting/Sections/MoneyOutSectionTrait.php
+++ b/api/tests/Behat/bootstrap/v2/Reporting/Sections/MoneyOutSectionTrait.php
@@ -95,18 +95,30 @@ trait MoneyOutSectionTrait
     }
 
     /**
-     * @When I add one of each type of money out payment
+     * @When I add one type of money out payment from each category
      */
-    public function iAddOneOfEachTypeOfMoneyOutPayment()
+    public function iAddOneTypeOfMoneyOutPaymentFromEachCategory()
     {
         $this->iAmOnMoneyOutAddPaymentPage();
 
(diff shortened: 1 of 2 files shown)


acsauk:
Just one Q - how did we decide on the 9 types? Did they all individually result in some changes to forms etc?

Gugandeep:
> Just one Q - how did we decide on the 9 types? Did they all individually result in some changes to forms etc?

All of the types use the same form layout, so there didn't seem a preferred way for choosing which ones we should include in the test. So instead I added one type from each group (we group up the types on the frontend - bills, gifts etc.)

Testing each type provided the best coverage and would have been useful as a regression test to ensure that if someone ever did change one of the types to use a different form, they would be forced to update/create tests as this test would fail. The downside was how prone it was to its steps getting out of sync with the browser.

This alternative doesn't have as much coverage and so if someone does update a type (not included in this test) to use a different form, this test would not fail. We would be relying on that person to create their own testing.

A different approach would be to break the test down instead of cutting steps out. So we would have each type tested over several scenarios, instead of in one.

acsauk (approved):
Cool - I think thats a good balance between coverage and speed/reliability. 

:shipit:

PR #758, opened 2022-01-19
DDPB-3879: Reject uploaded supporting documents where extension and mimetype are mismatched

acsauk:
## Purpose
We currently support uploading a file where the mimetype is valid but file extension does not match. This results in API errors from sirius when trying to POST the document as part of the document sync process. This PR enforces a match between mimetype and file extension and rejects the file upload with a specific user-facing error.

Fixes DDPB-3879

## Approach
We re-used the existing mimetype map to easily check if mimetype and extension match - the only edge case we had to support was `jpg` vs `jpeg` - they are essentially the same file but due to historic windows extension limitations we receive both file types. We just update the file to have the full `jpeg` extension when encountering `jpg`.

We also need to keep in mind this will reject iOS/ macOS created .HEIC files, which we have seen in prod have the mimetype `image/jpeg`. WE should look into ways of converting these filetypes in a future ticket given how widespread iPhones are.

## Checklist


Gugandeep (approved):
Looks good to go!

PR #766, opened 2022-02-09
DDPB 4186 - Report for admin accounts (Joiners, Movers, Leavers) 

MiaGordon91:
## Purpose
The JML team need retention data on Digideps user accounts, so that we can help form a security process for leavers and movers. 

Fixes DDPB-4186

## Approach
- Created a new reports page (only accessible for super admins) that enable admins to view a summary of the current admin account metrics based on the requested search criteria. 
- Also moved the existing links from the button menu to the reports page to avoid overloading the existing dropdown menu.

## Checklist

### Frontend

### DIFF
--- a/api/src/Controller/StatsController.php
+++ b/api/src/Controller/StatsController.php
@@ -5,6 +5,7 @@ declare(strict_types=1);
 namespace App\Controller;
 
 use App\Repository\UserRepository;
+use App\Service\Formatter\RestFormatter;
 use App\Service\Stats\QueryFactory;
 use App\Service\Stats\StatsQueryParameters;
 use Sensio\Bundle\FrameworkExtraBundle\Configuration\Security;
@@ -43,5 +44,38 @@ class StatsController extends RestController
     public function getActiveLays()
     {
         return $this->userRepository->findActiveLaysInLastYear();
+    }
+
+    /**
+     * @Route("stats/admins/report_data", methods={"GET"})
+     * @Security("is_granted('ROLE_SUPER_ADMIN')")
+     */
+    public function getAdminUserAccountReportData(Request $request, RestFormatter $formatter)
+    {
+        $serialisedGroups = (array) $request->query->get('groups');
+        $formatter->setJmsSerialiserGroups($serialisedGroups);
+
+        $adminAccounts = $this->userRepository->getAllAdminAccounts();
+        $countOfAdminAccounts = count($adminAccounts);
+
+        $inactiveAdminAccounts = $this->userRepository->getAllAdminAccountsCreatedButNotActivatedWithin('-60 days');
+        $countOfInactiveAdminAccounts = count($inactiveAdminAccounts);
+
+        $activatedAdminAccounts = $this->userRepository->getAllActivatedAdminAccounts();
+        $countOfActivatedAdminAccounts = count($activatedAdminAccounts);
+
+        $activatedAdminAccountsNotUsedWithin90Days = $this->userRepository->getAllAdminAccountsNotUsedWithin('-90 days');
+        $countOfActivatedAdminAccountsNotUsedWithin90Days = count($activatedAdminAccountsNotUsedWithin90Days);
+
+        $activatedAdminAccountsUsedWithin90Days = $this->userRepository->getAllAdminAccountsUsedWithin('-90 days');
+        $countOfActivatedAdminAccountsUsedWithin90Days = count($activatedAdminAccountsUsedWithin90Days);
+
+        return [
+            'TotalAdminAccounts' => $countOfAdminAccounts,
+            'InactiveAdminAccounts' => $countOfInactiveAdminAccounts,
+            'ActivatedAdminAccounts' => $countOfActivatedAdminAccounts,
+            'ActivatedAdminAccountsNotUsedWithin90Days' => $countOfActivatedAdminAccountsNotUsedWithin90Days,
+            'ActivatedAdminAccountsUsedWithin90Days' => $countOfActivatedAdminAccountsUsedWithin90Days,
+        ];
     }
 }
--- a/api/src/Controller/StatsController.php
+++ b/api/src/Controller/StatsController.php
@@ -50,7 +50,7 @@ class StatsController extends RestController
      * @Route("stats/admins/report_data", methods={"GET"})
      * @Security("is_granted('ROLE_SUPER_ADMIN')")
      */
-    public function getAdminUserAccountReportData(Request $request, RestFormatter $formatter)
+    public function getAdminUserAccountReportData(Request $request, RestFormatter $formatter): array
     {
         $serialisedGroups = (array) $request->query->get('groups');
         $formatter->setJmsSerialiserGroups($serialisedGroups);
--- a/api/src/Repository/UserRepository.php
+++ b/api/src/Repository/UserRepository.php
@@ -235,15 +235,16 @@ SQL;
 
     public function getAllAdminAccountsCreatedButNotActivatedWithin(string $timeframe)
     {
-        $date = (new DateTime())->modify($timeframe)->format('Y-m-d H:i:s');
+        $date = (new DateTime())->modify($timeframe);
 
         $dql = "SELECT u FROM App\Entity\User u WHERE u.roleName IN('ROLE_ADMIN', 'ROLE_SUPER_ADMIN', 'ROLE_ADMIN_MANAGER')
                 AND u.lastLoggedIn IS NULL
-                AND u.registrationDate < '{$date}' ";
+                AND u.registrationDate < :date ";
 
         $query = $this
             ->getEntityManager()
-            ->createQuery($dql);
+            ->createQuery($dql)
+            ->setParameter('date', $date);
 
         return $query->getResult();
     }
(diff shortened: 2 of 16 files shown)


api/src/Repository/UserRepository.php
```diff
@@ -221,4 +221,70 @@ public function findActiveLaysInLastYear()
...
+
+        return $query->getResult();
+    }
+
+    public function getAllAdminAccountsUsedWithin(string $timeframe)
+    {
+        $date = (new DateTime())->modify($timeframe)->format('Y-m-d H:i:s');
+
+        $dql = "SELECT u FROM App\Entity\User u WHERE u.roleName IN('ROLE_ADMIN', 'ROLE_SUPER_ADMIN', 'ROLE_ADMIN_MANAGER')
+                AND u.lastLoggedIn > '{$date}'  ";
```
mattmachell, line 282: Would suggest using a parameter here rather than dropping the variable into the string. Even though you define the date above it's a good habit to get into and as the code drifts there's less chance of somebody forgetting that was how it was done and opening up a SQL injection vulnerability. https://www.doctrine-project.org/projects/doctrine-orm/en/2.9/reference/dql-doctrine-query-language.html#named-and-positional-parameters

api/src/Repository/UserRepository.php
```diff
@@ -221,4 +221,70 @@ public function findActiveLaysInLastYear()
...
+        $query = $this
+            ->getEntityManager()
+            ->createQuery($dql);
+
+        return $query->getResult();
+    }
+
+    public function getAllAdminAccountsUsedWithin(string $timeframe)
+    {
+        $date = (new DateTime())->modify($timeframe)->format('Y-m-d H:i:s');
```
mattmachell, line 279: You might also want to look at doing the date calculus in the SQL. https://www.postgresql.org/docs/9.1/functions-datetime.html provide a lot of options for doing current_date minus an interval.

api/tests/Unit/Entity/Repository/UserRepositoryTest.php
```diff
@@ -126,6 +126,180 @@ public function findActiveLaysInLastYear()
         self::assertNotContains($inactiveUserTwo->getId(), $resultsUserIds);
     }
 
+    /** @test */
+    public function getAllAdminAccounts()
```
mattmachell, line 130: More a for your info, you can definitely use `@test ` annotation, owever it's generally more common to just prefix the test method name with test. Phpunit automatically picks that up as a test to run and not a helper method. So `testGetAllAdminAccounts()` would work here and helps with readability of the methods. I think that's what we use across the rest of the codebase so good to be consistent, but I'll defer to @acsauk on that.

api/src/Repository/UserRepository.php
```diff
@@ -221,4 +221,70 @@ public function findActiveLaysInLastYear()
...
+
+        return $query->getResult();
+    }
+
+    public function getAllAdminAccountsNotUsedWithin(string $timeframe)
+    {
+        $date = (new DateTime())->modify($timeframe)->format('Y-m-d H:i:s');
+
+        $dql = "SELECT u FROM App\Entity\User u WHERE u.roleName IN('ROLE_ADMIN', 'ROLE_SUPER_ADMIN', 'ROLE_ADMIN_MANAGER')
+                AND u.lastLoggedIn < '{$date}'  ";
```
mattmachell, line 268: Same as below re parameters.

api/tests/Unit/Entity/Repository/UserRepositoryTest.php
coderste: We a mixture of both looking at the codebase which obviously can be quite confusing to be honest. But yeah for this file we use the `test` prefix on the function rather than the `@test` annotation comment so it would be wise to update the new tests in here to be consistent with what's already in here 👍🏻

api/src/Controller/StatsController.php
```diff
@@ -44,4 +45,37 @@ public function getActiveLays()
     {
         return $this->userRepository->findActiveLaysInLastYear();
     }
+
+    /**
+     * @Route("stats/admins/report_data", methods={"GET"})
+     * @Security("is_granted('ROLE_SUPER_ADMIN')")
+     */
+    public function getAdminUserAccountReportData(Request $request, RestFormatter $formatter)
```
coderste, line 53: I think we should add the return type to this function (which would be `array`). As we're writing more modern PHP these days then adding types to function parameters and return types is helpful to everyone when using IDEs :)

api/tests/Unit/Entity/Repository/UserRepositoryTest.php
```diff
@@ -126,6 +126,180 @@ public function findActiveLaysInLastYear()
...
+        $actualAdminUsers = $this->sut->getAllAdminAccounts();
+
+        self::assertEquals($expectedAdminUsersReturned, $actualAdminUsers);
+
+        foreach ($expectedDeputyUsersNotReturned as $deputyUser) {
+            self::assertNotContains($deputyUser, $actualAdminUsers);
+        }
+    }
+
+    /** @test */
```
coderste, line 158: Remove annotation for prefix

api/tests/Unit/Entity/Repository/UserRepositoryTest.php
```diff
@@ -126,6 +126,180 @@ public function findActiveLaysInLastYear()
...
+        $actualAdminUsers = $this->sut->getAllAdminAccountsCreatedButNotActivatedWithin('-60 days');
+
+        self::assertEquals($expectedAdminUsersReturned, $actualAdminUsers);
+
+        foreach ($expectedAdminUsersNotReturned as $user) {
+            self::assertNotContains($user, $actualAdminUsers);
+        }
+    }
+
+    /** @test */
```
coderste, line 198: Remove annotation for prefix

api/tests/Unit/Entity/Repository/UserRepositoryTest.php
```diff
@@ -126,6 +126,180 @@ public function findActiveLaysInLastYear()
...
+        $actualAdminUsers = $this->sut->getAllActivatedAdminAccounts();
+
+        self::assertEquals($expectedActiveAdminUsersReturned, $actualAdminUsers);
+
+        foreach ($expectedAdminUsersNotReturned as $user) {
+            self::assertNotContains($user, $actualAdminUsers);
+        }
+    }
+
+    /** @test */
```
coderste, line 233: Remove annotation for prefix

api/tests/Unit/Entity/Repository/UserRepositoryTest.php
```diff
@@ -126,6 +126,180 @@ public function findActiveLaysInLastYear()
...
+        $actualLoggedOutUsers = $this->sut->getAllAdminAccountsNotUsedWithin('-90 days');
+
+        self::assertEquals($expectedLoggedInAdminUsers, $actualLoggedOutUsers);
+
+        foreach ($expectedRecentlyLoggedInUsersNotReturned as $user) {
+            self::assertNotContains($user, $actualLoggedOutUsers);
+        }
+    }
+
+    /** @test */
```
coderste, line 268: Remove annotation for prefix

client/src/Service/Client/Internal/StatsApi.php
```diff
@@ -28,4 +29,16 @@ public function getActiveLayReportData()
             ['active-users']
         );
     }
+
+    /**
+     * @return array
+     */
+    public function getAdminUserAccountReportData()
```
coderste, line 36: Change the comment annotation for the PHP annotation :)

client/templates/Admin/Stats/metrics.html.twig
```diff
@@ -9,140 +9,135 @@
...
+    <div class="govuk-grid-row">
+        <div class="govuk-grid-column-full text--center">
       <span class="govuk-caption-l govuk-!-margin-top-2" id="metric-satisfaction-total-label">
         {{ 'metrics.satisfaction.totalLabel' | trans }}
       </span>
-      <span class="govuk-heading-xl govuk-!-margin-bottom-4 govuk-!-font-size-80" aria-labelledby="metric-satisfaction-total-label">
+            <span class="govuk-heading-xl govuk-!-margin-bottom-4 govuk-!-font-size-80"
+                  aria-labelledby="metric-satisfaction-total-label">
         {{ stats.satisfaction.all | default('-') }}%
       </span>
```
coderste, line 80: Formatting issue here. While this is nothing major the indentations make it easier to see where a HTML element starts and finishes :) seems like this `span` needs a tab over 👍🏻

client/templates/Admin/Stats/metrics.html.twig
```diff
@@ -9,140 +9,135 @@
...
-    </div>
+        </div>
 
-    <div class="govuk-grid-column-one-third text--center">
+        <div class="govuk-grid-column-one-third text--center">
       <span class="govuk-caption-l govuk-!-margin-top-2" id="metric-clients-total-label">
         {{ 'metrics.clients.totalLabel' | trans }}
       </span>
-      <span class="govuk-heading-xl govuk-!-margin-bottom-4" aria-labelledby="metric-clients-total-label">
+            <span class="govuk-heading-xl govuk-!-margin-bottom-4" aria-labelledby="metric-clients-total-label">
```
coderste, line 109: Think this `span` is too far over

client/src/Service/Client/Internal/StatsApi.php
```diff
@@ -9,6 +9,7 @@
 class StatsApi
```
coderste, line 9: Few formatting issues in this comment with some `span` not lining up with their closing counterparts. While this is purely a alignment issue it helps easily identify where HTML start and end when you don't have a fancy IDE/Editor that highlights them for you :)

client/templates/Admin/Stats/userAccountReports.html.twig
```diff
@@ -0,0 +1,78 @@
+{% extends '@App/Layouts/application.html.twig' %}
```
coderste, line 1: Some formatting issues within here also

coderste (commented):
Some minor changes that need to be made but other than that great work 👍🏻

api/src/Controller/StatsController.php
MiaGordon91: Just spoke to Alex, although it is an array behind the scenes it will be a response coming back so I have added 'Response' as the return type.

acsauk: My bad, looks like it does return an array and Symfony has some kind of listener that passes the array into a `Response` object. We'll set to `array`.

api/tests/Unit/Entity/Repository/UserRepositoryTest.php
MiaGordon91: Done!

acsauk:
@mattmachell @coderste - test fixes pushed by me and @MiaGordon91 has fixed up comments you have left so ready for re-review 👍

coderste (approved):

mattmachell (approved):

PR #770, opened 2022-02-11
DDPB-4169 - Alert when a data ingestion CSV has not been uploaded within the last 24 hours

Gugandeep:
## Purpose
The purpose of this ticket is to help identify when users have not uploaded CSVs containing court order information.

Occasionally, deputies will complain about not being able to register. This is sometimes the result of the CSV containing the deputies information having not been uploaded to Digideps, therefore the registration process cannot find any data to match up against.

This ticket will alert when a data ingestion CSV has not been uploaded within the last 24 hours, as well as identify which CSVs have not been uploaded.

Fixes DDPB-4169

## Approach
Carrying on from the work done in [DDPB-4139](https://github.com/ministryofjustice/opg-digideps/pull/728#issue-1060185020), we created a command to check and read events from the Audit log stream in AWS Cloudwatch.

After checking the log stream, we post messages directly to Slack to alert which CSVs have not been uploaded.

## Learning

## Checklist

### Frontend


client/src/Service/OrgService.php
```diff
@@ -224,6 +222,12 @@ protected function processChunks($chunks)
 
             $this->storeChunkOutput($upload);
             $this->logProgress($index + 1, $chunkCount);
+
+            if (!$logged) {
```
acsauk, line 226: May be missing something here but is `$logged` always false? I don't see any assignment happening in this class.

environment/check_csv_uploaded.tf
```diff
@@ -0,0 +1,116 @@
+locals {
+  check_csv_uploaded_sg_rules = {
+    ecr     = local.common_sg_rules.ecr
+    logs    = local.common_sg_rules.logs
+    s3      = local.common_sg_rules.s3
```
acsauk, line 5: Does this command need to talk to the S3?

acsauk (changes requested):
Nice job! Some comments on the tests and some security groups which may be providing too much access for the ECS service command (although happy to be proven wrong on this one if there's some tangle of dependencies!)

client/tests/phpunit/Command/CheckCSVUploadedCommandTest.php
```diff
@@ -0,0 +1,427 @@
...
+
+        $result = $this->commandTester->execute([]);
+
+        $this->assertEquals(0, $result, sprintf('Expected command to return 0, got %d', $result));
+    }
+
+    /**
+     * @test
+     */
+    public function executeOnNonBankHolidaysWhereOnlyNonCSVEventExistsSlackIsPostedTo()
```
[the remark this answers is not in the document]
Gugandeep: Yeah this test case has a really confusing name.
Essentially I wanted to have a test case for when there are events in the log stream, but they don't match any CSV upload events. It's a very unlikely scenario to be honest, but I think it was best to ensure that the command still works in this case.
In the scenario although there are events in the log stream, because none of them match the CSV upload events, we will alert for all CSVs.

acsauk (approved):
:shipit:

PR #771, opened 2022-02-14
DDPB-4233: Stop updating archived clients details during CSV upload

acsauk:
## Purpose
We noticed recently that archived clients details were still being updated during CSV upload. Once a client has been archived we should no longer touch their data.

Fixes DDPB-4233

## Approach
I've added a check at the start of each row to skip the row completely if the client has an archived_at value set. I also added some frontend feedback to show how many clients/rows have been skipped.

## Checklist


Gugandeep (dismissed):
Looks good to go.

MiaGordon91 (approved):

PR #772, opened 2022-02-14
Poc go app

acsauk:
To run you need generate a private and public key (in .pem format):

```bash
$ openssl genrsa -out private.pem 2048
$ openssl rsa -in private.pem -outform PEM -pubout -out public.pem
```

Ensure the keys are in `api/config/jwt`.

I've made one frontend route that you can access once logged in and get some simple JSON back from the new API - `v2/test-registration-api`.

See https://miro.com/app/board/uXjVOcGDxvU=/ for proposed flow.


api/src/Service/JWT/JWTService.php
```diff
@@ -0,0 +1,60 @@
...
+
+    public function createNewJWT(User $user)
+    {
+        $privateKey = file_get_contents(sprintf('%s/config/jwt/private.pem', $this->rootDir));
+        $publicKey = file_get_contents(sprintf('%s/config/jwt/public.pem', $this->rootDir));
+
+        $kid = openssl_digest($publicKey, 'sha256');
+        $payload = [
+            'username' => $user->getEmail(),
+            'userId' => $user->getId(),
```
mattmachell, line 27: So one thing I've been thinking about for JWT auth across services/products is where we have identifiers that are service specific, making them [URN](https://en.wikipedia.org/wiki/Uniform_Resource_Name)s so that we can pass down chain and still have valid auth provided by another service. Something like `urn:opg:users:digideps:123`

registration_service/Dockerfile
```diff
@@ -0,0 +1,23 @@
...
+COPY go.mod .
+#COPY go.sum .
+
+RUN go mod download
+
+COPY . .
+
+RUN CGO_ENABLED=0 GOOS=linux GOARCH=amd64 go build -a -installsuffix cgo -o /go/bin/registration_service
+
+FROM alpine:3.13
```
mattmachell, line 14: Take a look at the golang hot reload that Sirius have in place. Useful here: https://github.com/ministryofjustice/opg-sirius-supervision-deputy-hub/blob/main/docker/sirius-deputy-hub/Dockerfile#L26

acsauk: Lifesaver - rebuilding the container was getting realllllly annoying.

acsauk: Implemented ✅

api/src/Controller/JWTController.php
```diff
@@ -0,0 +1,28 @@
...
+use App\Service\JWT\JWTService;
+use Symfony\Bundle\FrameworkBundle\Controller\AbstractController as SymfonyAbstractController;
+use Symfony\Component\HttpFoundation\JsonResponse;
+use Symfony\Component\Routing\Annotation\Route;
+
+class JWTController extends SymfonyAbstractController
+{
+    private JWTService $JWTService;
+
+    public function __construct(JWTService $JWTService)
```
mattmachell, line 16: Use constructor promotion to reduce boilerplate? https://stitcher.io/blog/constructor-promotion-in-php-8

api/src/Service/JWT/JWTService.php
mattmachell: Yeah, where we have shared services like a PDF generator, say, it would allow a token generated by either Sirius or DigiDeps to be valid.

api/src/Controller/JWTController.php
mattmachell: Yeah, it's quite a hard habit to break in after years of setting constructor $thing to $this->thing.

api/src/Service/JWT/JWTService.php
```diff
@@ -0,0 +1,57 @@
...
+            'aud' => 'registration_service',
+            'iat' => strtotime('now'),
+            'exp' => strtotime('+1 hour'),
+            'nbf' => strtotime('-10 seconds'),
+            'iss' => 'digideps',
+            'sub' => $user->getEmail(),
```
mattmachell, line 30: Worth noting that jwt's don't encrypt, they just validate the payload, so email references might be considered a potential for data leakage if the token got exposed.

acsauk: True - I guess we could drop the two user fields and substitute the value of `sub` to be user ID and then grab from the DB where needed

registration_service/main.go
```diff
@@ -0,0 +1,165 @@
...
+
+	if err != nil {
+		log.Fatalf("Cannot parse JWT from headers: %v", err)
+	}
+
+	// Get public key from frontend client (will likely be standalone service in the future)
+	JWKs, err := fetchJWKsByJKU(JWT)
+
+	if err != nil {
+		log.Fatalf("could not create jwks request: %v", err)
```
wordshaker, line 37: really tiny this but this is the only error you don't use sentence case with

registration_service/main.go
```diff
@@ -0,0 +1,165 @@
...
+	tr := &http.Transport{
+		TLSClientConfig: &tls.Config{InsecureSkipVerify: true},
+	}
+
+	client := &http.Client{Transport: tr}
+
+	req, err := http.NewRequest("GET", jku, nil)
+
+	if err != nil {
+		return nil, fmt.Errorf("could not create jwks request: %w", err)
```
wordshaker, line 108: I stand corrected. You use different casing on different errors

coderste:
Conflicts also 👍🏻

registration_service/main.go
```diff
@@ -0,0 +1,165 @@
...
+	"gopkg.in/square/go-jose.v2/jwt"
+	"io/ioutil"
+	"log"
+	"net/http"
+	"strings"
+	"time"
+)
+
+type regServiceResponse struct {
+	Name      string
```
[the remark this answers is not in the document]
acsauk: Nope, good call 👍

PR #774, opened 2022-02-15
DDPB-4236, DDPB-4187: Update client benefits check section with new layout, field and content

acsauk:
## Purpose
I'm fully aware this is a ridiculous amount of file changes for an update like this but...I have no explanation 😬 .

Fixes DDPB-4236, DDPB-4187.

## Approach
Lots and lots and lots of refactor -> rename, searching for the word `income` and other variations of strings.

New field was a cinch in comparison and handily the behat tests only took 30 minutes to update which suggests they were written the right way 💪 .

## Checklist

### Frontend


Gugandeep (approved):
Looks good to me.

Could have been a smaller change if it was just content changes, but I think changing the file names and column names is a good idea to prevent any confusion in the future around this section.

coderste (approved):

PR #775, opened 2022-02-18
DDPB-4175: Return JSON for report rather than whole objects, fix CSV download

acsauk:
## Purpose
As we were working with hydrating full models and returning them from the API to the frontend, our API was timing out whenever more than a day or two of reports was selected. This change returns an array of raw data in JSON that we manually parse into the CSV on the frontend.

Fixes DDPB-4175

## Approach
From some manual testing it looked like the slowdown was in Doctrine hydrating the objects fully to then turn back into JSON. We now skip this step and just return the raw data. I tested this locally with 10000 user research responses and it was returning the CSV is around 3 seconds. Our timeout for the API is 20 seconds so we should be able to go comfortably above this without timing out.

Unfortunately, I wasn't able to come up with a way for testing this without having a very slow running test. To create the required data in the DB (User, Client, Report, Report Submission, Satisfaction and User Research Response) it was taking around 45 seconds to create 2000 records. We may want to look in the future into having some raw SQL persisted somewhere that can create these records but as this is a super-admin-only feature I didn't want to sink too much time into it.

## Checklist

### DIFF
--- a/api/src/Controller/UserResearchController.php
+++ b/api/src/Controller/UserResearchController.php
@@ -11,6 +11,7 @@ use App\Service\Formatter\RestFormatter;
 use DateTime;
 use RuntimeException;
 use Sensio\Bundle\FrameworkExtraBundle\Configuration\Security;
+use Symfony\Component\HttpFoundation\JsonResponse;
 use Symfony\Component\HttpFoundation\Request;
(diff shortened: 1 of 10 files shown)


coderste (approved):
👍🏻

PR #785, opened 2022-03-09
DDPB-4245 - Cloud9 memberships

xaithe:
## Purpose
Handle Cloud9 access programatically

Fixes DDPB-4245

## Approach
The "cloud9_users" list in `shared/variables.tf` is iterated through in the `aws_cloud9_environment_membership` block in `shared/cloud9.tf`, assigning read/write c9 access to each listed user. User name has to be the same as AWS

## Learning

## Checklist

### Frontend

### DIFF
--- a/shared/variables.tf
+++ b/shared/variables.tf
@@ -28,4 +28,10 @@ locals {
   }
 
   s3_bucket = local.account.name == "production" ? "${local.account.name}02" : local.account.name
+
+  cloud9_users = [
+    "alex.saunders",
+    "gugandeep.chani",
+    "stephen.hinett",
+  ]
 }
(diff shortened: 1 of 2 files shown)


Gugandeep (approved):
Looks good to me

shared/variables.tf
```diff
@@ -28,4 +28,10 @@ locals {
   }
 
   s3_bucket = local.account.name == "production" ? "${local.account.name}02" : local.account.name
+
+  cloud9_users = [
+    "alex.saunders",
+    "gugandeep.chani",
+    "stephen.hinett",
```
acsauk, line 35: Could we add Mia in here? Probably worth having you in there too @xaithe if you need to do any DB work in the future.

acsauk (changes requested):
Looks good! Just one update required - could you add a PR description to give this some context for when our future selves look back at this PR and wonder what its about?

shared/variables.tf
Gugandeep: Think Tom will have access as he has set himself to Owner.

PR #786, opened 2022-03-09
DDPB-4269: Give CSV upload check open internet access for bank holidays API

acsauk:
## Purpose
This became a bit more involved than providing open internet access. We had a few places where error handling and other AWS permissions needed updating but from testing on a branch we're getting the output we expect now.

Fixes DDPB-4269.

## Approach
I did a small amount of refactoring in the slack posting code to make it clear what each step was doing. We had also hard coded the log group so this is now injected from an env var like the rest of our logging code.

## Learning
AWS docs for PHP are not always so clear on the type of errors to expect. I had to manually play around with `awslocal` in the localstack container to see the types of AWS exceptions that were returned for certain situations.

## Checklist


coderste (approved):
LGTM 👍🏻

xaithe (approved):

PR #790, opened 2022-03-11
DDPB-4286: Allow null values for whoReceivedMoney to support migrations

acsauk:
## Purpose
As part of expanding the new benefits section we added a field to record where money was received from. The migration to create the new column ran fine locally and in envs where the DB was dropped and created but in preproduction we have some existing rows in the DB that would have NULL values in this column post migration. This change removes the DB constraint and relies on code constraints instead to allow the migration to run.

Fixes DDPB-4286

## Checklist


Gugandeep (approved):
Ah, first noticed this issue when we were trying to create fixtures.
We updated the fixtures to set `who_received_money`.

Looks good to go!

coderste (approved):

PR #791, opened 2022-03-11
DDPB-4285: Add report to metrics page for total liquid and non-liquid assets

acsauk:
## Purpose
OPG require a report showing the total value of liquid and non-liquid assets managed in digideps broken down by type. This change adds a CSV download link to the metrics admin page showing the values.

Fixes DDPB-4285

## Approach
I started to just write an SQL query for this report but it was getting complex once I started to try to exclude NDRs that had already been submitted so I pivoted to creating a downloaded report via code.

Given the tight deadlines I've not TDDed this or written tests but have carried out some manual checks to ensure we're getting back the expected figures. We should play a follow-up ticket to add tests to this once we have the time next week.

## Checklist

### DIFF
--- a/api/src/Controller/StatsController.php
+++ b/api/src/Controller/StatsController.php
@@ -4,25 +4,25 @@ declare(strict_types=1);
 
 namespace App\Controller;
 
+use App\Repository\NdrRepository;
+use App\Repository\ReportRepository;
 use App\Repository\UserRepository;
 use App\Service\Formatter\RestFormatter;
 use App\Service\Stats\QueryFactory;
 use App\Service\Stats\StatsQueryParameters;
 use Sensio\Bundle\FrameworkExtraBundle\Configuration\Security;
+use Symfony\Component\HttpFoundation\JsonResponse;
 use Symfony\Component\HttpFoundation\Request;
 use Symfony\Component\Routing\Annotation\Route;
 
 class StatsController extends RestController
 {
-    private QueryFactory $QueryFactory;
-    private UserRepository $userRepository;
-
     public function __construct(
-        QueryFactory $QueryFactory,
-        UserRepository $userRepository
+        private QueryFactory $QueryFactory,
+        private UserRepository $userRepository,
+        private ReportRepository $reportRepository,
+        private NdrRepository $ndrRepository,
     ) {
-        $this->QueryFactory = $QueryFactory;
-        $this->userRepository = $userRepository;
     }
 
     /**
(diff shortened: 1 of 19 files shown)


api/src/Entity/Ndr/AssetProperty.php
```diff
@@ -393,4 +394,16 @@ public function isEqual(AssetInterface $asset)
             $asset->getAddress2() === $this->getAddress2() &&
             $asset->getPostcode() === $this->getPostcode();
     }
+
+    /**
+     * @return float|null
+     */
+    public function getValueTotal()
```
coderste, line 401: Can we not use the PHP `public function getValueTotal(): float | null` here rather than the comment?

api/src/Repository/ReportRepository.php
```diff
@@ -230,4 +234,41 @@ public function countAllEntities()
...
     }
+
+    /**
+     * @return Report[]
+     */
+    public function getAllSubmittedReportsWithin12Months(string $deputyType)
+    {
+        $oneYearAgo = new DateTime('-1 year');
+
+        switch ($deputyType) {
```
coderste, line 245: Considering these are only a single-line expression maybe it's worth changing the `switch` for a `match` as it's shorter and just looks cleaner as you don't need all the breaks

```
$types = match ($deputyType) {
    'LAY' => Report::getAllLayTypes(),
    'PROF' => Report::getAllProfTypes(),
    'PA' => Report::getAllPaTypes(),
    default => [],
};
```

More on PHP8 match vs switch: https://stitcher.io/blog/php-8-match-or-switch

client/src/Service/Csv/AssetsTotalsCSVGenerator.php
```diff
@@ -0,0 +1,45 @@
...
+
+    public function __construct(CsvBuilder $csvBuilder)
+    {
+        $this->csvBuilder = $csvBuilder;
+    }
+
+    /**
+     * @return string
+     */
+    public function generateAssetsTotalValuesCSV(array $assetsTotals)
```
coderste, line 19: Remove comment for actual PHP return type hinting

coderste (changes requested):

api/src/Controller/StatsController.php
```diff
@@ -78,4 +78,68 @@ public function getAdminUserAccountReportData(Request $request, RestFormatter $f
...
+    public function getAssetsTotalValueData()
+    {
+        $ret = [
+            'lays' => ['liquid' => 0, 'non-liquid' => 0],
+            'profs' => ['liquid' => 0, 'non-liquid' => 0],
+            'pas' => ['liquid' => 0, 'non-liquid' => 0],
+            'grandTotal' => 0,
+        ];
+
+        $lays = $this->reportRepository->getAllSubmittedReportsWithin12Months('LAY');
```
mattmachell, line 95: What's the indexing like on these tables? Any potential for slowness on the whole data set or impact on other parts of the system?

api/src/Controller/StatsController.php
```diff
@@ -78,4 +78,68 @@ public function getAdminUserAccountReportData(Request $request, RestFormatter $f
...
+        $profs = $this->reportRepository->getAllSubmittedReportsWithin12Months('PROF');
+        $pas = $this->reportRepository->getAllSubmittedReportsWithin12Months('PA');
+        $ndrs = $this->ndrRepository->getAllSubmittedNdrsWithin12Months();
+
+        $layClientIds = [];
+
+        foreach ($lays as $layReport) {
```
mattmachell, line 102: Given the size of these chunks, worth checking how much memory this takes up in PHP. Might be better pushed down to SQL.

acsauk: I was trying to avoid the complexity that's going to be involved with handling the different report types and then excluding NDRs for reports that we've already counted. 

I've taken a look in preprod and in the last 12 months we have 32K~ submitted reports so shouldn't have any systemwide impact querying that many rows. I'd expect PHP to be able to handle chunks of that size without running out of memory but will run on preprod to be sure before releasing though. 

Given the short notice for the request (Monday) I've not added tests for large data sets which I hate but just trying to prioritise against the CSV work.

api/src/Controller/StatsController.php
acsauk: We index on `submit_date`, `report_status_cached` and `submitted`. Querying preprod was pretty instantaneous in Cloud9 so I'm comfortable this shouldn't impact the live service.

api/src/Controller/StatsController.php
acsauk: Interesting...will see what happens on preprod 🤞

coderste (approved):

PR #794, opened 2022-03-15
DDPB-4291: Fix assets report

acsauk:
## Purpose
The original code to generate a report of asset values relied on iterating over PHP entities and summing values. This worked fine locally with a dataset of 10 reports and NDRs but took too long to response once run on prod. This change pushes the summing logic down into DQL to make sure we can actually run the report on prod.

Fixes DDPB-4291.

## Learning
Having relied on ORMs so heavily in the past whenever I get a data ticket I tend to avoid trying to come up with a DQL/SQL solution as its out of my comfort zone. This (along with previous time out bugs) should serve as a reminder that its not scary and SO is my friend 😂 . Interstingly this solution took around the same amount of time to come up with compared to the original.

Also, test against large datasets!

## Checklist

### DIFF
--- a/.prettierrc.json
+++ b/.prettierrc.json
@@ -1,3 +1,1 @@
-{
-  "twigPrintWidth": 140
-}
+{}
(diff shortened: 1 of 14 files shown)


Gugandeep (approved):
Looks good to go.

PR #798, opened 2022-03-16
DDPB-4278: Set client benefits section section release date to 16 March 2022

acsauk:
## Purpose
We've had the new benefits check section toggled off while developing for full reports. This change enables the section for all reports (plus 60 days from the date set here). It's already live for NDRs.

Fixes DDPB-4278.


Gugandeep (approved):
Looks good to go!

PR #799, opened 2022-03-17
DDPB-4288: Allow checklists linked to a discharged client to be synced with Sirius

acsauk:
## Purpose
Following some reports from case managers on checklists failing to sync we investigated and noticed that checklists linked to a report belonging to a discharged deputy (e.g. client soft deleted) were throwing errors during the sync process. This change allows checklists with discharged deputies to be synced, adds a sync status to the checklist and covers the feature in behat tests.

Fixes DDPB-4288.

## Checklist

### DIFF
--- a/api/src/Controller/Report/ReportController.php
+++ b/api/src/Controller/Report/ReportController.php
@@ -931,6 +931,7 @@ class ReportController extends RestController
             $filter->disableForEntity(Client::class);
 
             $reports[] = $this->findEntityBy(Report::class, $reportId);
+            $filter->enableForEntity(Client::class);
         }
 
         $this->formatter->setJmsSerialiserGroups($this->checklistGroups);
--- a/api/src/Controller/Report/ReportController.php
+++ b/api/src/Controller/Report/ReportController.php
@@ -4,6 +4,7 @@ namespace App\Controller\Report;
 
 use App\Controller\RestController;
 use App\Entity as EntityDir;
+use App\Entity\Client;
 use App\Entity\Report\Report;
 use App\Entity\User;
 use App\Exception\UnauthorisedException;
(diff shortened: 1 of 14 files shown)


client/templates/Admin/Client/Report/sidebar/_statuses.html.twig
```diff
@@ -21,6 +21,10 @@
         <br>
         {{ checklist.submittedBy.fullName}}, {{ checklist.submittedBy.roleFullName }}
     </p>
-{% endif %}
 
-{#<hr class="govuk-section-break govuk-section-break--m govuk-section-break--visible">#}
```
Gugandeep, line 26: Just to confirm - this is a commented out line that we're removing?

api/src/Controller/Report/ReportController.php
```diff
@@ -926,6 +927,9 @@ public function getReportsWithQueuedChecklists(Request $request): array
 
         $reports = [];
         foreach ($queuedReportIds as $reportId) {
+            $filter = $this->em->getFilters()->getFilter('softdeleteable');
+            $filter->disableForEntity(Client::class);
```
Gugandeep, line 931: Annoying to have to disable the soft delete filter to do stuff like this.
Think it's because in Digideps we still make use of things that are soft deleted, which may not be the expected behaviour when you soft delete something.

I've seen places in our code where we enable the filter again afterwards.  
[See here for example](https://github.com/ministryofjustice/opg-digideps/blob/main/api/src/Repository/ClientRepository.php#L52)

 I'm not 100% if this is needed or not. Unfortunately haven't been able to find any resources to dictate whether we need to enable afterwards or not.

Gugandeep (dismissed):

Gugandeep (approved):
Looks good to go!

PR #804, opened 2022-03-24
DDPB - 4208: Investigate and fix submissions appearing in New column

MiaGordon91:
## Purpose
As a hangover from when document syncing with Sirius was a manual process, we have a New column on the submissions page. After implementing the automated sync process the submission/document flow was changed so the first status a Document will have is Queued rather than New. However, we are still seeing the occasional submission appear in the New column.

Fixes DDPB-4208

## Approach
- Tested various scenarios to try and identify how a new submission lands in the 'New' column on the admin side.
- Identified that a submission lands in the 'new' column when there are no documents attached.
- Added additional logic to the api side to prevent a submission from saving to the database if no documents are attached. 

## Learning
- As mentioned above, we identified that a submission lands in the 'new' column when there are no documents attached. 
- This happens when a user submits a document and then clicks back to the previous page and submits again.

## Checklist


acsauk (dismissed):
:shipit:

Gugandeep (approved):

client/src/Exception/NoDocumentAttached.php
```diff
@@ -0,0 +1,9 @@
+<?php
+
+declare(strict_types=1);
+
+namespace App\Exception;
+
+class NoDocumentAttached extends \RuntimeException
+{
+}
```
acsauk, line 9: We never ended up using this so we can drop it.

acsauk (approved):
:shipit:

PR #816, opened 2022-04-08
DDPB-4312: Stabilise CI tests and only alert container vulns on main

acsauk:
## Purpose
We've had extremely flaky integration tests over the past couple of weeks which is starting to have a material impact to releasing code. This ticket aims to stabilise the tests to ensure we can release when we need to and have confidence that our app is working as intended.

Fixes DDPB-4312

## Approach
I broke this problem down as follows:

- Reverted to running the tests in serial in a single job (where they all passed but took 10+ minutes)
- Stopped scaling the ECS tasks prior to tests (where they all passed but ran very slowly)
- Scaled the ECS tasks to 2 and then 3 where I started to see some intermittent failures

I looked at the failures that were happening and they all eventually fell under the category of entities not existing in the DB. To make sure it wasn't our test code using old IDs I queried the DB using the ids from failing tests and the entities really weren't in the DB which led me further up the test flow to the test scripts. I noticed that for the sequential tests we have a step that loads the test fixtures (and truncates all data from the DB) that runs when the test script starts up. Based on when the script starts we could have instances where a different test container has created an entity that is then wiped by the sequential container. Removing the fixture loading step (which isn't even required any more for our integration tests) seems to have calmed things down. 

I tested out a number of replica levels and I've settled on 6 to give us a slightly faster test run time while not going overboard on task replicas. See below for comparisons:

Screenshot 2022-04-11 at 12 09 16

I also took the opportunity to move all tests that manually run Symfony commands (that update entity statuses at an app level) into the sequential tests to ensure there aren't any conflicting statuses between test containers.

Also included on this branch is disabling posting security vulnerabilities on PR branches. Going forward we will only get alerts when building our app containers on `main`

## Checklist


Makefile
```diff
@@ -97,7 +97,8 @@ else
 endif
 
 behat-tests-v2-goutte-parallel: up-app-integration-tests reset-fixtures disable-debug ##@behat Run the integration tests in parallel
-	docker-compose -f docker-compose.yml -f docker-compose.dev.yml run --rm test sh ./tests/Behat/run-tests-parallel.sh --tags @v2
+	docker-compose -f docker-compose.yml -f docker-compose.dev.yml run --rm test sh ./tests/Behat/run-tests.sh --tags @v2_sequential
+	docker-compose -f docker-compose.yml -f docker-compose.dev.yml run --rm test sh ./tests/Behat/run-tests-parallel.sh --tags @v2&&~@v2_sequential
```
Gugandeep, line 101: `@v2&&~@v2_sequential`

Not sure if this is a mistype or an actual operator. Wanted to raise it just in case

acsauk: Yeah totally intentional - I'm saying here to run all tests with a v2 tag but to skip any that have v2_sequential (as we want them to run in sequence). This puts us more in line with how our tests run in CI which is nice.

Gugandeep (approved):

PR #818, opened 2022-04-13
DDPB 4218 - 'New' tabs visibility on Submissions page toggled based on Document Sync Enabled flag

MiaGordon91:
## Purpose
The New tab on the submissions page was designed for manually syncing document between Digideps and Sirius. It doesn’t serve a purpose while we have the document sync feature enabled but if there's ever an issue with Sirius we need the ability to revert to manual syncing. We should make the New tab conditionally revealed based on value of the document sync enabled feature flag.

Fixes DDPB-4218

## Approach
'New' tab now only appears when document sync enabled flag is disabled. The page is conditionally rendered based on the value of the Parameter Store.

## Checklist


api/tests/Behat/bootstrap/v2/Common/BaseFeatureContext.php
```diff
@@ -111,29 +112,18 @@ class BaseFeatureContext extends MinkContext
...
-    private ReportTestHelper $reportTestHelper;
-
     public function __construct(
-        FixtureHelper $fixtureHelper,
-        KernelInterface $symfonyKernel,
-        EntityManagerInterface $em,
-        ReportTestHelper $reportTestHelper
+        private FixtureHelper $fixtureHelper,
+        private KernelInterface $symfonyKernel,
+        private EntityManagerInterface $em,
```
acsauk, line 118: Ah my bad, I should have made this protected rather than private.

client/tests/phpunit/Service/ParameterStoreServiceTest.php
```diff
@@ -37,4 +38,35 @@ public function getParameter()
 
         self::assertEquals('result', $sut->getParameter('test-flag'));
     }
+
+    /**
+     * @dataProvider parameterDataProvider
+     * @test
+     */
+    public function addParameter($parameterName, $parameterValue)
```
acsauk, line 46: We can revert the name of this test back to `putParameter` now.

client/tests/phpunit/Service/ParameterStoreServiceTest.php
```diff
@@ -37,4 +38,35 @@ public function getParameter()
...
+        $sut->putParameter($parameterName, $parameterValue);
+    }
+
+    public function parameterDataProvider()
+    {
+        return [
+            'document sync set to true' => ['document-sync', 1],
+            'checklist sync set to false' => ['checklist-sync', 0],
+        ];
+    }
```
acsauk, line 71: And for completeness sake we should cover the `putFlag` function we wote under a test here (basically the same layout as this but varying the `$flagPrefix` var rather than the `$parameterPrefix` var.

We should also add this test to the API app seeing as we duplicated this code.

coderste (dismissed):

api/src/Service/ParameterStoreService.php
```diff
@@ -39,4 +41,16 @@ public function getFeatureFlag(string $flagKey)
 
         return $flag['Parameter']['Value'];
     }
+
+    public function putParameter($parameterName, $parameterValue)
```
coderste, line 45: Should we be type hinting these parameters?

api/src/Service/ParameterStoreService.php
```diff
@@ -39,4 +41,16 @@ public function getFeatureFlag(string $flagKey)
...
+    {
+        $parameterName = $this->parameterPrefix.$parameterName;
+        $this->ssmClient->putParameter(['Name' => $parameterName, 'Value' => $parameterValue, 'Overwrite' => true]);
+    }
+
+    public function putFeatureFlag($flagName, $flagValue)
```
coderste, line 51: Same again here as above

client/src/Service/ParameterStoreService.php
```diff
@@ -47,4 +47,16 @@ public function getFeatureFlag(string $flagKey)
 
         return $flag['Parameter']['Value'];
     }
+
+    public function putParameter($parameterName, $parameterValue)
```
coderste, line 51: More type hinting

client/src/Service/ParameterStoreService.php
```diff
@@ -47,4 +47,16 @@ public function getFeatureFlag(string $flagKey)
...
+    {
+        $parameterName = $this->parameterPrefix.$parameterName;
+        $this->ssmClient->putParameter(['Name' => $parameterName, 'Value' => $parameterValue, 'Overwrite' => true]);
+    }
+
+    public function putFeatureFlag($flagName, $flagValue)
```
coderste, line 57: One more

coderste (changes requested):
Everything looks good to me but think we should be type hinting a few parameters :)

coderste (approved):

PR #824, opened 2022-05-03
DDPB-3342 - Add audit log around creation of organisations

MiaGordon91:
Create a record when organisations are created so the PM is aware of when they were initially added.

Fixes DDPB-3342

## Approach
- New event and subscriber has been created for the creation of organisations
- AuditEvents has been updated
- EventDispatcher has been added to the manual process and CSV upload options available when creating an organisation in the admin side of the app 

## Checklist


client/src/Service/Audit/AuditEvents.php
```diff
@@ -251,11 +253,35 @@ public function csvUploaded(
...
+     *
+     * @return array|string[]
+     *
+     * @throws \Exception
+     */
+    public function orgCreated(string $trigger, User $currentUser, Organisation $organisation): array
+    {
+        $event = [
+            'trigger' => $trigger,
+            'current_user' => $currentUser->getEmail(),
```
acsauk, line 269: I think this is more like `created_by` as in either case it will be someone uploading a CSV or clicking a button.

api/src/v2/Registration/Uploader/OrgDeputyshipUploader.php
```diff
@@ -134,7 +134,7 @@ private function handleOrganisation(OrgDeputyshipDto $dto)
 
             $this->currentOrganisation = $organisation;
 
-            $this->added['organisations'][] = $organisation->getId();
+            $this->added['organisations'][] = $organisation;
```
acsauk, line 137: Hmmm, we may need to see what impact this has when running at scale. If we have 1000s of orgs in here we may get out of memory errors.

client/src/EventSubscriber/OrgCreatedSubscriber.php
```diff
@@ -0,0 +1,47 @@
...
+{
+    /** @var DateTimeProvider */
+    private $dateTimeProvider;
+
+    /** @var LoggerInterface */
+    private $logger;
+
+    public function __construct(
+        LoggerInterface $logger,
+        DateTimeProvider $dateTimeProvider
```
acsauk, line 23: We can use constructor property promotion here to remove some boilerplate:

```suggestion
        private LoggerInterface $logger,
        private DateTimeProvider $dateTimeProvider
```

And then drop lines 15-19 and 25-26.

client/src/Service/Audit/AuditEvents.php
```diff
@@ -251,11 +253,35 @@ public function csvUploaded(
...
+            'organisation_id' => $organisation->getId(),
+            'organisation_name' => $organisation->getName(),
+            'organisation_identifier' => $organisation->getEmailIdentifier(),
+            'organisation_status' => $organisation->isActivated(),
+            'created_on' => $this->dateTimeProvider->getDateTime()->format(DateTime::ATOM),
```
acsauk, line 274: Good level of context added here 👍

acsauk (changes requested):
Excellent work 🙌  this all looks great and gets us the logs where we need them. Could we get a couple of tests for:

- `AuditEvents::orgCreated()`
- `OrgCreatedSubscriber::getSubscribedEvents()`
- `OrgCreatedSubscriber:: auditLog()`

coderste (approved):

acsauk (approved):
Great job :shipit: 

Although lets hold off merging until we've released the CSV code 👍

PR #826, opened 2022-05-03
Patches libraries on alpine

mattmachell:
## Purpose

Updates `curl gzip xz xz-libs libcurl` from alpine defaults


acsauk (approved):
:shipit:

PR #830, opened 2022-05-09
[DDPB-4206] Track how often users are completing information in the new benefits section

coderste:
## Purpose
We are implementing the first new section in a long time and it would be good to use this as an opportunity to get some stats on a new piece of work. 

It is difficult to find information about how users typically respond to this question because it is less commonly answered with supporting information and it is difficult to track the information across paper only reports. 

Fixes DDPB-4206

## Approach

Stuff I'm not too keen on in this PR but was unable to figure out a different way to do it:
- `api/src/Controller/StatsController::getBenefitsReportMetrics()` don't like the way I have to filter the results out based on the query parameters sent in the download request. Feel like the foreach is doing too much work and it's duplicating a whole section but was unsure how to filter the results when it came done to removing them based on the date filter
- `api/src/Repository/ReportRepository.php::getBenefitsReportMetrics()` this is a plain SQL query as using DQL wasn't allowing me to `LEFT JOIN` the tables the same and therefore was putting a whole report within the metric because the `report_id` inside `client_benefits_check` table is associated with a whole report. Although looking around I think SQL is the only way to go for this.

## Learning

## Checklist
- [ ] I have added tests to prove my work (TODO added one test but this could probably do with more)

### DIFF
--- a/api/src/Controller/StatsController.php
+++ b/api/src/Controller/StatsController.php
@@ -135,4 +135,28 @@ class StatsController extends RestController
 
         return new JsonResponse($ret);
     }
+
+    /**
+     * @Route("stats/report/benefits-report-metrics", name="benefits_reoprt_metrics")
+     * @Security("is_granted('ROLE_SUPER_ADMIN')")
+     *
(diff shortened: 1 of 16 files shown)


acsauk (approved):
:shipit:

PR #831, opened 2022-05-09
DDPB-4239 - Skip rows with blank fields which are mandatory

Gugandeep:
## Purpose
Found an issue where when transforming rows from the CSV into the DTO, an error occurred resulting in the upload failing.

Discovered that the CSV sometimes contains rows with blank fields. This error in particular was caused by having a blank row for the `ReportType` field in the CSV.

Logic was unable to determine report type, and was not throwing any exceptions.

Fixes DDPB-4329

## Approach

## Learning

## Checklist

### Frontend

### DIFF
--- a/api/src/v2/Registration/Assembler/SiriusToLayDeputyshipDtoAssembler.php
+++ b/api/src/v2/Registration/Assembler/SiriusToLayDeputyshipDtoAssembler.php
@@ -87,6 +87,10 @@ class SiriusToLayDeputyshipDtoAssembler implements LayDeputyshipDtoAssemblerInte
             default => false
         };
 
-        return $supported ? $reportType : '';
+        if ($supported) {
+            return $reportType;
+        } else {
+            throw new InvalidArgumentException();
+        }
     }
 }


api/src/v2/Registration/Assembler/LayDeputyshipDtoCollectionAssembler.php
```diff
@@ -10,30 +10,26 @@ class LayDeputyshipDtoCollectionAssembler
     /** @var LayDeputyshipDtoAssemblerInterface */
     private $layDeputyshipDtoAssembler;
 
-    /**
-     * @param LayDeputyshipDtoAssemblerInterface $layDeputyshipDtoAssembler
-     */
     public function __construct(LayDeputyshipDtoAssemblerInterface $layDeputyshipDtoAssembler)
```
coderste, line 13: Can change this to the constructor property promotion
```
public function __construct(private LayDeputyshipDtoAssemblerInterface $layDeputyshipDtoAssembler)
    {
    }
```

api/src/v2/Registration/Assembler/LayDeputyshipDtoCollectionAssembler.php
```diff
@@ -10,30 +10,26 @@ class LayDeputyshipDtoCollectionAssembler
...
     {
         $this->layDeputyshipDtoAssembler = $layDeputyshipDtoAssembler;
     }
```
acsauk, line 16: Could we use constructor property promotion here to drop out some boilerplate?

api/src/v2/Registration/Assembler/LayDeputyshipDtoCollectionAssembler.php
```diff
@@ -10,30 +10,26 @@ class LayDeputyshipDtoCollectionAssembler
...
+        $skipped = [];
         $collection = new LayDeputyshipDtoCollection();
 
-        foreach ($data as $uploadRow) {
+        foreach ($data as $line => $uploadRow) {
             $item = $this->layDeputyshipDtoAssembler->assembleFromArray($uploadRow);
             if ($item instanceof LayDeputyshipDto) {
                 $collection->append($item);
+            } else {
+                $skipped[] = sprintf('SKIPPED LINE %d:', $line + 2);
```
acsauk, line 28: Nice, this is super useful feedback.

api/src/v2/Registration/Assembler/SiriusToLayDeputyshipDtoAssembler.php
```diff
@@ -87,6 +87,10 @@ private function determineReportTypeIsSupported(?string $reportType)
             default => false
         };
 
-        return $supported ? $reportType : '';
```
coderste, line 90: Maybe it's the Go side of me but I believe writing something like:
```
if (!$supported) {
    throw new InvalidArgumentException();
}

return $reportType
```

Feels cleaner (to me anyways), as from a Go perspective you're taught to deal with any errors you may potentially have first and then if you have none you just return what's expected. 

Although I know this is not Go I still feel like that way of thinking feels more natural to me. Check for an error first -> handle the error if there is one -> if no error return what we're expecting 

This way you also make the code a little cleaner as there's no `else` statement :)

api/tests/Behat/bootstrap/v2/Registration/IngestTrait.php
```diff
@@ -808,4 +832,28 @@ public function namedDeputyForCaseNumberShouldHaveAddress(string $caseNumber, st
...
+            throw new BehatException($missingDivMessage);
+        }
+
+        $infoMessageFound = str_contains($flashDiv->getHtml(), $expectedMessage);
+
+        if (!$infoMessageFound) {
+            throw new BehatException(sprintf('The error summary did not contain the expected error message. Expected: "%s", got (full HTML): %s', $errorMessage, $errorHtml));
+        }
+    }
 }
```
acsauk, line 859: I think `AssertTrait::assertOnAlertMessage()` may work for this OR could be tweaked slightly to save repetition?

coderste (commented):
Added some comments :)

api/src/v2/Registration/Assembler/SiriusToLayDeputyshipDtoAssembler.php
```diff
@@ -87,6 +87,10 @@ private function determineReportTypeIsSupported(?string $reportType)
...
+        if ($supported) {
```
coderste, line 90: Maybe it's the Go side of me but I believe writing something like:
```
if (!$supported) {
    throw new InvalidArgumentException();
}

return $reportType
```

Feels cleaner (to me anyways), as from a Go perspective you're taught to deal with any errors you may potentially have first and then if you have none you just return what's expected. 

Although I know this is not Go I still feel like that way of thinking feels more natural to me. Check for an error first -> handle the error if there is one -> if no error return what we're expecting 

This way you also make the code a little cleaner as there's no `else` statement :)

acsauk (changes requested):
Nice job @Gugandeep - couple of small changes and then ready to go 💪

api/tests/Behat/bootstrap/v2/Registration/IngestTrait.php
Gugandeep: I saw the method for errors, and for whatever reason just assumed we weren't doing this for the alerts!

`AssertTrait::assertOnAlertMessage()` worked perfectly!

api/src/v2/Registration/Assembler/SiriusToLayDeputyshipDtoAssembler.php
```diff
@@ -87,6 +87,10 @@ private function determineReportTypeIsSupported(?string $reportType)
             default => false
         };
 
-        return $supported ? $reportType : '';
+        if (!$supported) {
```
coderste, line 90: Nice 😄

coderste (dismissed):

acsauk (approved):
:shipit:

PR #835, opened 2022-05-20
DDPB-4338: Start setting named deputy name based on org name when is an org deputy

acsauk:
## Purpose
With the move from casrec to sirius as our source of truth there has been a change in the domain in how professional and PA deputies are being ingested during our CSV upload. In Sirius, a deputy can be either a person or an organisation (e.g. law firms and local authorities). This detail was partially present in casrec but hidden behind the naming of the CSV columns - an organisation deputies firstname and lastname was usually a combination of `Deputy Officer` and the name of the organisation (e.g. `DEPUTYSHIP OFFICER LAW FIRM INC`). As the column format was the same for both people deputies and organisation deputies, digideps modelled named deputies as people that would have a first and last name.

Once we started to receive CSVs from Sirius where we had a combination of person and organisation deputies it meant that our logic for assigning named deputy names was broken for organisation deputies as in Sirius an organisation deputy does not have a first and last name - just an organisation name.

This change checks which type of deputy the row refers to and sets the name accordingly. We should update our domain models in the future to make this link explicit with the introduction of some type of flag or identifier to list a named deputy as an organisation and potentially add `NamedDeputy::OrganisationDeputyName`.

Fixes DDPB-4338.

## Approach
A small and temporary update to ensure we can reinstate deputy names for organisation deputies as over 24K rows had their deputy names wiped. 

## Checklist


Gugandeep (approved):

api/src/Entity/NamedDeputy.php
```diff
@@ -511,4 +511,14 @@ public function addressHasChanged(OrgDeputyshipDto $dto)
              $this->getAddress5() !== $dto->getDeputyAddress5() ||
              $this->getAddressPostcode() !== $dto->getDeputyPostcode();
     }
+
+    public function nameHasChanged(OrgDeputyshipDto $dto): bool
+    {
+        if ($dto->deputyIsAnOrganisation()) {
```
coderste, line 517: Although I know this is PHP I believe we can take some learnings from Go when writing `if/else` statements. In this scenario we're checking the happy path first and returning. If we do end up inside the `if` because we have an instant return we then know the code no longer executes any further in this function therefore making the `else` block redundant.

One of Go's other standards is you should always try and deal with the unhappy scenario first. In doing so I believe this function could be written as.

```
if (!$dto->deputyIsAnOrganisation()) {
    return $dto->getDeputyFirstname() !== $this->getFirstname() ||
            $dto->getDeputyLastname() !== $this->getLastname();
}

return return $dto->getOrganisationName() !== $this->getFirstname();
``` 

In my opinion I believe this just looks cleaner, and also we no longer have the extra block of code in the `else` statement as we deal with the unhappy path first and if we don't fall into it we just return the expected value.

PR #837, opened 2022-05-23
DDPB-4341 allow us to recover from disasters in automated fashion

jamesrwarren:
## Purpose

Restore from snapshots and point in time as well as from backup account.
Fixes DDPB-4341

## Approach
The snapshot method is fairly annoying and obtuse and works differently based on serverless and clusters. Given the size of the database I would be half tempted to just use pg_dumps but snapshots are more AWS-ey and I had already got far with the snapshots by the time I realised how annoying it was. Also I'm not sure how to easily do PITR with pg_dumps method.

Anyway there are some examples and explanations of how it works in the readme. It's quite slow because of all the renaming, deleting, restoring etc... but not horrific but still slow.

## Learning
Some of the defaults aren't good so learned we had to use more settings from the original or terraform tries to recreate DB!

## Checklist

### Frontend


disaster_recovery/restore/disaster_recovery.py
```diff
@@ -0,0 +1,621 @@
...
+        if self.db_cluster_identifier_source is not None:
+            self.get_cluster_info()
+            if self.EngineMode != 'serverless':
+                self.get_instance_info()
+            if self.db_cluster_identifier_source == self.db_cluster_identifier_target and self.DeletionProtection:
+                self.drop_protection()
+
+        self.apply_overrides()
+
+        if self.PointInTimeRecovery != datetime.strptime('1900-01-01 00:00:00', '%Y-%m-%d %H:%M:%S'):
```
acsauk, line 102: Assuming this is the format `self.PointInTimeRecovery` must be set to in order to restore to a point in time we could add a little more context here by extracting `'1900-01-01 00:00:00'` to a var with a name describing its intent (something like `requiredDateTimeFormat`?)

disaster_recovery/restore/disaster_recovery.py
```diff
@@ -0,0 +1,621 @@
...
+            self.db_cluster_identifier_target = f"{self.db_cluster_identifier_source}-temp"
+
+        self.restore_cluster_ss()
+        if self.EngineMode != 'serverless':
+            self.create_db_instances()
+
+        if self.same_target:
+            self.overwrite_existing_cluster()
+
+    def restore_cluster_ss(self):
```
acsauk, line 180: What does `ss` stand for? Could we get it expanded to the full words?

disaster_recovery/restore/disaster_recovery.py
```diff
@@ -0,0 +1,621 @@
...
+    def restore_to_point_in_time(self):
+        if self.same_target:
+            self.db_cluster_identifier_target = f"{self.db_cluster_identifier_source}-temp"
+
+        self.restore_cluster_pitr()
+
+        if self.same_target:
+            self.overwrite_existing_cluster()
+
+    def restore_cluster_pitr(self):
```
acsauk, line 224: Same as `ss` - could we get this expanded to `point_in_time_xxxx` (not sure what the r stands for 😅 )

docs/DISASTER_RECOVERY.md
```diff
@@ -23,11 +23,65 @@ All of our measures to restore the service fall within the 48 hours specified by
...
+Title: Restore to a different cluster
+Example disaster: You have no idea when something bad happened to the data and want to go back in time on 
+another instance and see if the data was ok at a point in time without affecting your main DB operation.
+Remediation: Restore from an existing DB to a point in time into another cluster.
+```
+aws-vault exec identity --duration=2h -- \
+docker-compose -f docker-compose.commands.yml run -rm dr-restore \
+python3 disaster_recovery.py \
+--cluster_from api-ddpb9999 --cluster_to api-new_test_cluster --pitr '2022-01-01 09:10:00' 
+```
```
acsauk, line 84: Loving these docs 😍

disaster_recovery/restore/disaster_recovery.py
jamesrwarren: Good call

disaster_recovery/restore/disaster_recovery.py
jamesrwarren: Yeah not sure why i did that. Very naughty.

disaster_recovery/restore/disaster_recovery.py
jamesrwarren: :)

acsauk (approved):
:shipit:

PR #841, opened 2022-05-25
DDPB 3914 : Logging around admin managers

MiaGordon91:
## Purpose
Audit logging added for the creation and deletion of admin manager roles so we have sight on who is accessing our data and using our system.

Fixes DDPB-3914

## Approach
- AdminManagerCreatedEvent and AdminManagerDeletedEvent created
- Both events covered in the AdminUserLifeCycleSubscriber 
- AuditEvents class updated to cover both events 

## Checklist
- [x ] I have performed a self-review of my own code
- [x ] I have added tests to prove my work


api/src/Service/DateTimeProvider.php
```diff
@@ -2,10 +2,11 @@
 
 namespace App\Service;
 
-class DateTimeProvider
+class DateTimeProvider extends \App\Service\Time\DateTimeProvider
```
acsauk, line 5: I don't think we need this here - we have two classes named `DateTimeProvider` in API and Client so trying to extend one from the other will break our code as the classes aren't available to each other. Was there a reason you needed to do this?

client/src/Controller/Admin/IndexController.php
```diff
@@ -518,4 +525,18 @@ private function dispatchCSVUploadEvent()
 
         $this->eventDispatcher->dispatch($csvUploadedEvent, CSVUploadedEvent::NAME);
     }
+
+    private function dispatchAdminManagerDeletedEvent(User $userToDelete)
+    {
+        $trigger = AuditEvents::TRIGGER_ADMIN_MANAGER_MANUALLY_DELETED;
+        $currentUser = $this->tokenStorage->getToken()->getUser();
+
+        $adminManagerDeletedEvent = new adminManagerDeletedEvent(
```
acsauk, line 534: Technically I think PHP will let this work but the class name should be capitalised:

`AdminManagerDeletedEvent`

client/src/Service/Audit/AuditEvents.php
```diff
@@ -286,6 +290,50 @@ public function selfRegistrationFailed(array $failureData): array
...
+     * @param User   $currentUser         , the logged in user
+     *
+     * @throws \Exception
+     */
+    public function adminManagerCreated(string $trigger, User $currentUser, User $createdAdminManager): array
+    {
+        $event = [
+            'trigger' => $trigger,
+            'logged_in_user_first_name' => $currentUser->getFirstname(),
+            'logged_in_last_name' => $currentUser->getLastname(),
```
acsauk, line 305: It may be useful to include the logged-in users' email here too in case we need to contact or better identify them after looking at the logs.

client/src/Service/Audit/AuditEvents.php
```diff
@@ -286,6 +290,50 @@ public function selfRegistrationFailed(array $failureData): array
...
+     * @param User   $currentUser         , the logged in user
+     *
+     * @throws \Exception
+     */
+    public function adminManagerDeleted(string $trigger, User $currentUser, User $deletedAdminManager): array
+    {
+        $event = [
+            'trigger' => $trigger,
+            'logged_in_user_first_name' => $currentUser->getFirstname(),
+            'logged_in_last_name' => $currentUser->getLastname(),
```
acsauk, line 327: Same as above, including logged-in users' email would be useful.

client/src/Controller/Admin/IndexController.php
```diff
@@ -276,6 +279,10 @@ public function deleteConfirmAction($id)
 
         $this->denyAccessUnlessGranted(UserVoter::DELETE_USER, $userToDelete, 'Unable to delete this user');
 
+        if (User::ROLE_ADMIN_MANAGER === $userToDelete->getRoleName()) {
+            $this->dispatchAdminManagerDeletedEvent($userToDelete);
+        }
+
```
acsauk, line 285: Is this definitely where the user gets deleted? I can see in the `deleteAction` below this that we're calling `$this->userApi->delete()`. We would want to be sure that the user was definitely deleted so would need to wait until after this call has been made before logging an audit event.

acsauk (changes requested):
Looking good @MiaGordon91 🎉  a couple of points to change and check - let me know if you need any more details on it.

api/src/Service/DateTimeProvider.php
MiaGordon91: Sorry I think I did this in error,  I'll revert this back.

acsauk (approved):
:shipit: LGTM!

PR #847, opened 2022-05-30
DDPB:4333 - Add banner to show super admin users which version of the app they are using

MiaGordon91:
## Purpose
Add a visible banner that shows the version of the app to super admins only. This banner applies to branch, preproduction, integration and training. 

Fixes DDPB-4333

## Approach
- Created a new global variable under the globals key in twig.yml file and passed this into the admin_moj_template twig file.
- Created behat tests to check banner is visible to super admins only and tested with a Lay Deputy fixture also. 

## Checklist
- [x ] I have added tests to prove my work


api/tests/Behat/bootstrap/v2/AppNotification/AppNotificationTrait.php
```diff
@@ -92,4 +92,20 @@ public function iShouldSeeAValidationError()
...
+    {
+        $this->assertPageContainsText(sprintf('You are now logged into the %s environment', $hostedEnvironment));
+    }
+
+    /**
+     * @Then I should not see the banner confirming the version of the app I am using
+     */
+    public function iShouldNotSeeTheBannerConfirmingTheVersionOfTheAppIAmUsing()
+    {
+        $this->assertElementNotOnPage('govuk-notification-banner__heading');
```
acsauk, line 109: Its good to check for the banner itself but to make this less brittle (e.g. easy to break) we should probably assert the text doesn't appear anywhere on the page either - mainly because we could change the banner we use for this in the future and we wouldn't then never know if this test is actually passing.

api/tests/Behat/features-v2/app-notification/super-admin-banner-visibility.feature
```diff
@@ -0,0 +1,21 @@
...
+    @super-admin
+    Scenario: A super admin has visibility of the banner notification in the admin app
+        Given a super admin user accesses the admin app
+        Then I should see a banner confirming the 'admin' version of the app I am using
+
+    @admin-manager
+    Scenario: An admin manager does not have visibility of the banner notification
+        Given an admin manager user accesses the admin app
+        Then I should not see the banner confirming the version of the app I am using
+
```
acsauk, line 16: Could we add an extra test case for regular admin users too? You can use the `@admin` tag and `Given an admin user accesses the admin app`

api/tests/Behat/features-v2/app-notification/super-admin-banner-visibility.feature
```diff
@@ -0,0 +1,21 @@
+@v2 @v2_admin @app-notification
+Feature: Restricting visibility of banner notification to super admin only
+    To ensure that I know what version of the app I am using
+    As a super admin user
+    I want to see a banner notification confirming the hosted environment I am logged in to
+
+    @super-admin
+    Scenario: A super admin has visibility of the banner notification in the admin app
+        Given a super admin user accesses the admin app
+        Then I should see a banner confirming the 'admin' version of the app I am using
```
acsauk, line 10: So this is going to be showing the name of the hosted environment (e.g. ddpb4333, integration, preproduction etc) rather than which view of the app you're looking at (e.g. frontend or admin). I think the best way to update this test will be to match on the text you expect but allow the name of the environment to be any string. Heres a good SO post on explaining the type of regex you could use for this https://stackoverflow.com/questions/12677178/regular-expression-with-wildcards-to-match-any-character

acsauk (changes requested):
A couple of test updates but looking good! Given the changes you've made, there's no need for any terraform updates as we've only added a twig variable rather than changed the name of an environment variable. The test that's failed in the pipeline is a valid failure - you can see in the app on your branch env the message doesn't match your expectation:

Screenshot 2022-05-31 at 10 22 33

api/tests/Behat/features-v2/app-notification/super-admin-banner-visibility.feature
```diff
@@ -1,4 +1,4 @@
-@v2 @v2_admin @app-notification
+@v2 @v2_admin @app-notification @mia
```
acsauk, line 1: Missed one 😁

acsauk (changes requested):
One behat tag to remove then good to go 👍

acsauk (approved):
:shipit:

PR #848, opened 2022-05-31
DDPB-4352 bump terraform version and provider versions (specifically AWS)

jamesrwarren:
## Purpose
We are running a bit of an old version of both terraform and the AWS provider. Update them.

Fixes DDPB-4352

## Approach

So as of 4.9 we're not forced to make all the deprecation changes to the s3 providers so I have skipped neatly to latest version. I have however done the changes on the s3 access log bucket. I checked and it doesn't seem to try and destroy the bucket and the applies work fine against an env where bucket is in it's current format.

However, for the purpose of risk mitigation I was going to put this all the way through to prod and then do a small follow up (call it 4352b or something) for the deprecation warnings on our main s3 bucket and follow then plan through very carefully to prod.

Updated a few other deprecation warnings. Also including the lock file as it's best practice:

`Terraform automatically creates or updates the dependency lock file each time you run the terraform init command. You should include this file in your version control repository so that you can discuss potential changes to your external dependencies via code review, just as you would discuss potential changes to your configuration itself.`

## Learning
NA

## Checklist

### Frontend

### DIFF
--- a/environment/.terraform.lock.hcl
+++ b/environment/.terraform.lock.hcl
@@ -1,0 +1,59 @@
+# This file is maintained automatically by "terraform init".
+# Manual edits may be lost in future updates.
+
+provider "registry.terraform.io/hashicorp/archive" {
+  version = "2.2.0"
+  hashes = [
+    "h1:2K5LQkuWRS2YN1/YoNaHn9MAzjuTX8Gaqy6i8Mbfv8Y=",
+    "zh:06bd875932288f235c16e2237142b493c2c2b6aba0e82e8c85068332a8d2a29e",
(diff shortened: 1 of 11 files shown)


environment/.terraform.lock.hcl
```diff
@@ -0,0 +1,59 @@
+# This file is maintained automatically by "terraform init".
+# Manual edits may be lost in future updates.
+
+provider "registry.terraform.io/hashicorp/archive" {
```
acsauk, line 4: This is cool to see - easier to track changes in PRs.

acsauk (approved):
:shipit:

PR #857, opened 2022-06-15
amend sync timings

jamesrwarren:
## Purpose
Costing a whole lot in config because of the creation of multiple resources every time the cloudwatch event triggers

## Approach
For now shall we just set prod to be every 5 mins for both jobs and everything else to be once every 24 hours. Also do we need checklists and reports to be separate commands? Could half the number of calls if we just dumped all the code in one command called sync to sirius or something.

## Learning
NA

## Checklist

### Frontend


acsauk (approved):
:shipit:

PR #858, opened 2022-06-17
DDPB-4318: Implement JWT generation and lock down one route

acsauk:
## Purpose
Given that we will be integrating with different services within OPG going forward (and they will be integrating with us) we need a way to authenticate access to our app that doesn't rely on user login in every instance.

I've gone for JWT here to open up the possibility of providing tokens to other services from a secure location that they can use to connect to our app. Specifically this will be useful for the CSV automation process that will be POSTing CSVs to our API and won't have a user account in digideps to authenticate with.

Fixes DDPB-4318

## Approach
See the [miro flow diagram](https://miro.com/app/board/uXjVOcGDxvU=/?moveToWidget=3458764524209948402&cot=14) for the current flow. 

Right now, JWTs will only be returned for users with a SUPER_ADMIN role and are being used to lock down a single reports endpoint as a POC. The current code is saving the JWT in session to be accessed on future requests but I'm moving more towards saving this in a cookie after reading around on some security implications of compromised session attacks. However, as users aren't the focus of the use case for this (yet) and users still need to authenticate using username and password at all times the risk of any issues arising is low and we can revisit once we roll this out to all users.

Once this has been rolled out to all users we can replace the secret string approach we currently have.

I'm also thinking through how having a completely separate authentication service could work as some placing of the endpoints here in the client and API doesn't feel quite right.

## Learning
Endless posts and videos describe how JWTs work helped with this PR. Here are a few:

* [Auth0 JWT Handbook](https://auth0.com/resources/ebooks/jwt-handbook)
* [Angular JWT Guide](https://blog.angular-university.io/angular-jwt/)
* [PHP JWT Talk](https://youtu.be/p2tItlr_3QI?t=904)

## Checklist


api/src/Service/JWT/JWTService.php
```diff
@@ -0,0 +1,81 @@
...
+    }
+
+    public function createNewJWT(User $user)
+    {
+        $privateKey = base64_decode($this->secretManager->getSecret(SecretManagerService::PRIVATE_JWT_KEY_BASE64_SECRET_NAME));
+        $publicKey = base64_decode($this->secretManager->getSecret(SecretManagerService::PUBLIC_JWT_KEY_BASE64_SECRET_NAME));
+
+        $kid = openssl_digest($publicKey, 'sha256');
+        $payload = [
+            'aud' => 'registration_service',
```
mattmachell, line 32: I've started a very early draft of an OPG JWT standard over in this [Google doc](https://docs.google.com/document/d/1AccZBj-YZdhRk3AG_TlV9Bdyf17HaGW7qmH01_igfr8/edit?usp=sharing). In particular I think the bits about using urns for subject and audiences might be useful to shape this code.

acsauk: Thanks Matt - that's useful to see how it would work in practice and we can implement that here. I guess we'd need to think through how we're going to standardise the urns and make available some way to access a list of existing ones to use in services rather than hard coding.

One thing I've been thinking through is the use case I'm implementing this change for; services without a concept of a user that need to interact with digideps. My initial thought was to have some central token store where you register the service that is going to have an integration and then receive some type of API key (or the JWT itself) that you can pull into or set in the service. A quick iteration of this would be to manually or automatically generate a JWT and store in secret manager to be pulled into a service.

Can you see any issues with this approach?

mattmachell: I seem to remember this is very similar to a pattern that @andrewpearce-digital had started on in the metrics service. i.e. a job that periodically regenerates a service JWT and pushes to secrets manager.

Although to be honest if we can get all services to generate a JWT the need for service level accounts is reduced as every end user could be issued a claim appropriate for  interacting with multiple services.

api/src/Entity/User.php
```diff
@@ -643,7 +653,7 @@ public function getUsername()
 
     public function getSalt()
     {
-        //return $this->salt;
+        // return $this->salt;
```
mattmachell, line 656: Why is this commented out?

api/tests/Unit/Service/Auth/AuthServiceTest.php
```diff
@@ -179,6 +193,39 @@ public function testisSecretValidForRole($clientSecret, $role, $expectedResult)
         $this->assertEquals($expectedResult, $this->authService->isSecretValidForRole($role, $request));
     }
 
+    /** @test */
+    public function jWTIsValid()
+    {
+        $this->JWTService->shouldReceive('verify')->with('not-a.real-jwt')->andReturn(true);
```
mattmachell, line 199: Might bee worth adding variations for the common JWT highjack attacks here. Setting `alg=none `on an otherwise valid token and so on. There's a good summary of them over here: https://www.invicti.com/blog/web-security/json-web-token-jwt-attacks-vulnerabilities/ and here: https://infosecwriteups.com/attacks-on-json-web-token-jwt-278a49a1ad2e

api/src/Entity/Traits/AddressTrait.php
```diff
@@ -167,30 +167,30 @@ public function setAddressCountry($addressCountry)
         return $this;
```
coderste, line 167: Should we type-hint the rest of the returns for this file?

api/src/Service/Auth/AuthService.php
```diff
@@ -90,12 +47,9 @@ public function isSecretValid(Request $request)
     }
 
     /**
-     * @param string $email
-     * @param string $pass
-     *
      * @return User|bool|null or null if the user it not found or password is wrong
      */
-    public function getUserByEmailAndPassword($email, $pass)
+    public function getUserByEmailAndPassword($email, $pass): User|bool|null
```
coderste, line 52: We've removed the DocBlock `@param` for the types but then we don't type hit the actual parameters in the function should we do this?

api/src/Service/Auth/AuthService.php
```diff
@@ -123,12 +77,7 @@ public function getUserByEmailAndPassword($email, $pass)
         return null;
     }
 
-    /**
-     * @param string $token
-     *
-     * @return User|null
-     */
-    public function getUserByToken($token)
+    public function getUserByToken($token): User|null
```
coderste, line 80: Again should we type-hint the parameter?

coderste (approved):

api/src/Entity/User.php
acsauk: No idea - the change was made in 2015 and hasn't been touched since.

mattmachell: That's an odd one then. Worth spinning up a ticket to investigate.

api/tests/Unit/Service/Auth/AuthServiceTest.php
acsauk: Yeah, fair point - I'd taken the approach that the package we're using to generate JWTs makes it impossible to use a none algorithm but that doesn't mean a JWT coming in from another source won't be malicious.

I think this is going to mean me changing the JWT package we're using out as its physically impossible to override `alg` but I don't think this is a bad thing as [another package](https://github.com/lcobucci/jwt) I was looking at has a much friendlier APIwhich should make the code a bit more flexible in the future and easier to write tests for.

mattmachell: `lcobucci/jwt` is the same package we use over in Sirius, so good for consistency.

There's a particular balance with authentication mechanisms where I want to be sure I have tests that can verify common attack vectors can't be used against my app and if the abstraction doesn't allow that you're at the mercy of the 3rd party's tests.

api/src/Service/JWT/JWTService.php
```diff
@@ -0,0 +1,135 @@
...
+        $config = Configuration::forAsymmetricSigner(
+            new Signer\Rsa\Sha256(),
+            Signer\Key\InMemory::plainText($privateKey),
+            Signer\Key\InMemory::plainText($publicKey),
+        );
+
+        $config->setValidationConstraints(
+            new IssuedBy(self::generateIss()),
+            new PermittedFor(self::generateAud('registration_service')),
+            new SignedWith(new Signer\Rsa\Sha256(), Signer\Key\InMemory::plainText($publicKey))
```
acsauk, line 130: @mattmachell this guards against `none` algorithms now

api/tests/Unit/JWTServiceTest.php
```diff
@@ -0,0 +1,177 @@
...
+
+        self::assertTrue($this->sut->verify($jwt));
+    }
+
+    /** @test */
+    public function verifyWithMissingAlgorithmFails()
+    {
+        $jwt = $this->createUnsignedJWTString($this->publicKeyPem, 'https://example.org');
+
+        self::assertFalse($this->sut->verify($jwt));
```
acsauk, line 63: And covered with some tests here too @mattmachell

client/Dockerfile
```diff
@@ -46,8 +46,8 @@ ENV TIMEOUT=60
 ENV PHP_EXT_DIR=/usr/local/lib/php/extensions/no-debug-non-zts-20200930/
 
 # Install core PHP extensions
-RUN apk add --no-cache openssl nginx nginx-mod-http-headers-more su-exec postgresql-client libzip-dev unzip php8-pecl-igbinary php8-pecl-redis php8-pecl-imagick
-RUN docker-php-ext-install pcntl zip opcache
+RUN apk add --no-cache openssl nginx nginx-mod-http-headers-more su-exec postgresql-client libzip-dev unzip php8-pecl-igbinary php8-pecl-redis php8-pecl-imagick gmp gmp-dev
+RUN docker-php-ext-install pcntl zip opcache gmp
```
acsauk, line 50: This was required to ensure generating the public keys from JWKs happens quickly.

api/src/Service/Auth/AuthService.php
acsauk: I've reverted this change and will leave fixing these up to the PHPStan tickets. Cant face having to debug any code this may break after type hinting 🙄

api/src/Service/Auth/AuthService.php
acsauk: As above, leaving this change to the PHPStan tickets.

api/src/Entity/Traits/AddressTrait.php
acsauk: Going to defer this to the PHPStan ticket to tackle all at once.

coderste (approved):

PR #860, opened 2022-06-21
DDPB-4373 change the teardown schedule on ephemeral envs

jamesrwarren:
## Purpose
Cut down on costs by being a bit more ruthless with the env teardown

Fixes DDPB-4373

## Approach
Only thing of note here is I put the workspace protection before the plan and all that malarkey. The reason for this is that now that we're running hourly I can imagine a situation where we try and delete at the same time as rerunning a workflow and doing an apply. This should avoid this from happening as the workspace will be protected before it runs the deletes...

However thinking about it, the destroy could happen first and then the job will fail on plan or apply. However there will be a state lock so this shouldn't cause issues and job can be rerun. If it does end up causing issues then we will need to chuck in a waiter script that looks for the deletion runs and checks if they are running but it seems this would be pretty edge case imo.

## Learning
NA

## Checklist

### Frontend

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -24,21 +24,20 @@ workflows:
           tf_workspace: development
           tf_command: plan
 
+      - workspace-protection:
+          name: protect branch workspace
+          requires: [cancel redundant builds]
+          filters: { branches: { ignore: [main] } }
+
       - terraform-command:
           name: plan branch env
-          requires: [cancel redundant builds]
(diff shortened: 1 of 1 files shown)


acsauk (approved):
Nice nice - will keep an eye on any weirdness with jobs but as long as there's no permanently locked states we should be fine to just re-run 🙌

coderste (approved):

PR #864, opened 2022-06-23
[cert-doc-update] update README docs for certificates

coderste:
After talking @MiaGordon91 through the process of getting a trust SSL certificate for our local sites I noticed I missed a few steps out in the README so Chrome knows to trust the new certificates. I have now added the extra steps into the README so everyone else can follow it.

I've only added steps for the macOS system. I don't have a Windows system on hand to write a README up on how I would get a Windows system to trust the certificates 

Updated the `container_name` for the api/frontend service as the `make db-terminal` command I used I entered the `container_name` which was showing on my environment but found out on other environments it can be different so defining it here allows the command to work for everyone


mattmachell:
Don't know if it's useful, but if you need a temp workaround before installing the certs if you type `thisisunsafe` on the unsafe connection page it will let you past. Or at least it used to.

coderste:
> Don't know if it's useful, but if you need a temp workaround before installing the certs if you type `thisisunsafe` on the unsafe connection page it will let you past. Or at least it used to.

PR #865, opened 2022-06-23
DDPB 4275 - Track registration form errors via google analytics

Gugandeep:
## Purpose
Currently, when a deputy attempts to register for the service with invalid details, they receive a generic error.
Therefore, we are unable to track which fields or pieces of information deputies struggle to get right.

The basis of this ticket is to track which fields deputies are getting wrong. However, to do this we also need to provide specific errors.

This ticket is to add specific errors to the registration form, and enable tracking of these errors through Google Analytics.

Fixes DDPB-4275

## Approach
Initial work was to break down the matching process to be done into separate stages based on the matching criteria provided by the deputy. These matching criteria is as follows:

Case Number
Client Lastname
Deputy Lastname
Deputy Postcode

By breaking the process into stages, we will be able to determine which criteria failed to be matched on, and can specify the error to the deputy.

The next step was to then include functionality in our javascript to be able to read form errors, and then send these as events to Google Analytics. One beneficial side effect to doing this is that we now have form error tracking enabled for all of our forms.

## Learning

## Checklist

### Frontend


Makefile
```diff
@@ -89,6 +89,9 @@ api-unit-tests: reset-fixtures ##@unit-tests Run the api unit tests
 behat-tests: up-app-integration-tests reset-fixtures ##@behat Run the whole behat test suite
 	docker-compose -f docker-compose.yml -f docker-compose.dev.yml run --rm test sh ./tests/Behat/run-tests.sh
 
+behat-tests-tag: up-app-integration-tests reset-fixtures ##@behat Run behat tests with specified tag e.g. make behat-tests-tag tag=<tag> (Do not include the @)
+	docker-compose -f docker-compose.yml -f docker-compose.dev.yml run --rm test sh ./tests/Behat/run-tests.sh --profile v2-tests-goutte --tags @$(tag)
+
```
acsauk, line 94: 🙌

api/src/Service/PreRegistrationVerificationService.php
```diff
@@ -97,27 +80,116 @@ public function isMultiDeputyCase(string $caseNumber): bool
...
+            if (mb_strtolower($match->getDeputySurname()) == mb_strtolower($detailsToMatchOn['deputyLastname'])) {
+                $deputyLastnameMatches[] = $match;
             }
         }
 
-        if ($preRegistrationWithPostcodeCount < count($preRegistrationMatches)) {
-            $filteredResults = $preRegistrationMatches;
+        if (0 === count($deputyLastnameMatches)) {
+            $matchingErrors['deputy_lastname'] = true;
+            $deputyLastnameMatches = $clientLastnameMatches;
```
acsauk, line 133: I may be misunderstanding the intent here, but do we need to keep updating and creating these match variables in each block or could you just have s single variable called something like `$matches`?

api/tests/Unit/Controller/SelfRegisterControllerTest.php
```diff
@@ -266,6 +265,169 @@ public function throwErrorForValidCaseNumberButDetailsNotMatching()
...
+            'DeputyAddress1' => 'Victoria Road',
+            'DeputyPostcode' => 'SW1',
+            'ReportType' => 'OPG102',
+            'MadeDate' => '2010-03-30',
+            'OrderType' => 'pfa',
+            'NDR' => 'yes',
+        ], $now);
+
+        $this->fixtures()->persist($casRec);
+        $this->fixtures()->flush();
```
acsauk, line 299: There's some repetition in this, the previous and next test in regards to setting up data that could be refactored. Giving it a descriptive function name will also make the intent clear on what the data is.

api/tests/Unit/Service/PreRegistrationVerificationServiceTest.php
```diff
@@ -26,46 +26,54 @@ public function setUp(): void
         $crLayHasPC = m::mock('\App\Entity\PreRegistration')
```
acsauk, line 26: I remember updating this test when we moved to pre-registration wording, it's horrendous!

api/src/Service/PreRegistrationVerificationService.php
Gugandeep: Yeah this part is slightly confusing, and unfortunately I couldn't think of a better way of writing this code.

Since we want to break down the matching process into smaller steps so we can report errors, we need to sure after each step we are getting some feedback. We also need to ensure that when there is a failure to match on a specific step, we carry on the next step, so we can record all the errors.

So put simply, at each step we are checking if a criteria matches against our array of PreRegistration records. If there are any matches, we put this into a new array.

If there isn't anything in the array, we know there wasn't any matches, and we record the error.
When there isn't a match, the next step still needs to run, but we can't pass it the empty array, so we have to pass it the last set of successful matches.

So that's why we are keeping a record of all the past matches.

api/tests/Unit/Controller/SelfRegisterControllerTest.php
Gugandeep: That's a good shout. Will look into refactoring this.

api/src/Service/PreRegistrationVerificationService.php
acsauk: Ah OK cool - lets go with this then.

PR #866, opened 2022-06-27
DDPB-4365: Get list of caseworker accounts not used in 1 year and 1 month

MiaGordon91:
## Purpose
A downloadable report was needed to collate old admin user accounts that have not been used in 13 months. The report will be shared with OPG to confirm if accounts can be deleted so that we remain secure against non-required access and comply with GDPR standards.  

Fixes DDPB-4365

## Approach
- Function has been added to UserRepository and unit tested
- StatsControllers have been updated and tested 
- CsvGenerator class has been created
- Download link has been added to twig file
- Tested in 'super-admin-only' feature 

## Checklist


api/src/Controller/StatsController.php
```diff
@@ -92,6 +92,25 @@ public function getAdminUserAccountReportData(Request $request, RestFormatter $f
         ];
     }
 
+    /**
+     * @Route("stats/admins/old_report_data", methods={"GET"})
+     * @Security("is_granted('ROLE_SUPER_ADMIN')")
+     */
+    public function getOldAdminUserReportData(Request $request, Restformatter $formatter): array
+    {
+
```
acsauk, line 101: We can drop the new line here

api/src/Controller/StatsController.php
```diff
@@ -92,6 +92,25 @@ public function getAdminUserAccountReportData(Request $request, RestFormatter $f
...
+        $adminUserAccountsNotUsedWithin13Months = $this->userRepository->getAllAdminUserAccountsNotUsedWithin('-13 months');
+
+        return [
+            'AdminUserAccountsNotUsedWithin13Months' => $adminUserAccountsNotUsedWithin13Months
+        ];
+    }
+    
+    
+
```
acsauk, line 113: And these too

api/src/Entity/User.php
```diff
@@ -74,7 +74,7 @@ class User implements UserInterface
     /**
      * @var int
      * @JMS\Type("integer")
-     * @JMS\Groups({"user", "report-submitted-by", "user-id", "user-list"})
+     * @JMS\Groups({"user", "report-submitted-by", "user-id", "user-list", "admin-account-reports"})
```
acsauk, line 77: Seeing as all of the properties have `user` and `user-list` as JMS groups that you need you could try using this when calling out the the API rather than adding your own as the group lists can start to bloat quite quickly when adding them for individual functions. It's worth checking if this impacts performance with prod-like data.

api/src/Repository/UserRepository.php
```diff
@@ -290,4 +290,20 @@ public function getAllAdminAccountsUsedWithin(string $timeframe)
 
         return $query->getResult();
     }
+
+
+    public function getAllAdminUserAccountsNotUsedWithin(string $timeframe)
+
```
acsauk, line 296: We can drop this newline

api/tests/Behat/bootstrap/v2/Analytics/AnalyticsTrait.php
```diff
@@ -292,4 +293,6 @@ public function shouldHaveNoIssuesDownloadingFile()
             throw new BehatException(sprintf('%s, returned a %s response', $this->currentLinkText, strval($responseStatus)));
         }
     }
+
+
```
acsauk, line 297: We can drop thses too

api/tests/Behat/bootstrap/v2/Common/IVisitAdminTrait.php
```diff
@@ -176,6 +176,15 @@ public function iVisitAdminStatsReportsPage()
...
 
+    /**
+     * @When I visit the admin users not logged in within last 13 months report page
+     */
+    public function iVisitOldAdminUsersReportsPage()
+    {
+        $this->visitAdminPath($this->getOldAdminUsersReportUrl());
+    }
+
+
```
acsauk, line 187: And we can drop this one

client/src/Controller/Admin/StatsController.php
```diff
@@ -338,4 +342,30 @@ public function downloadAssetsTotalValues()
 
         return $response;
     }
+
```
acsauk, line 345: We can drop this new line

client/src/Controller/Admin/StatsController.php
```diff
@@ -338,4 +342,30 @@ public function downloadAssetsTotalValues()
...
+        $response->headers->set('Content-Type', 'text/csv; charset=utf-8');
+        $disposition = $response->headers->makeDisposition(
+            ResponseHeaderBag::DISPOSITION_ATTACHMENT,
+            'inactiveAdminUsers.csv'
+        );
+
+        $response->headers->set('Content-Disposition', $disposition);
+
+        return $response;
+
```
acsauk, line 369: And this one 😁

client/src/Service/Csv/OldAdminUserCsvGenerator.php
```diff
@@ -0,0 +1,44 @@
+<?php
+
```
acsauk, line 2: Could we add:

```suggestion
declare(strict_types=1);

```

This will ensure we're type safe when using this class.

composer.json
```diff
@@ -0,0 +1,5 @@
+{
+    "require-dev": {
+        "jangregor/phpstan-prophecy": "^1.0"
+    }
+}
```
acsauk, line 5: It looks like `composer require` was run at the root of the project which created this file. If we need this package it should be installed in either client or API so we can drop this file.

composer.lock
```diff
@@ -0,0 +1,143 @@
+{
```
acsauk, line 1: Same as `composer.json` - we can drop this file.

acsauk (changes requested):
Nice job 👏 

Few changes and then this is ready to go.

acsauk (approved):
:shipit:

PR #867, opened 2022-06-28
DDPB-4409: Ensure we lowercase casenumbers when matching during lay CSV upload

acsauk:
## Purpose
After we started receiving CSVs from Sirius rather than casrec, case numbers with letters were appearing with uppercase letters whereas previously they were lowercased. As we weren't lowercasing when searching on case numbers in the lookup during CSV upload any case with letters was not updated with report type changes. This change ensures we lowercase both the property on Report and the case number in the CSV.

Fixes DDPB-4409.

## Checklist

### DIFF
--- a/api/tests/Unit/Entity/Repository/ReportRepositoryTest.php
+++ b/api/tests/Unit/Entity/Repository/ReportRepositoryTest.php
@@ -26,13 +26,13 @@ class ReportRepositoryTest extends ApiBaseTestCase
     /** @var ReportRepository */
     private $sut;
 
-    /** @var Report | MockInterface */
+    /** @var Report|MockInterface */
     private $mockReport;
 
-    /** @var ManagerRegistry | MockInterface */
+    /** @var ManagerRegistry|MockInterface */
     private $mockManagerRegistry;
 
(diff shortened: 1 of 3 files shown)


api/tests/Unit/Entity/Repository/ReportRepositoryTest.php
```diff
@@ -132,6 +132,23 @@ public function fetchQueuedChecklists()
...
+
+        $existingReport = $this->buildReport($client);
+
+        $this->entityManager->flush();
+        $this->entityManager->refresh($existingReport);
+        $this->entityManager->refresh($client);
+        $this->entityManager->refresh($client->getUsers()[0]);
+
+        $result = $this->repository->findAllActiveReportsByCaseNumbersAndRole(['4932965T'], $client->getUsers()[0]->getRoleName());
+        self::assertEquals($existingReport[0], $result);
```
Gugandeep, line 149: Think this should be:

`self::assertEquals($existingReport, $result[0]);`

Gugandeep (changes requested):

Gugandeep (dismissed):

Gugandeep (approved):

PR #868, opened 2022-06-29
[DDPB-4353] Add created at and updated at to entities to provide context

coderste:
## Purpose
Provide more context on the lifecycle of an entity to help with debugging and data issues

Fixes DDPB-4353

## Approach
Created a `Timestamps trait` class which implements the `createdAt` and `updatedAt` functions. Then we use the `@ORM\HasLifecycleCallbacks` annotation from Symfony which for each class that `use Timestamps` will call to the `createdAt` function when an entity is about to persist for the first time or the `updatedAt` function when an entity has changed and `flush` is called 

## Checklist


api/src/Entity/Traits/Timestamps.php
```diff
@@ -0,0 +1,60 @@
+<?php
+
+namespace App\Entity\Traits;
+
+use DateTime;
+use DateTimeInterface;
+use Doctrine\ORM\Mapping as ORM;
+
+trait Timestamps
```
mattmachell, line 9: I wonder if `Timestamps` is descriptive enough? Maybe something like `ChangeTrackingTimeStamps` so people know the intent?

coderste: Agreed although maybe just `TrackingTimestamps`? Change to me suggest we are changing something but for the majority of these entities we are adding a tracking timestamp rather than changing because they don't exist currently

mattmachell: The `CreateUpdateTimestamps` is much clearer.

PR #870, opened 2022-06-29
DDPB-4417b tfsec compliance

jamesrwarren:
This PR installs TFsec and runs it as part of the pipeline. 

We currently are using the default setup. It is scanning both environment and shared folders. I have got the results from environment folder from about 50+ issues down to 7 issues. These will need separate tickets. 

I haven't done anything on shared yet as it will be harder to prove that it doesn't break things without possibly breaking things for people. 

My method was to apply a set of changes, check the tests, commit the changes if tests passed. These all pass integration tests so there shouldn't be any issues releasing to live.

### DIFF
--- a/environment/front.tf
+++ b/environment/front.tf
@@ -72,25 +72,6 @@ data "aws_iam_policy_document" "front_query_ssm" {
   }
 }
 
-#resource "aws_iam_role_policy" "ecs_scheduled_tasks" {
-#  name   = "front-ecs-scheduled-task.${local.environment}"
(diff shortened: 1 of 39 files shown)


environment/front.tf
```diff
@@ -111,7 +85,17 @@ data "aws_iam_policy_document" "front_query_secretsmanager" {
     actions = [
       "secretsmanager:GetSecretValue"
     ]
-    resources = ["*"]
+    resources = [
```
jamesrwarren, line 88: I think this is still a bit relaxed but didn't want to spend too long working out exactly what the app code needs. Any advice welcomed :)

PR #872, opened 2022-07-06
DDPB-4434: Add secret prefixes for integration and preproduction envs

acsauk:
## Purpose
Standardising secret prefixes across environments as our current approach doesn't work in preproduction environments. I've gone with the same secret prefix so we don't need to duplicate secrets as there is currently no value in this approach.

Fixes DDPB-4434


environment/terraform.tfvars.json
```diff
@@ -127,7 +127,7 @@
       "ga_gds": "NotApplicable",
       "subdomain_enabled": true,
       "is_production": 0,
-      "secrets_prefix": "",
+      "secrets_prefix": "preproduction",
```
jamesrwarren, line 130: erm

PR #873, opened 2022-07-06
DDPB-4411 : Option to show password

MiaGordon91:
## Purpose
Deputy users now have the option to view their password to check if their data input is correct. This has also been applied to the password reset page.

Fixes DDPB-4411

## Approach
- The data-module attribute (taken from the password reveal component) has been added to the form password field in the login and password_reset twig files
- logic has been added to the _input.html twig file to add the data-module attribute to the form_widget if defined.
- MOJ Frontend has been updated to the latest version and is imported in main.js 

- Separate to this ticket, the ndr report wasn't submitting successfully on the frontend because it wasn't recognising its report type. To rectify this, the PreRegistrationFactory class now defaults 'ndr' report types to OPG102.

## Checklist

### Frontend
- [x] Any links or buttons added are screen reader friendly and contextually complete


MiaGordon91:
The password reveal option has been added to both fields on the password reset page. 
image

package-lock.json
```diff
@@ -1,3 +1,6 @@
 {
```
acsauk, line 1: This file needs to be dropped as looks like it was created in error

acsauk (changes requested):
Just the extra package-lock.json to drop and then this is good to go!

acsauk (approved):
:shipit: ready for a demo

acsauk (dismissed):
:shipit:

PR #889, opened 2022-07-17
Ddpb 4444b

jamesrwarren:
## Purpose

Making some fixes based on tfsec warnings that shouldn't break anything in the shared infra

Fixes DDPB-4444

## Approach

Went through and added commentary or added the fix to anything that didn't look too high risk.

## Learning

Tried it with changes to the endpoint security groups in and broke dev and spent quite a while fixing it so have taken those out for now!

## Checklist

### Frontend


Gugandeep (approved):
Looks good to me!

PR #895, opened 2022-07-21
DDPB-4396 synchronise from command

jamesrwarren:
## Purpose
Stop spin up of containers every few minutes to lower config costs

Fixes DDPB-4396

## Approach
Have gone for the approach of kicking off a lambda controlled by a cloudwatch event on a schedule. This in turn calls an end point in the app to send the documents/checklists across to sirius.

This is slightly complicated by us wanting to use a JWT token to authenticate the request. The way we have done this is to turn off the session based login security on the synchronise/* endpoints and to instead add a jwt verification step in. 

In the long term, it may be best to split JWT out to a separate service so that we can easily rotate the keys and tokens. As this is a low impact endpoint in terms of what it can do, I think the level of security on it is fine for now.

Another option long term is build the security into the base security module and have some option saying whether to use JWT or standard session based auth. 

## Learning
JWT was very confusing
Golang unit tests with interfaces... blew my mind a bit.. spent hours trying to work out how that works! 

## Checklist

### Frontend


api/src/Controller/JWTController.php
```diff
@@ -32,4 +37,12 @@ public function getPublicJwkKey(): JsonResponse
 
         return new JsonResponse($jwk);
     }
+
+    /**
+     * @Route("/jwt/authorise", methods={"GET"})
+     */
+    public function checkJWTValid(Request $request): bool
```
mattmachell, line 44: Not sure why the validity checking is part of this controller and requiring an HTTP request to invoke. Would expect it to be either a base class that the jwt-authorised endpoints extend, a trait or some middleware. Part of the reason to use jwt is you can validate a token without an external request.

lambda_functions/functions/synchronisation/go.mod
```diff
@@ -0,0 +1,20 @@
+module synchronisation
+
+go 1.17
```
coderste, line 3: This says 1.17 but in the Dockerfile we use 1.18 -- should we change to make sure they correlate?

api/src/Controller/JWTController.php
jamesrwarren: Yeah it was to avoid duplicating the code on the client side. So we have the verification on the API side but all of the document sync process on the client side. I wasn't sure how I would validate the JWT otherwise without rewriting all of the code to be API side or duplicating the validation to the client side (as we still want to be able to auth API side).

If we were to have client side API auth with JWT, how should that be processed then? What would be your thoughts on a separate JWT auth service eventually that sat outside the front and api containers? Is that a legit thing to do?

jamesrwarren: Couldn't quite bring myself to completely rewrite everything, not least as I was going for a 2 stage approach of releasing this out with the rest of the config in place, then toggling it to run for a while in prod to bed in before removing all the original task spin ups. I have duplicated a couple of the backend functions (to be removed in cleanup PR) so all the auth logic is done in backend only and we're only passing the token through (as that was your main concern). Hopefully that's ok :)

coderste (approved):

PR #896, opened 2022-07-25
Dual Xdebug setup

acsauk:
Our dev workflow is to open PHPStorm at the root of this repo. This gives us access to all the mono-repo sub-apps/directories, which is helpful for quickly navigating the codebase. However, when it comes to running Xdebug in PHPStorm this setup means we can only have one active Xdebug connection enabled at a time; client or API. This is down to how PHPStorm listens for incoming Xdebug connections - we expect there to be a string that corresponds to an incoming request that we set in PHPStorm setup and match against. When opening the repo at root level PHPStorm has no way of determining which directory the incoming request should be rooted to Xdebug is running in multiple containers.

A solution to this problem is to open two instances of PHPStorm; one for API and one for Client. With the additional changes in this PR this enables having Xdebug running on both containers so we can add breakpoints in files in both directories and debug app issues. 

The existing Xdebug setup should continue to work if you only require Xdebug enabled in one app at a time but this setup solves the annoying step of having the re-build docker containers to start to debug the other app when you need to. Just note there are slightly different setup steps for enabling the new approach but once setup will not have to be revisited.


jamesrwarren (approved):
Great instructions

coderste (commented):
LGTM 👍🏼

coderste (dismissed):

PR #904, opened 2022-08-15
Ddpb 4035b dependabot weekly updates

jamesrwarren:
## Purpose
Have weekly upgrades to our dependencies via dependabot.

Fixes DDPB-4035

## Approach
Initially, I chose weekly for a couple of reasons. Firstly, by putting the check on the Friday it's a kind of natural workaround for not merging them as soon as they come out. This seems like the most straight forward solution for now.

I started doing the dependabot workflows in github actions but then realised the check for approval wouldn't pass as it's expecting the circle one so quickly changed tack and just wrote it out in circle. 

~~I'm doing local integration checks. Not totally sure if that adds value. Maybe if something hasn't been covered by a unit test and a dependency then does unexpected things.. will give us more of a warm fuzzy feeling anyway.~~

^ After wasting a bit of time on this totally not worth it. Takes ages to build and had problems running it all locally in CI. Also we do integration tests as part of pipeline in to main so it would safely break on that if there were an issue before a release to prod anyway so overall much better to just not do this and content ourselves with the unit tests passing and then reverting if by some fluke there was coverage on an integration test that wasn't on a unit test and the dependency was a bad un.. Kept another branch with the code on just in case we want to come back to this..

I chose to limit to 3 version fixes per lock file per week to start with but we can modify this up or down. This doesn't include sec vulnerabilities that will still come in on top of these. 

## Learning
NA

## Checklist

### Frontend


docker-compose.yml
```diff
@@ -10,10 +10,6 @@ services:
       - admin
       - localstack
     restart: always
-    volumes:
```
jamesrwarren, line 13: changed this as we are trying to keep volume mounts separate

Gugandeep (approved):
Looks good to me.

PR #905, opened 2022-08-15
DDPB-4471 change log levels to be more generic

jamesrwarren:
## Purpose
Make the csv logger a bit more generic to avoid us having to have loads of difference (also add it to sync controller)

Fixes DDPB-4471

## Approach

So as part of a previous PR, devs added a new logger as we didn't want to change the log level across the app as it may have meant logging things we didn't want to in prod. Adding the extra logger gave us more fine grained control for individual classes. However, its name made it not very reusable. I think rather than having a new logger every time we want to capture notices, we can have verbose logger and use it for those time we want to capture more but not call it a warning.

## Learning
NA

## Checklist

### Frontend

### DIFF
--- a/api/config/packages/dev/monolog.yml
+++ b/api/config/packages/dev/monolog.yml
@@ -1,12 +1,12 @@
 monolog:
-    channels: ["csv"]
+    channels: ["verbose"]
     handlers:
         main:
             type: stream
             path: php://stderr
             level: warning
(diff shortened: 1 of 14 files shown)


Gugandeep (approved):
Looks good to me.

Good idea to have a more generic and suitable name for the logger/channels!

PR #947, opened 2022-09-27
DDPB-4501 remove golang vulnerability

jamesrwarren:
## Purpose
Fix for vulnerability picked up in ECR scan.

Fixes DDPB-4501

## Approach
Ironically the security vulnerability is in local-php-security-checker which does the same checks as dependabot is what was causing the golang security vulnerability warning from ECR. I tried to update it to latest version but that still had the issue so decided to remove completely as it does just check the composer lock files for know vulnerabilities the same as dependabot does so it's not really adding anything for us.

Updated php version to use latest alpine whilst I was there. 

## Learning
NA

## Checklist

### Frontend

### DIFF
--- a/api/Dockerfile
+++ b/api/Dockerfile
@@ -16,7 +16,7 @@ RUN composer run-script post-install-cmd --no-interaction
 
 RUN composer dump-autoload --optimize
 
-FROM php:8.0.19-fpm-alpine3.15
+FROM php:8.0.23-fpm-alpine3.16
 WORKDIR /var/www
 EXPOSE 80
 EXPOSE 443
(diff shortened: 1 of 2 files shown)


api/Dockerfile
```diff
@@ -30,8 +30,8 @@ RUN apk --no-cache add \
   nginx \
   nginx-mod-http-headers-more \
   su-exec \
-  php7-igbinary \
-  php7-pecl-redis
+  php8-pecl-igbinary \
```
charlesmmarshall, line 33: hey @jamesrwarren - is the php8 swap meant to be in this one too?

charlesmmarshall (approved):

PR #949, opened 2022-09-29
DDPB-4493 upgrade to postgres 13 local development and integration

jamesrwarren:
## Purpose
Upgrade to pg13 in local and dev environments

Fixes DDPB-4493

## Approach

- Use serverless (not real serverless) v2 to bring envs closer together
- Fix an api test. Did some digging and it looks like the previous way it was giving the Ids didn't seem logical when doing a bulk update.. we need to look out for this though as it's possible there's a place where this will be relevant.
- upgrade remote docker version on api tests as there was an error that was unreproducible locally

## Learning
NA

## Checklist

### Frontend


api/tests/Unit/ControllerReport/AssetControllerTest.php
```diff
@@ -66,13 +65,13 @@ public function testgetAssets()
 
         $this->assertCount(2, $data);
 
-        $this->assertEquals(self::$assetp1->getId(), $data[0]['id']);
-        $this->assertEquals('ha1', $data[0]['address']);
-        $this->assertEquals('property', $data[0]['type']);
+        $this->assertEquals(self::$asset1->getId(), $data[0]['id']);
+        $this->assertEquals('asset1', $data[0]['title']);
+        $this->assertEquals('other', $data[0]['type']);
```
Gugandeep, line 70: I think this might be to do with the default ordering of things in the new version.
Wonder if that changes the ordering in which things are displayed on the report.

Gugandeep (approved):

api/tests/Unit/ControllerReport/AssetControllerTest.php
jamesrwarren: So things I have tried:
- added multiple properties, checked ordering was same in integration as in this env.
- updated assets on multiple fields and checked that correct asset is updated and doesn't change ordering.
- added non property assets and checked they appear before properties.
Properties always seem to appear in reverse order that you enter them in which is kind of weird but is consistent with what is currently happening in integration.

PR #975, opened 2022-10-10
Remote docker changes, but with some additional config.

mattmachell:
Tom's work but with PHP unit verbose on, to see if it tells us anything.
.

### DIFF
--- a/client/scripts/client-unit-tests.sh
+++ b/client/scripts/client-unit-tests.sh
@@ -2,6 +2,6 @@
 # exit on error
 set -e
 
-php vendor/bin/phpunit -c tests/phpunit --coverage-php tests/phpunit/coverage/client-unit-tests.cov
+php vendor/bin/phpunit -c tests/phpunit --coverage-php tests/phpunit/coverage/client-unit-tests.cov --verbose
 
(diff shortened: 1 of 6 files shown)


client/scripts/client-unit-tests.sh
```diff
@@ -2,6 +2,6 @@
 # exit on error
 set -e
 
-php vendor/bin/phpunit -c tests/phpunit --coverage-php tests/phpunit/coverage/client-unit-tests.cov
+php vendor/bin/phpunit -c tests/phpunit --coverage-php tests/phpunit/coverage/client-unit-tests.cov --verbose
```
xaithe, line 5: We can remove verbose now

client/tests/phpunit/Validator/Constraints/CommonPasswordValidatorTest.php
```diff
@@ -13,10 +13,10 @@ class CommonPasswordValidatorTest extends TestCase
 
     public function setUp(): void
```
xaithe, line 14: We could just remove this function now instead of leaving it empty

xaithe (approved):

PR #988, opened 2022-10-21
DDPB-4538: Users cannot change their passwords themselves

Gugandeep:
## Purpose
After the upgrade to Symfony 5.4, users were no longer able to change their passwords themselves.

I initially believed that the issue was that the user provided current password was being hashed using the new password hashing algorithm, and then being compared to the password hash in the database which was done by a previous password hashing algorithm.

This ticket was to upgrade user passwords when they login.

Fixes DDPB-4538

## Approach

- Updated the `security.yml` config to migrate our sha512 hashed passwords to bcrypt.
- Updated the `UserRepository` to do the actual persisting to the database for the above.
- Updated the way in which the `UserController` compares the user provided current password.
- Updated the `SettingsController` to use send the correct key when making the set password request. (not sure this would have ever worked before?)

## Learning

- We are not using the `RedisUserProvider` as we thought we were.
- There are quite a few moving parts and behind the scene work done by Symfony when authenticating that can be hard to discover and follow.
- That we needed to update how we prepared and compared the current password provided by the user with the stored password in the database. The reason for this is because we are now using salts in passwords, and can no longer expect the password hasher to return the exact same hash as stored in our database.

## Checklist

### Frontend


coderste (approved):
LGTM 👍🏼

api/tests/Behat/features-v2/acl/login/login.feature
```diff
@@ -23,3 +23,16 @@ Feature: Users are not able to login with an invalid password, and the app preve
         Then I should see "You've entered an invalid email or password. Please try again."
         Given a super admin user tries to login with an invalid password
         Then I should see "You've tried to sign in with the wrong password too many times. Please check your password, wait for 30 minutes, and try again."
+
+    @lay-pfa-high-not-started-legacy-password-hash @gsc
```
coderste, line 27: Remove name tag?

coderste (changes requested):

jamesrwarren (approved):
looks good after demo from Gugs

jamesrwarren (approved):

coderste (approved):

PR #997, opened 2022-10-26
DDPB-4359: Missing radio button option for Money Out section

Gugandeep:
## Purpose
The option to select professional fees for the Money out section were not being displayed after the symfony upgrade.
These options are displayed depending on the role of the user that is logged in.

Fixes DDPB-4539

## Approach
Having played around with the logic that deals with displaying the options, I could not figure out why Symfony could not verify that the user had the role that was required.

After doing some looking around to see where else we were using the same logic to check user roles, I noticed we were only ever checking for a single role, and were passing in the role as a string. Here on the other-hand, we were passing in an array of roles.

Once I changed the code to pass in a single role as a string, the options re-appeared.

I then added a few behat tests to provide coverage of this logic.

## Learning

## Checklist

### Frontend


jamesrwarren (approved):
Nice

PR #1001, opened 2022-10-28
[DDPB-4477] add button to process Lay and Org CSVs in Digideps via a background task

coderste:
Added a fake csv file to the client side so we can import this into localstack S3 bucket. We also run a command in the background to process the CSVs. This background process is linked to the `KernelEvents::TERMINATE` listener. So the task will keep running until either it finishes or the server shuts down.

Keep in mind deploying to prod while this task is running will most likely mean we have to run it again.

Fixes DDPB-4477


client/src/Command/ProcessLayCSVCommand.php
```diff
@@ -0,0 +1,128 @@
...
+        try {
+            $this->s3->getObject([
+                'Bucket' => $bucket,
+                'Key' => $layDeputyReportFile,
+                'SaveAs' => "/tmp/layReport.csv"
+            ]);
+        } catch (S3Exception $e) {
+            if (in_array($e->getAwsErrorCode(), S3Storage::MISSING_FILE_AWS_ERROR_CODES)) {
+                $this->logger->log('error', sprintf('File %s not found in bucket %s', $layDeputyReportFile, $bucket));
+            }
```
mattmachell, line 63: You probably want a generic else clause with some error logging here for non-file missing error conditions, otherwise you won't detect connectivity or permissions issues.

client/src/Command/ProcessLayCSVCommand.php
```diff
@@ -0,0 +1,128 @@
...
+    protected function execute(InputInterface $input, OutputInterface $output): int
+    {
+        $bucket = $this->params->get('s3_sirius_bucket');
+        $layDeputyReportFile = 'layDeputyReport.csv';
+
+        try {
+            $this->s3->getObject([
+                'Bucket' => $bucket,
+                'Key' => $layDeputyReportFile,
+                'SaveAs' => "/tmp/layReport.csv"
```
mattmachell, line 58: Make sure you add code to delete the csv from the container once the process finishes or errors, otherwise personal data might get left around on the file system.

client/src/Controller/Admin/IndexController.php
```diff
@@ -526,8 +536,54 @@ public function uploadOrgUsersAction(Request $request)
...
+
+
+            $this->addFlash("notice", "CSV import process started. Keep an eye on your emails for completion.");
+
+            return $this->redirect($this->generateUrl('admin_org_upload'));
+        }
+
+        /** S3 bucket information */
+        $bucket = $this->params->get('s3_sirius_bucket');
+        $paProReportFile = 'paProDeputyReport.csv';
```
mattmachell, line 561: Worth making this use an environment var or constant so if naming changes at any point we can handle it easily.

client/src/Controller/Admin/IndexController.php
```diff
@@ -442,73 +438,87 @@ public function uploadUsersAction(Request $request, ClientInterface $redisClient
             }
         }
 
+        if ($processForm->isSubmitted() && $processForm->isValid()) {
+            /** Run the lay CSV command as a background task */
+            $email = $this->tokenStorage->getToken()->getUser()->getUserIdentifier();
+            $this->dispatcher->addListener(KernelEvents::TERMINATE, function () use ($email) {
```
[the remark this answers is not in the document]
coderste: Yeah we set the `set_time_limit(0)` in both the client side and api side here:
- https://github.com/ministryofjustice/opg-digideps/blob/main/api/app/console#L13
- https://github.com/ministryofjustice/opg-digideps/blob/main/client/app/console#L13

Basically we have no time limit on a script :)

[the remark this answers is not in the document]
mattmachell: Do those console controller settings apply for code triggered by the kernel terminate event?

mattmachell (approved):
Looks good from my side, but might be worth @Gugandeep looking at it as well in case I miss Symfony issues.

Gugandeep (approved):
Looks good to me.

PR #1006, opened 2022-11-01
DDPB-4513 create our own file scanner

jamesrwarren:
## Purpose
Have our own file scanner

Fixes DDPB-4513

## Approach
Initially looked at how sirius did their file scanning but it would have involved a fair amount of setup to work for us and quite a few major changes on the application side. I wanted to slot something in with as few app changes as possible so all our old tests would run to give us assurance.

I searched around for packages. Clam do their own docker image now but I couldn't guarantee they will keep it bang up to date and it still needs a rest container to interact with it. 

I found an all in one rest and clam container written in go and used that as the starting point though it was a bit out of date and had loads of stuff we didn't need in it. Made sure everything is updated to latest versions and removed all the useless bulk and redundant features from the code that we won't use. 

Hooked it up and got it working then spent about a bajillion hours (probably 70% of the tickets time) trying to get unit tests working.. Was very challenging to get a fake mock TCP server that works reliably with something like this. We should now be in full control of the image though so we can update packages/images as needed.

*EDIT* had to comment out the tests in circle. Tried many many things to get them to work to no avail. They work locally from command line and in vscode. I'm sure they will work in GH actions too. Not working in circle :(

## Learning
So struggling with the unit tests meant I did debug through every line of code many times which means I know the app well..
Relearned go testing
Learned about TCP servers in go and specifically about requests and how infuriatingly important you get the bytes size of buffer and new lines right.. 
Got pretty good at debugging tests in VS code

## Checklist
- [x] I have checked my work for potential security issues and refered to the [OWASP top 10](https://owasp.org/www-project-top-ten/)

### Frontend


xaithe (dismissed):

file_scanning/clamd/clamd.go
```diff
@@ -0,0 +1,229 @@
...
+		} else if strings.HasPrefix(s.Raw, "THREADS") {
+			stats.Threads = s.Raw
+		} else if strings.HasPrefix(s.Raw, "QUEUE") {
+			stats.Queue = s.Raw
+		} else if strings.HasPrefix(s.Raw, "MEMSTATS") {
+			stats.Memstats = s.Raw
+		} else if strings.HasPrefix(s.Raw, "END") {
+		} else {
+			return nil, errors.New(fmt.Sprintf("Unknown response, got %s.", s))
+		}
```
coderste, line 132: This could be written as a simple `switch` statement
```
switch {
case strings.HasPrefix(s.Raw, "POOLS"):
	stats.Pools = strings.Trim(s.Raw[6:], " ")
case strings.HasPrefix(s.Raw, "STATE"):
	stats.State = s.Raw
case strings.HasPrefix(s.Raw, "THREADS"):
	stats.Threads = s.Raw
case strings.HasPrefix(s.Raw, "MEMSTATS"):
	stats.Memstats = s.Raw
case strings.HasPrefix(s.Raw, "END"):
default:
	return nil, fmt.Errorf("invalid response, got %v", s)
}
```

Also change `errors.New()` to just use `fmt.Errorf` as that function just returns an `error` like you're expecting. Also based on the Go handbook, error's should not have capitalisation to start unless beginning with proper nouns or acronyms also they shouldn't end in punctuation

file_scanning/main.go
```diff
@@ -0,0 +1,144 @@
+package main
```
coderste, line 1: Other than the other comments most the code looks good to me. Only nit-picks I have are more about how we should follow the Go's handbook called [Effective Go](https://go.dev/doc/effective_go) when writing comments etc. I'll point them out in the review but will leave a `(can be ignored)` on the comment

file_scanning/main.go
```diff
@@ -0,0 +1,144 @@
...
+var opts map[string]string
+
+func init() {
+	log.SetOutput(ioutil.Discard)
+}
+
+// This is where the action happens.
+func scanHandler(w http.ResponseWriter, r *http.Request) {
+	switch r.Method {
+	//POST takes the uploaded file(s) and saves it to disk.
```
coderste, line 24: Should be a space after the `//` for the comment `// POST...`

`(can be ignored)`

file_scanning/clamd/conn.go
```diff
@@ -0,0 +1,143 @@
...
+)
+
+const CHUNK_SIZE = 1024
+const TCP_TIMEOUT = time.Second * 2
+
+var resultRegex = regexp.MustCompile(
+	`^(?P<path>[^:]+): ((?P<desc>[^:]+)(\((?P<virhash>([^:]+)):(?P<virsize>\d+)\))? )?(?P<status>FOUND|ERROR|OK)$`,
+)
+
+type CLAMDConn struct {
```
coderste, line 22: All exported types/functions in Go are what would be classed as `public` in a language like PHP. Therefore Go says they should all have comments to give an explanation on what it does because technically it's a public facing API and therefore can be used by other packages.

`(can be ignored)`

file_scanning/clamd/clamd.go
```diff
@@ -0,0 +1,229 @@
...
+)
+
+const (
+	RES_OK          = "OK"
+	RES_FOUND       = "FOUND"
+	RES_ERROR       = "ERROR"
+	RES_PARSE_ERROR = "PARSE ERROR"
+)
+
+type Clamd struct {
```
coderste, line 18: Again should really be commented following the Go handbook
https://github.com/golang/go/wiki/CodeReviewComments#comment-sentences

`(can be ignored)`

file_scanning/clamd/clamd.go
```diff
@@ -0,0 +1,229 @@
...
+		conn.Close()
+	}()
+
+	return ch, err
+}
+
+/*
+Check the daemon's state (should reply with PONG).
+*/
+func (c *Clamd) Ping() error {
```
coderste, line 78: Comment should start with the function name first 
https://github.com/golang/go/wiki/CodeReviewComments#comment-sentences

Also based on https://go.dev/doc/effective_go#commentary while Go does support `/* */` style comments `//` is the preferred one when commenting code and `/* */` is normally reserved for package level comments

`(can be ignored)`

coderste (changes requested):

file_scanning/main.go
jamesrwarren: Thanks Ste for the comments. They were really useful. I've added all the changes suggested. It was good doing the reading on the go standards. All really good stuff :)

file_scanning/clamd/conn.go
```diff
@@ -0,0 +1,143 @@
...
+	lenData := len(data)
+	buf[0] = byte(lenData >> 24)
+	buf[1] = byte(lenData >> 16)
+	buf[2] = byte(lenData >> 8)
+	buf[3] = byte(lenData >> 0)
+
+	a := buf
+
+	b := make([]byte, len(a))
+	for i := range a {
```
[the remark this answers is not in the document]
jamesrwarren: Done. good call

coderste (approved):

file_scanning/Dockerfile
```diff
@@ -0,0 +1,55 @@
+FROM alpine:3.16 as build
+
+# Update & Install Go
+RUN apk update upgrade && apk add --no-cache go
```
mattmachell, line 4: What was the reason for not using golang alpine `FROM golang:1.19.3-alpine as base`?

file_scanning/Dockerfile
```diff
@@ -0,0 +1,55 @@
...
+    && sed -i 's/^#TCPSocket .*$/TCPSocket 3310/g' /etc/clamav/clamd.conf \
+    && sed -i 's/^#Foreground .*$/Foreground true/g' /etc/clamav/freshclam.conf
+
+RUN freshclam --quiet --no-dns
+
+COPY entrypoint.sh /usr/bin/
+
+EXPOSE 8080
+
+ENTRYPOINT [ "entrypoint.sh" ]
```
mattmachell, line 55: Check you're not running the container as root.

coderste (dismissed):

file_scanning/Dockerfile
jamesrwarren: Well alpine is just under 3mb and golang is over a hundred.. having said that, installing go on alpine makes it even bigger. build time wise they're within a couple of seconds of each other so yes lets switch to golang one for the build container as it's less code :) Good call

jackgoodby (approved):

PR #1028, opened 2022-11-14
DDPB-4558 - User sessions that timeout result in a general application error

Gugandeep:
## Purpose
User sessions that timeout result in a general application error rather than clearing showing to the user that they have timed out.

Fixes DDPB-4558

## Approach
The way in which authentication is handed has changed since we updated to Symfony 5.4
Symfony is now producing 401 (Unauthorized) and 403 (Forbidden) exceptions instead of the custom exceptions we were expecting.

We now have listeners for these exceptions that will redirect the user to the right pages.

## Learning

## Checklist

### Frontend


client/translations/error-pages.en.yml
```diff
@@ -4,3 +4,8 @@ errors:
       htmlTitle: "503 - Service currently unavailable"
       pageTitle: "Service currently unavailable (503 error)"
       pageSectionDescription: "Try refreshing the page or returning later"
+  access-denied:
+    page:
+      htmlTitle: "Access Denied"
+      pageTitle: "Access Denied"
+      pageSectionDescription: "You do not have permission to view this page."
```
mattmachell, line 11: Is this the content from the previous error? Would be good to provide a remediation option "Your session may have timed out please log in again."

Gugandeep: Very good point. No this content is for the forbidden pages where you don't have the correct role. For example, if a regular admin tries to access the fixtures page which is restricted to super admins, this is the content that will display.

For users timing it, it is a little different.

From what I can tell, Symfony doesn't actually know that a user has timed out. When a user logs in, we set their token in Redis with an expiry time which tells Redis how long to keep the token for. When it expires in Redis, the next time Symfony goes looking for it, it just sees that the token not there and so marks it as invalid.

Put briefly, it doesn't look like Symfony can distinguish between someone trying to access a page whilst not being logged in and someone who has been timed out, because behind the scenes looks the same.

So previously, we didn't actually have any content to display to the user to tell them they had timed out probably because of the above. If we were to include some content, it would possible have to be quite generic.
Although that's probably better than nothing I suppose.

mattmachell: Yeah, handling this kind of thing gets a bit weird in the model that cookies/HTTP use. It's generally worth adding some kind of content to cover the option that the forbidden error is caused by a timeout as the generic error can confuse. 

We might have some error content in another service (make or use) we can repurpose, since I know we don't have much content support at the moment.

MiaGordon91 (approved):

PR #1038, opened 2022-11-30
[DDPB-4577] fixed bug where hybrid were not being caught in the process

coderste:
Also added a status indicator to the current process

Fixes DDPB-4577


MiaGordon91 (approved):
Looks good to me! :)

PR #1055, opened 2022-12-12
Ddpb 4528b

Gugandeep:
## Purpose

Fixes DDPB-####

## Approach

## Learning

## Checklist

### Frontend

### DIFF
--- a/api/src/v2/Registration/Uploader/LayDeputyshipUploader.php
+++ b/api/src/v2/Registration/Uploader/LayDeputyshipUploader.php
@@ -3,6 +3,7 @@
 namespace App\v2\Registration\Uploader;
 
 use App\Entity\PreRegistration;
+use App\Entity\Report\Report;
 use App\Entity\User;
 use App\Repository\ReportRepository;
 use App\v2\Registration\DTO\LayDeputyshipDto;
@@ -13,10 +14,7 @@ use Doctrine\Common\Persistence\Mapping\MappingException;
 use Doctrine\ORM\EntityManagerInterface;
 use Doctrine\ORM\OptimisticLockException;
 use Doctrine\ORM\ORMException;
-use Exception;
 use Psr\Log\LoggerInterface;
-use RuntimeException;
-use Throwable;
 
 class LayDeputyshipUploader
 {
@@ -66,7 +64,7 @@ class LayDeputyshipUploader
             $this
                 ->updateReportTypes()
                 ->commitTransactionToDatabase();
-        } catch (Throwable $e) {
+        } catch (\Throwable $e) {
             $this->verboseLogger->error($e->getMessage());
 
             return ['added' => $added, 'errors' => [$e->getMessage()]];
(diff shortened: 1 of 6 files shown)


api/src/v2/Registration/Uploader/OrgDeputyshipUploader.php
```diff
@@ -303,11 +299,21 @@ private function handleReport(OrgDeputyshipDto $dto)
```
coderste, line 299: I feel like this function has got too complex to read. Too many nested `if`'s statements. I _think_ we can simplify if we write something like this.

I haven't tested the below but if the function is the exact same as before I think going this way will help with readability

```
// rather than an else for if a report exists we can immediately 
// check if no report exists do what we have to do and then presist/flush
// and return out the function so it stops executing
if (!$report) {
    $report = new Report(
        $this->client,
        $dto->getReportType(),
        $dto->getReportStartDate(),
        $dto->getReportEndDate()
    );

    $this->client->addReport($report);

    $this->added['reports'][] = $this->client->getCaseNumber().'-'.$dto->getReportEndDate()->format('Y-m-d');

    $this->em->persist($report);
    $this->em->flush();

    return;
}

// next we checked if report was NOT submitted and if the un-submitted date was empty
// instead we check if the report has been submitted and that the un-submitted date is NOT empty
// so basically the opposite of the original conditions. If they are met we just return again.
// As we didn't have an else attached to this and we don't want to do anything if these are true.
if ($report->getSubmitted() && !empty($report->getUnSubmitDate())) {
    return;
}

// Now the rest of the function is the same as before just with 
// else nested if's and easier to follow and understand.
if (OrgDeputyshipDto::DUAL_TYPE == $dto->getHybrid()) {
    if ($this->client->getNamedDeputy()->getDeputyUid() == $dto->getDeputyUid()) {
        if ($report->getType() !== $dto->getReportType()) {
            $report->setType($dto->getReportType());

            $this->updated['reports'][] = $report->getId();
        }
    }
} else {
    if ($report->getType() !== $dto->getReportType()) {
        // Add audit logging for report type changing
        $report->setType($dto->getReportType());

        $this->updated['reports'][] = $report->getId();
    }
}

$this->em->persist($report);
$this->em->flush();
```

api/src/v2/Registration/Uploader/LayDeputyshipUploader.php
```diff
@@ -111,21 +109,32 @@ private function updateReportTypes(): LayDeputyshipUploader
...
+                        if ($currentActiveReport->getType() != $determinedReportType) {
+                            $currentActiveReport->setType($determinedReportType);
+                            $this->reportsUpdated[] = $reportCaseNumber;
+                        }
+                    }
+                } else {
+                    if ($currentActiveReport->getType() != $determinedReportType) {
+                        $currentActiveReport->setType($determinedReportType);
+                        $this->reportsUpdated[] = $reportCaseNumber;
+                    }
```
coderste, line 132: Not as bad here as the previous comment but I still think doing the below would keep the if's statements to just a 2 nested statement.

```
// For Dual Cases, deputy uid needs to match for the report type to be updated
if (PreRegistration::DUAL_TYPE != $preRegistration->getHybrid()) {
    if ($currentActiveReport->getType() != $determinedReportType) {
        $currentActiveReport->setType($determinedReportType);
        $this->reportsUpdated[] = $reportCaseNumber;
        continue;
    }
}

if ($currentActiveReport->getClient()->getUsers()[0]->getDeputyNo() == $preRegistration->getDeputyUid()) {
    if ($currentActiveReport->getType() != $determinedReportType) {
        $currentActiveReport->setType($determinedReportType);
        $this->reportsUpdated[] = $reportCaseNumber;
    }
}
```

coderste (commented):
I haven't tested the 2 comments I've made but they _should_ work as we haven't changed any logic, just changed the way the logic is processed just for readability. If you try the changes and they don't work how expected then just revert back to all the nested if's

coderste (approved):

PR #1058, opened 2022-12-15
[DDPB-4572] remove the ability to manually upload csv's

coderste:
## Purpose
Remove the ability to manually upload csv's. This does not solve the original tickets ACs but the comments on the ticket should explain why.

Fixes DDPB-4572


coderste:
PR closed. Running the cron on the client side would require a full authentication API token to access the resource and this is not good for security. We have a ticket to move this code to the API side and living there would mean we could run the command without the authentication route (_potentially_). 

I then removed the manual form in this ticket but forgot that our integration tests take use of the CSV manual uploader so that's not an option until we figure out a way to upload the correct CSVs for a integration test without the form

PR #1067, opened 2022-12-21
DDPB-4597 rotate digideps db secret

jamesrwarren:
## Purpose
Rotate the DB secrets on a nightly basis

Fixes DDPB-4597

## Approach
Created wrapper around connection. Password was previously stored as environment variable. Now it is stored in API redis. I had some misgivings about this but actually I think this is fine as:
- Anyone who gained access to the API container would be able to see environment variables (which is where it was stored previously) so this doesn't make the secret any more accessible than it was.
- Redis for our API container is separate endpoint to the frontend and has SG rules that only allow API container to access it. 
- Redis for our API container is not accessible from the internet as it's in private subnet
- Redis is encrypted at rest (if anyone were to gain access from aws to the backend data)

So the basic flow is:

- connection to db is called
- we update the password in params to be what is stored in redis
- ** if it succeeds then we keep connection open for remainder of the backend request (1 request can have multiple calls to the DB)
- if it fails we regrab the secret from secret manager, store it in redis and retry the connection.
- if this succeeds go to step **
- if this fails then we throw exception

This way we can rotate the secret value first but the app users will still be using the old value stored in redis which is needed as the DB won't have yet been updated. Then the DB is updated and the users first connection after this will fail and we will retry after grabbing the new secret from secrets manager (that is used as the input to the DBs secret)

## Learning
Most of learning came from trying the iam auth (that was too slow)

## Checklist

### Frontend


.github/workflows/scheduled-cycle-secrets.yml
```diff
@@ -0,0 +1,128 @@
+name: "[Scheduled] Cycle Secrets"
+
+on:
+  schedule:
+    # 5AM from Sunday to Friday
+    - cron: "0 5 * * 0-5"
```
charlesmmarshall, line 6: hey @jamesrwarren - any reason skip saturday?

charlesmmarshall (approved):

.github/workflows/scheduled-cycle-secrets.yml
jamesrwarren: Yeah this is not quite right actually. Going to update to Sun-Thur. It's to protect us from some issue breaking it over the weekend where we have no support basically. I can't see what exactly might break but as we're doing TF applies as part of this rotation there's always something that might break (one thought is where an updated version of something that wasn't pinned properly gets applied). We're somewhat protected as it does the environments one after the other rather than in parallel but still a possibility of differences.

Gugandeep (dismissed):

PR #1084, opened 2023-01-13
[DDPB-4115] Bring JavaScript up to a more modern standard

coderste:
## Purpose
This ticket refactors a lot of our frontend JavaScript. Stop relying on jQuery as vanilla JavaScript is supported by all major browser that we support.

Also added jest tests around the majority of the new files. Regarding the test files some could be expanded upon as in they don't cover a lot in the file they are testing other than testing the function is callable etc.

Currently I've just created a new folder called `modules_new` to store our new JavaScript modules in. And I've moved some and refactored some files that I believe live in the globals directory better e.g the `Redirector` file. Once all files have been rewritten to vanilla JS we can just rename `modules_new` back to `modules`

Fixes DDPB-4115

### Remaining
The files below are the ones that still need refactoring or potentially even deleted if they are no longer required.

- [ ] `bind.js` **delete**: I believe once everything is refactored we won't need this `bind.js` file anymore. Based off https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_objects/Function/bind but we shouldn't need to be doing anything special with the `this` keyword once refactored.
- [ ] `detailsExpander.js` **refactor**: This is adding classes and removing them on `keyup, input, paste, change` events for HTML tags with the `js-details-expander, js-details-expander-travel-costs, js-details-expander-specialist-service` class. Should probably live inside a page `folder` or something rather than a `modules` folder.
- [ ] `show-hide-content.js` **refactor** - big file will need a fair few tests to test the functions within. Believe it's doing all the show and hide on the radio buttons for forms that have an extra option and we need some extra input.
- [ ] `table-multiselect.js`**refactor** - to do with multi-select checkboxes. Relies heavily on jQuery but can be detached from it now.
- [ ] `uploadFile.js` **refactor** - some upload logic. Uses jQuery again but can be rewritten for vanilla JS. Doesn't use jQuery forms or anything so no need for `fetch` just some logic around stopping the form submitted at first and then resubmitting once some checks are done.

`ga.js/googleAnalyticsEvents.js` are sort of outliers I think the `googleAnalyticsEvents.js` does some form validation errors events etc. but not sure if these can all live inside tag manager now or not? Not too sure there though not that experienced with GA. If they can't then the `googleAnalyticsEvents.js` will probably need refactoring.

--- 

As a side-note, as the team is going to be short on developers for a little I don't believe this is a high enough priority to get picked up which is okay. Although I haven't finished it all -- as this is open source (yay!) I will probably look at just finishing this off when I get L&D time at my next place (as I don't like leaving stuff unfinished). If that happens then all I'd need is for someone to just merge and see the ticket through the pipeline :)


client/assets/javascripts/__tests__/globals/CharacterLimiter.test.js
```diff
@@ -0,0 +1,21 @@
...
+import CharacterLimiter from '../../globals/CharacterLimiter'
+
+describe('CharacterLimiter', () => {
+  it('should limit the number of characters on the required field', () => {
+    document.body.innerHTML = `
+    <div class="govuk-form">
+        <textarea class="govuk-textarea js-limit-chars-4" rows="5"></textarea>
+    </div>`
+
+    const textarea = document.querySelector('[class*="js-limit-chars-"]')
```
mattmachell, line 11: Is this JS just to avoid using the maxlength attribute on fields?

client/assets/javascripts/__tests__/modules_new/ButtonMenu.test.js
```diff
@@ -0,0 +1,40 @@
+import { describe, expect, it, jest } from '@jest/globals'
+import MOJButtonMenu from '../../modules_new/ButtonMenu'
+
+describe('MOJButtonMenu', () => {
+  describe('init', () => {
+    it('should call ButtonMenu if there are more than 2 children', () => {
+      document.body.innerHTML = `
+        <div class="moj-button-menu">
+          <div class="moj-button-menu__wrapper">
+            <a href="#" role="button" class="govuk-button">
```
mattmachell, line 10: Worth avoiding  `<a href="#" role="button"` and makign it an actual `<button>`

client/assets/javascripts/globals/Redirector.js
```diff
@@ -0,0 +1,12 @@
+const Redirector = function () {
+  const redirectUrlData = document.querySelectorAll('[data-redirect-url]')
+
+  if (redirectUrlData.length > 0) {
+    redirectUrlData.forEach(element => {
+      const url = element.getAttribute('data-redirect-url').trim()
+      window.location.href = url
```
mattmachell, line 7: Should really be url encoded to avoid potential XSS vector.

client/assets/javascripts/modules_new/TextAreaAutoSize.js
```diff
@@ -0,0 +1,22 @@
+const TextAreaAutoSize = {
+  init: function (formArea) {
+    const textAreas = formArea.querySelectorAll('[class*="js-auto-size"] textarea')
+
+    textAreas.forEach(textArea => {
```
mattmachell, line 5: Better to do event delegation rather than add a listener to every textarea.

client/assets/javascripts/__tests__/globals/CharacterLimiter.test.js
coderste: Believe so -- which now makes me think we don't actually need it and I can delete it :)

client/assets/javascripts/modules_new/TextAreaAutoSize.js
coderste: True I didn't even think about event delegation while rewriting these. Suppose that's what happens when you don't write it often anymore 😅

client/assets/javascripts/__tests__/modules_new/ButtonMenu.test.js
coderste: Seems like some of the `<a>` tags with the `role=button` have `href=""` values like here:

```
// client/templates/Admin/Client/Client/archived.html.twig:42
<a href="{{ path('admin_client_unarchived', {id: client.id}) }}" role="button" data-module="govuk-button" class="govuk-button govuk-button--secondary">
    Un-archive client
</a>
```

mattmachell: Yeah, some weird mixed messages in that code. If it's a link, then it goes somewhere, so it doesn't need the `role="button"` even if it is styled to look like one.

mattmachell: Looking at that code it may also mean we may be doing state changes (to unarchive) on a GET, which is bad.

client/assets/javascripts/globals/Redirector.js
```diff
@@ -0,0 +1,12 @@
+const Redirector = function () {
+  const redirectUrlData = document.querySelectorAll('[data-redirect-url]')
+
+  if (redirectUrlData.length > 0) {
+    redirectUrlData.forEach(element => {
+      const url = element.getAttribute('data-redirect-url').trim()
+      window.location.href = decodeURI(url)
```
mattmachell, line 7: Should be encode not decode.

coderste:
> Currently does odd things to textareas.

Yeah I spotted this while I was manually testing 😅 I am currently working on the fix. It's all to do with vanilla JS and jQuery not having exactly the same methods/functions/settings

mattmachell:
> > Currently does odd things to textareas.
> 
> Yeah I spotted this while I was manually testing 😅 I am currently working on the fix. It's all to do with vanilla JS and jQuery not having exactly the same methods/functions/settings

Don't worry about it, you are on leave. I have a fix started.

jackgoodby (approved):

PR #1106, opened 2023-01-30
DDPB-4483 move to github actions

jamesrwarren:
## Purpose
Move path to github actions. 

Fixes DDPB-4483

## Approach
Moved the docker files around a bit to make them faster. Split out the builds and ran as matrix.

Tried out using composite functions instead of reusable workflows but decided not to use them and did a little write up.

Made all the folders have - instead of _ as this seems to be consistent with what other projects are doing.

Otherwise it's pretty similar to our current pipeline.

Note the large number of files. A lot of that is folder renames.

## Learning
NA

## Checklist

### Frontend


docs/CI_PIPELINE.md
```diff
@@ -0,0 +1,80 @@
+## CI Pipeline
+
+### Decisions on CI product and style
+
+We are currently using github actions for our CI pipeline. We prefer this over circleci as it is
```
mattmachell, line 5: Phrasing here makes it more an ADR, i.e. decision to use GitHub. Might be worth moving there.

jackgoodby (dismissed):

jackgoodby (approved):
I have had a walkthrough with Jim

PR #1151, opened 2023-03-06
DDPB - 3134 Redirect after timeout

Raffers:
## Purpose
https://opgtransform.atlassian.net/browse/DDPB-3134?atlOrigin=eyJpIjoiYWE5N2U4Njg2ZTMwNDM5Yzg3YjY0OTliY2UwM2MyMzciLCJwIjoiaiJ9

Fixes DDPB-3134

## Approach
Adjusts exception class to pass value to URI to be handled by the login form.

## Learning

## Checklist

### Frontend


client/src/Validator/RouteValidator.php
```diff
@@ -0,0 +1,22 @@
...
+
+use Symfony\Component\Routing\Exception\ResourceNotFoundException;
+use Symfony\Component\Routing\RouterInterface;
+
+class RouteValidator
+{
+    public static function validateRoute(RouterInterface $router, string $path): bool
+    {
+        try {
+            $router->match($path);
```
mattmachell, line 15: Worth adding a few unit tests around this helper to make sure a full url rather than a path doesn't do anything weird, and to cover us for symfony changing behaviour at some future date. 

Basically make sure we avoid any nasty [open redirect](https://cheatsheetseries.owasp.org/cheatsheets/Unvalidated_Redirects_and_Forwards_Cheat_Sheet.html ) issues.

client/src/Security/LoginFormAuthenticator.php
```diff
@@ -72,7 +71,13 @@ public function authenticate(Request $request): Passport
 
     public function onAuthenticationSuccess(Request $request, TokenInterface $token, string $firewallName): ?Response
     {
-        $redirectUrl = $this->redirector->getFirstPageAfterLogin($request->getSession());
+        $redirectUrl = (
+            $request->query->get('lastPage') &&
+            RouteValidator::validateRoute($this->router, $request->query->get('lastPage'))
+        ) ?
+            urldecode($request->query->get('lastPage')) :
```
mattmachell, line 78: url-decoding here, but not in what you pass to the validator? So the validator is working on an encoded url? Might potentially lead to unexpected behaviour.

client/src/EventListener/UnauthorizedExceptionListener.php
```diff
@@ -11,11 +11,14 @@ class UnauthorizedExceptionListener
 {
     public function onKernelException(ExceptionEvent $exceptionEvent)
     {
+        $request = $exceptionEvent->getRequest();
         $exception = $exceptionEvent->getThrowable();
 
         if ($exception instanceof HttpException) {
             if (Response::HTTP_UNAUTHORIZED == $exception->getStatusCode()) {
-                $response = new RedirectResponse('/login?from=api');
+                $url = '/login?from=api&lastPage=' . urlencode($request->getRequestUri());
```
mattmachell, line 19: Do we actually use the from=api anywhere? Seems like an odd construct.

client/src/Security/LoginFormAuthenticator.php
```diff
@@ -57,8 +58,6 @@ public function authenticate(Request $request): Passport
                 } catch (AuthenticationException $e) {
                     throw $e;
```
mattmachell, line 59: I know this isn't your code, but this is an odd pattern if we aren't doing anything before just re-throwing the same exception. Would expect logging or something.

client/src/EventListener/UnauthorizedExceptionListener.php
mattmachell, line 19: Additional thought. Can you craft a request url to Digideps that throws an unauthorised and injects arbitrary values here that are passed down the chain?

mattmachell:
Added some comments on the general code and a few places where it might be worth checking that Symfony has our back in terms of values getting injected.

I think the obvious missing thing is any tests at either unit or behat level to cover the functionality.

client/src/Validator/RouteValidator.php
Raffers: You're right, had skipped the tests as there was a lot more work for the other files I had touched. So we've got a ticket to adjust the testing for those. I was being overly trusting of Symfony, will add some now.

client/src/EventListener/UnauthorizedExceptionListener.php
Raffers: Not that I've seen. I thought it was weird, at a guess I'm assuming it's stating the error has been passed from the API

client/src/Security/LoginFormAuthenticator.php
Raffers: I had assumed that the logging was done elsewhere, part of the brute force protection.

client/src/EventListener/UnauthorizedExceptionListener.php
Raffers: I've tried to pass random values to it, trying to craft path traversal, offsite, other user valid URL's & special characters. They are all passed back into the URL but when it meets the validator it's all correctly ignored.

client/src/Security/LoginFormAuthenticator.php
Gugandeep: The exception is caught by the `onAuthenticationFailure()` method further down which then decides what to do with it. In this case, it sets the error and redirects to the login page (the login page will check the exception and show the appropriate error message)

Gugandeep (approved):

client/src/EventListener/UnauthorizedExceptionListener.php
jackgoodby: [Used here](https://github.com/ministryofjustice/opg-digideps/blob/main/client/src/Controller/IndexController.php#L110). Convoluted bit of code though!

PR #1189, opened 2023-03-20
DDPB-4623 add github actions path to live

jamesrwarren:
## Purpose

Path to live for github actions

Fixes DDPB-4623

## Approach

Did path to live and commented out all the circle config. Will do a final removal of circle config when we're happy everything works and combine it with the org infra change to remove the circle vars.

## Learning
NA

## Checklist

### Frontend


.github/workflows/_build-and-push.yml
```diff
@@ -109,6 +116,17 @@ jobs:
           IMAGE_TAG: ${{ inputs.tag }}
         run: echo "Tag to use - ${IMAGE_TAG}"
 
+      - name: tag and push development container
```
mattmachell, line 119: name says dev, but the branch check below is main.

.github/workflows/workflow-path-to-live.yml
```diff
@@ -0,0 +1,347 @@
...
+      - scale_services_up
+    with:
+      workspace: integration
+      task_name: "integration_test_v2"
+      timeout: "500"
+      override: "sh,./tests/Behat/reset-db.sh"
+    secrets: inherit
+
+  integration_tests_1:
+    name: integration tests frontend 1
```
mattmachell, line 164: There's a built of repetition on these integration_tests_* steps. Wonder if there's a nicer way to pass in the task name and tag once rather than repeat?

.github/workflows/workflow-path-to-live.yml
```diff
@@ -0,0 +1,347 @@
...
+
+  end_of_workflow:
+    name: end of workflow
+    runs-on: ubuntu-latest
+    needs:
+      - restore_to_preproduction
+      - codecov
+      - workflow_variables
+    steps:
+      - name: workflow has ended without issue
```
mattmachell, line 318: Minor thing. I'd be tempted to change this to "workflow ended succesfully", the "without issue" sounds a bit negative.

.circleci/config.yml
```diff
@@ -1,604 +1,604 @@
...
-  scheduled_workspace_deletion:
-    triggers:
-      - schedule:
-          cron: "0 8,20 * * 0-6" # twice a day at 8am and 8pm
-          filters: { branches: { only: [main] } }
-    jobs:
-      - destroy-workspaces:
-          name: destroy non protected workspaces
-          filters: { branches: { only: [main] } }
+#  dependabot_app:
```
mattmachell, line 4: Any particular reason for not just deleting this and using git history if you need to check it?

.github/workflows/workflow-path-to-live.yml
```diff
@@ -0,0 +1,347 @@
...
+    steps:
+      - uses: actions/checkout@2541b1294d2704b0964813337f33b291d3f8596b # pin@v3
+      - name: extract variables for workflow
+        id: variables
+        run: |
+          echo "short_sha=$(echo ${GITHUB_SHA:0:7})" >> $GITHUB_OUTPUT
+
+  docker_build_scan_push:
+    name: build, scan and push
+    uses: ./.github/workflows/_build-and-push.yml
```
mattmachell, line 43: The reused workflow do make this much easier to read!

client/Dockerfile-dev
```diff
@@ -1,5 +1,3 @@
-ARG AWS_REGISTRY
-ARG VERSION
-FROM $AWS_REGISTRY/digideps/client:$VERSION
+FROM client:latest
```
mattmachell, line 1: Where do we use this Dockerfile? Is it just deployed to the main dev environment or PRs as well? Might cause oddities if so.

.github/workflows/_build-and-push.yml
jamesrwarren: Yeah that's correct. This is for the annoying development environment that has the no robots setup which we use for external users. I was actually going to ask about environments when I'm back on digideps as we have too many imo and we may be able to ditch some. So the branch is main but the workspace to be used is development but this is the build part so we only need to build this image when it's a merge to main if that makes sense :)

.github/workflows/workflow-path-to-live.yml
```diff
@@ -0,0 +1,347 @@
...
+    secrets: inherit
+
+  scale_services_up:
+    name: scale up services
+    uses: ./.github/workflows/_scale-services.yml
+    needs:
+      - workflow_variables
+      - terraform_apply_integration
+    with:
+      replicas: 6
```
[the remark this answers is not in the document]
jamesrwarren: Yeah, I don't think much has changed since i set this up and I did test it out with different numbers and this was the sweet spot. I could have a go at rejigging this at some point.

.github/workflows/workflow-path-to-live.yml
jamesrwarren: I could make them variables I guess that get output from workflow_variables but I don't think it makes it more readable. Could do that though

.github/workflows/workflow-path-to-live.yml
jamesrwarren: Hehe, ok sold :)

.github/workflows/_build-and-push.yml
mattmachell: Ah, cool. Wasn't clear at first glance, so worth specifying that in the step name. "tag and push user research development container" or something.

.github/workflows/workflow-path-to-live.yml
mattmachell: Thinking about it, might be better to leave it for a later ticket to avoid bloating this.

client/Dockerfile-dev
jamesrwarren: So this is part of the build to create the development environment image. The development environment image is largely the same as our normal image with a minor tweak so we build client and tag it with client:latest locally and we then build this image from that so it's not pulling from any repository or anything :)

Gugandeep (approved):

PR #1215, opened 2023-04-04
remove fullstop from start of alarm

xaithe:
Left a fullstop in an alarm name, should probably remove that


Gugandeep (approved):
Missed this yesterday - glad you were able to spot it!

PR #1255, opened 2023-05-23
DDPB-4657b - Convert unicode characters in names

Gugandeep:
## Purpose
A second fix to convert unicode characters for users attempting to register to the service via the invitation process.

Fixes DDPB-4657

## Approach

## Learning

## Checklist

### Frontend


xaithe (approved):

api/src/Controller/PreRegistrationController.php
```diff
@@ -91,4 +93,11 @@ public function clientHasCoDeputies(string $caseNumber)
     {
         return $this->preRegistrationVerificationService->isMultiDeputyCase($caseNumber);
     }
+
+    private function replaceUnicodeChars(array $clientData)
```
mattmachell, line 97: Might be better to name this `normaliseCharacters` or something instead as it's not a blanket replace of all unicode, just some characters we know get confused across the system.

api/tests/Behat/bootstrap/v2/Registration/ActivateTrait.php
```diff
@@ -24,6 +24,14 @@ public function preRegistrationDetailsExistToAllowALayDeputyToRegisterForTheServ
         $this->existingPreRegistration = $this->fixtureHelper->createPreRegistration();
     }
 
+    /**
+     * @Given pre-registration details exist with no unicode characters$/
+     */
+    public function preRegistrationDetailsExistWithNoUnicodeCharacters()
+    {
+        $this->existingPreRegistration = $this->fixtureHelper->createPreRegistration('OPG102', 'PFA', 'O\'Shea');
```
mattmachell, line 32: So we scaffold this as if it was propagated from Sirius. Do we know if there might be cases where Sirius has confused characters and so we need to normalise both sides. i.e. Sirius has `O’Shea `or `O`Shea` and the user enters `O'Shea`, opposite of the bug this fixes.

Gugandeep: Yeah so it can be the case that the data has been entered incorrectly on the Sirius side or even when the court order was served. 
I'm not sure if normalising the data we receive from Sirius is that good of an idea, as it's supposed to be our source of truth. I think in Lay cases where the names are wrong on Sirius, we usually wait for them to be updated there first and then allow the deputy to register. Although this is linked to another issue where we are not currently updating names when they get updated on Sirius.

It is a good point though, and maybe we need to make sure that both Digideps and Sirius are using the same character sets. 

I was also potentially thinking of using [the slugger component](https://symfony.com/doc/current/components/string.html#slugger) in Symfony, but was worried that it could convert more characters than expected and result in some of the problems you mentioned above. So I stuck to the known characters that were causing issues for now.

mattmachell: Yeah, I wouldn't blanket change the way we ingest, but at compare time it would be useful to be comparing like and like. Slugger would likely convert too much. 

It's not even the character sets, so much as confusable characters and variance on who input data. You used to get the same issue with content pasted from word, where everything became smart quotes not apostrophes. The list is quite large: https://util.unicode.org/UnicodeJsps/confusables.jsp?a=%27&r=None

mattmachell (approved):

PR #1278, opened 2023-06-05
DDPB-4701 add double click prevention

jamesrwarren:
## Purpose

Fixes DDPB-4701

## Approach
Got rid of two bespoke double click protection JS scripts we had in favour of applying the govuk protection that is inbuilt to the govuk-button.

We will always want to apply this so I have some js that applies it to all objects that have the govuk-button tag.

Added a single bit of bespoke protection for download link... Considered having it on all the links but the others should take you to another page anyway... still might be worth while. Wasn't sure on styling etc so thought I'd just get this problem link fixed and we can talk about styling as no govuk element it seems.

Made sure all the buttons that were green standard buttons were govuk-buttons as I can see no reason they wouldn't be.

## Learning
NA

## Checklist
- [x] I have checked my work for potential security issues and refered to the [OWASP top 10](https://owasp.org/www-project-top-ten/)

### Frontend


Gugandeep (dismissed):
Looks good to me.

My approach would have been to add the attribute to the macros which we use to create the buttons, but there's no guarantee that we're using these macros every-time we create a button.

This approach seems better in that sense. Great that you also cleaned up the previous javascript that was responsible for this.

client/templates/Report/Report/_pdf_download_link.html.twig
```diff
@@ -2,6 +2,23 @@
...
+<style>
+    .disabled {
+        pointer-events: none;
+        opacity: 0.5;
+        /* You can style the disabled link as per your requirement */
+    }
+</style>
+
+<script>
+    function disableLink(link) {
```
mattmachell, line 19: Don't add any inline scripts, they'll be a problem when we fix the content security policy.

client/assets/javascripts/__tests__/modules_new/PreventDoubleClick.test.js
```diff
@@ -0,0 +1,27 @@
...
+    const button2 = document.createElement('button')
+    button2.classList.add('govuk-button')
+    document.body.appendChild(button2)
+
+    // Call the init function
+    PreventDoubleClick.init(document)
+
+    // Assert that the buttons have the "data-prevent-double-click" attribute
+    expect(button1.getAttribute('data-prevent-double-click')).toBe('true')
+    expect(button2.getAttribute('data-prevent-double-click')).toBe('true')
```
mattmachell, line 25: Only tests the attribute is there, not that it prevents double click.

client/templates/Report/Report/_pdf_download_link.html.twig
jamesrwarren: Ah thanks! Will change this now!

client/assets/javascripts/__tests__/modules_new/PreventDoubleClick.test.js
mattmachell: No worries, if it's part of the GDS component then we must assume they tested the actual attribute detection works.

client/templates/Report/Report/_pdf_download_link.html.twig
jamesrwarren: Wasn't a nice place to just put the function so had to follow the pattern for the others. Can you check and tick if that's ok? Added a test and checked manually and it works.

mattmachell (approved):

PR #1288, opened 2023-06-13
Upgrade AWS SDK here as well.

mattmachell:


charlesmmarshall (approved):

jamesrwarren:
I could be wrong, only had a quick look but I think it's the name of your branch. Because it starts with aws I don't think it's allowing you to create param store values with that name and hence TF is failing. I'd start a branch with a different name that doesn't start with AWS and it should work fine :)

mattmachell:
Thanks Jim, will move to a different branch name.

PR #1302, opened 2023-07-03
DDPB-4726 add new DNS records

jamesrwarren:
## Purpose
Add additional DNS records

Fixes DDPB-4726

## Approach
Added certs and new domain in previous PRs. This just adds non prod domains. We don't do prod as the SG rules on prod mean it would be open to public and we don't want them having a different route in.

## Learning
NA

## Checklist

### Frontend


jackgoodby (approved):
Acceptable.

PR #1312, opened 2023-07-10
DDPB-4724: Unhandled Auth Error - Raises 500

MiaGordon91:
## Purpose
Throw exception when user token is null.

Fixes DDPB-4724

## Approach

## Learning

## Checklist

### Frontend

### DIFF
--- a/api/src/Controller/AuthController.php
+++ b/api/src/Controller/AuthController.php
@@ -14,7 +14,6 @@ use Psr\Log\LoggerInterface;
 use Symfony\Component\Routing\Annotation\Route;
 use Symfony\Component\Security\Core\Authentication\Token\Storage\TokenStorageInterface;
 use Symfony\Component\Security\Core\Exception\UserNotFoundException;
-use Throwable;
 
 /**
  * @Route("/auth")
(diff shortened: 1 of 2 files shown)


api/src/Controller/AuthController.php
```diff
@@ -43,21 +42,27 @@ public function login(
...
+                $user->setLastLoggedIn(new \DateTime());
+                $em->persist($user);
+                $em->flush();
 
-            // add token into response
-            $restInputOutputFormatter->addResponseModifier(function ($response) use ($authToken) {
-                $response->headers->set(HeaderTokenAuthenticator::HEADER_NAME, $authToken);
-            });
+                // Now doing this inline rather than injecting RedisUserProvider
+                $authToken = $user->getId().'_'.sha1(microtime().spl_object_hash($user).rand(1, 999));
```
mattmachell, line 56: Is this just to generate a unique ID key for the token in Redis? Might be easier to use a uuid generator?

mattmachell: Also, the user_id is a prefix on the redis key outside the hash? Might be better so be inside the hash if we never need to reference it to avoid any undue info leakage.

MiaGordon91: Thanks for your feedback @mattmachell 
I've chatted to Jack and we're going to pull this into a separate tech debt ticket.

jackgoodby (approved):

PR #1314, opened 2023-07-12
DDPB-4735 split out frontend container

jamesrwarren:
## Purpose
Split out frontend container and end TLS at the LB

Fixes DDPB-4735

## Approach
This is a fairly hefty refactor with a fairly hefty amount of changes. Might be worth writing out the logic here so it can be understood before reviewing. Happy to change some bits around or names of things!

Previous folder structure:

image

We move this under client/app and split out the docker and confd setup into a docker folder and the webpack/js/css frontend stuff into a folder called resources. Structure looks like this:

image

We add a container for building the node resources that maps to a folder called public locally and use the docker volume mapping to map the public folder into both the nginx container and the app container.

For CI, we build the frontend web resources as a separate runner and bring them in as an artifact for client (app) and client-webserver (nginx). I normally like to have all my bits like tests and builds running in containers so errors are reproducible locally (even if this means slightly slower builds) but in this case it was too much of a hit. I manage drift by extracting the node version to use from the dockerfile so it only needs updating in one place.

We then modify the task definition so that it includes both containers and amend firewall sg rules to allow port 80 from LB to the app. 

ISSUE: for some unknown reason the integration tests are sometimes going via http rather than https when pressing buttons. As such I have had to allow port 80 to LB and added rerouting to 443 from 80 for both LBs. I have tried printing out current urls and base urls in all the places it sets this as part of the tests and it always seems to go to https according to that and yet somehow it is not! :(

We modify the nginx config to manage the new setup and make sure we have the same security headers as well as some additional ones that I believed were missing. Took the opportunity to harden the containers for additional security.

That is the crux of the PR. 

ISSUE: The integration tests seem slower in some places and faster in others. Would be nice to work out why.

## Learning
NA

## Checklist

### Frontend


client/app/config/services.yml
```diff
@@ -260,7 +260,7 @@ services:
         base_uri: "%api_base_url%"
         verify: false
         headers:
-          Content-Type: application/json
+          Content-Type: app/json
```
mattmachell, line 263: Find and replace issue?

jamesrwarren: Yeah, this isn't ready for review yet. That button shouldn't have been clicked.

jackgoodby: Mea culpa - Jim and I misunderstood each other

.github/workflows/_web-resources.yml
```diff
@@ -0,0 +1,42 @@
...
+        uses: actions/cache@04f198bf0b2a39f7230a4304bf07747a0bddf146 # pin@v3.0.8
+        env:
+          cache-name: cache-node-modules
+        with:
+          path: ./web-resources/public
+          key: ${{ runner.os }}-build-${{ env.cache-name }}-${{ hashFiles('**/web-resources/package-lock.json') }}
+      - name: build assets
+        run: |
+          # Copy in templates and translations
+          cp -r ../app/templates ./templates
```
mattmachell, line 29: Noticed this in passing. Templates are PHP/Twig code so shouldn't be needed in the node assets or the web server.

jamesrwarren: yeah this was in the build image from before. I thought it seemed strange and tried removing it but it was still expecting a file somewhere... I must have been misunderstanding the message though as I have now removed that cp in and it still seems fine as you would expect :)

Gugandeep (approved):

PR #1331, opened 2023-07-24
DDPB-4597 add db secret rotation to cyclesecrets script

jamesrwarren:
## Purpose

Rotate the db password at same time as secrets

Fixes DDPB-4597

## Approach
There were issues in some environments due to 2 factors:

- Secrets are paginated so we need to bring them all back.
- Apply immediately is turned off by default on some of our envs to stop patches getting applied during normal hours.

Only solution for the second issue that I could think of is to apply the DB secret update as part of the rotation job as there's no way of doing this in a non messy way in terraform. This should have the affect that connections will continue to work even if the tf apply fails. The TF apply is simply to update the container specs so everything is fully synced though it's not totally necessary now.

## Learning
NA

## Checklist

### Frontend

### DIFF
--- a/.github/workflows/scheduled-cycle-secrets.yml
+++ b/.github/workflows/scheduled-cycle-secrets.yml
@@ -1,108 +1,107 @@
-#name: "[Scheduled] Cycle Secrets"
-#
-#on:
-#  schedule:
-#    # 5AM from Monday to Friday
-#    - cron: "0 5 * * 1-5"
-#
-#permissions:
-#  contents: read
-#  security-events: none
-#  pull-requests: none
-#  actions: none
-#  checks: none
-#  deployments: none
-#  issues: none
-#  packages: none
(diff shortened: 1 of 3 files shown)


.github/workflows/scheduled-cycle-secrets.yml
```diff
@@ -1,117 +1,115 @@
...
+  checks: none
+  deployments: none
+  issues: none
+  packages: none
+  repository-projects: none
+  statuses: none
+
+jobs:
+  latest_deployed_image:
+    name: development environment apply terraform
```
xaithe, line 22: Is this the right name for the job?

jamesrwarren: Nope! Thanks

charlesmmarshall (approved):

PR #1345, opened 2023-08-14
Ddpb 3881b Improve CSP posture for frontend

jamesrwarren:
## Purpose
To remove unsafe-inline from csp for src-scripts 

Fixes DDPB-3881

## Approach
To remove unsafe-inline from csp for src-scripts we needed to move the bit of code that enables JS and the 3 bits of inline google analytics code to separate JS files that get bundled by webpack.

Also added  X-XSS-Protection header as we should be using this. 

Split out the other sources to be a bit more granular. Makes the line a bit more ugly looking but it's better to be granular. 

Set up GA4 for local in analytics and checked this all still is receiving data. To do this, I accessed particular pages and checked they appeared in analytics.

Example:

image

**Information:** [GoogleAnalyticsObject.js](https://github.com/ministryofjustice/opg-digideps/pull/1345/files#diff-e21e38dded19f81cb030081e1f7e43c535a5cd31c1daeb8fdfbc5afbe18a5827) now seems to be redundant but it's sending stuff to GDS so my thoughts are run this in as is to get the CSP working then do a small PR to remove this and update the tagging and check the GDS analytics is now working correctly. I did check if we could remove this and everything still worked when we did and docs suggest it is so.

## Learning
NA

## Checklist

### Frontend

### DIFF
--- a/.github/workflows/_unit-tests-client.yml
+++ b/.github/workflows/_unit-tests-client.yml
@@ -41,6 +41,7 @@ jobs:
         run: |
           APP_ENV=dev APP_DEBUG=0 docker-compose -f docker-compose.yml -f docker-compose.ci.client-unit-tests.yml up -d --no-deps client-unit-tests pact-mock
           sleep 3
+          docker-compose -f docker-compose.yml -f docker-compose.ci.client-unit-tests.yml exec -T client-unit-tests chown root /tmp/commonpasswords.txt
           docker-compose -f docker-compose.yml -f docker-compose.ci.client-unit-tests.yml exec -T client-unit-tests sh scripts/client-unit-tests.sh
           docker-compose -f docker-compose.yml -f docker-compose.ci.client-unit-tests.yml exec -T client-unit-tests chmod -R 777 tests/phpunit/coverage/client-unit-tests.xml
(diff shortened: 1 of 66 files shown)


.github/workflows/_unit-tests-client.yml
```diff
@@ -41,6 +41,7 @@ jobs:
         run: |
           APP_ENV=dev APP_DEBUG=0 docker-compose -f docker-compose.yml -f docker-compose.ci.client-unit-tests.yml up -d --no-deps client-unit-tests pact-mock
           sleep 3
+          docker-compose -f docker-compose.yml -f docker-compose.ci.client-unit-tests.yml exec -T client-unit-tests chown root /tmp/commonpasswords.txt
```
jamesrwarren, line 44: this doesn't always cause a problem and couldn't work out why but for some caches the fact that unit tests are run as root and this file is owned by www-data, it causes an issue. This isn't part of the build and this 'fixes' the issue. I did try running unit tests as www-data but that caused other permission issues.

jackgoodby (approved):

jackgoodby (approved):

PR #1351, opened 2023-08-22
DDPB-4570: Delete Staff users who don't exist in Sirius 

MiaGordon91:
## Purpose
Remove staff accounts that have been inactivate for less than 2 years, yet do not exist in Sirius as we use Sirius as a source of truth for user accounts. 

Fixes DDPB-4570

## Approach
- Created a delete sql query 
- Query is called by a custom command 
- Added logic in the command that excludes current digideps team

## Learning

## Checklist
- [x ] I have performed a self-review of my own code

### DIFF
--- a/api/src/Command/StaffUserDeletionCommand.php
+++ b/api/src/Command/StaffUserDeletionCommand.php
@@ -1,0 +1,64 @@
+<?php
+
+namespace App\Command;
+
+use App\Repository\UserRepository;
+use Symfony\Component\Console\Command\Command;
+use Symfony\Component\Console\Input\InputArgument;
+use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Output\OutputInterface;
+
+class StaffUserDeletionCommand extends Command
+{
+    public static $defaultName = 'digideps:nonexistent-sirius-staff-accounts';
+
+    public static array $digidepsStaffToExcludeFromDeletion =
+        [
+            35636 => 'gugandeep.chani@digital.justice.gov.uk',
+            34368 => 'james.warren@digital.justice.gov.uk',
+            49768 => 'nicola.mcmillan@digital.justice.gov.uk',
+            46417 => 'oludara.oginni@digital.justice.gov.uk',
+            29876 => 'jack.goodby@digital.justice.gov.uk',
+            45515 => 'mia.gordon@digital.justice.gov.uk',
+            33833 => 'caroline.hufton@digital.justice.gov.uk',
+            43009 => 'jude.rattle@digital.justice.gov.uk',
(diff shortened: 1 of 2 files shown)


api/src/Command/StaffUserDeletionCommand.php
```diff
@@ -0,0 +1,64 @@
...
+            35636 => 'gugandeep.chani@digital.justice.gov.uk',
+            34368 => 'james.warren@digital.justice.gov.uk',
+            49768 => 'nicola.mcmillan@digital.justice.gov.uk',
+            46417 => 'oludara.oginni@digital.justice.gov.uk',
+            29876 => 'jack.goodby@digital.justice.gov.uk',
+            45515 => 'mia.gordon@digital.justice.gov.uk',
+            33833 => 'caroline.hufton@digital.justice.gov.uk',
+            43009 => 'jude.rattle@digital.justice.gov.uk',
+            43411 => 'tom.gulliver@digital.justice.gov.uk',
+            43177 => 'peter.scottreid@digital.justice.gov.uk',
```
jackgoodby, line 26: Might wanna give Danny Collins the headsup before we delete him, also...

jackgoodby: OK ignore that - I've just checked with Danny and he has two accounts, and we're ok to delete the one as it stands

api/src/Repository/UserRepository.php
```diff
@@ -341,4 +341,20 @@ public function deleteInactiveAdminUsers(array $inactiveAdminUserIds)
...
+    {
+        $em = $this->getEntityManager();
+        $rsm = new ResultSetMappingBuilder($em);
+
+        $sql = 'DELETE FROM dd_user WHERE id IN (:ids)';
+        $params = [
+            'ids' => $staffUsersToBeDeleted,
+        ];
+
+        $stmt = $em->createNativeQuery($sql, $rsm);
```
jackgoodby, line 355: Have you checked this works on Dev? I was under the impression createNativeQuery() was just for SELECT queries rather than updates/deletes...  (if it works it works, though!) https://www.doctrine-project.org/projects/doctrine-orm/en/2.16/reference/native-sql.html#the-nativequery-class

jackgoodby (commented):
Please can we chat about this approach in a bit after meetings - just want to check it actually works as you think – let's go nuke a dev env later!

Normally I'd stay away from involving Doctrine and the entity manager on a bulk operation job, but I know there are only a handful of rows, so there'll be no massive performance issue I shouldn't think.

Would be nice to show some documentation for how to use the command – though as it's a one-time operation maybe not necessary - oh, we should raise another ticket to remove this code after it's run.

PR #1353, opened 2023-08-24
DDPB-4756 - move environment modules into modules folder

xaithe:
Moves all the environment terraform modules into a modules folder, as per the OPG Terraform Guidance


environment/modules/aurora/.terraform.lock.hcl
jamesrwarren, line 1: We should probably delete these out and put them in .gitignore for anything but root folder. I know this is just the validation that keeps creating them.

jamesrwarren (approved):

PR #1387, opened 2023-09-19
DDPB-4774 - Add rate limiting for password reset token consumption

MiaGordon91:
## Purpose
Apply rate limiting to protect against brute force attacks on the consumption of password reset tokens. 

Fixes DDPB-4774

## Approach
- Symfony rate limiter component installed and injected into controller
- Request limit and window is configured in yml file
- Token now includes users id to identify and rate limit against the user 
- Each unsuccessful and successful request is consumed, once the limit is reached an error message is rendered on the UI.
- Feature tests added to test password reset flow as there weren't any previously 

## Learning

## Checklist

### Frontend


client/app/src/Controller/UserController.php
```diff
@@ -244,14 +256,14 @@ public function passwordForgottenAction(Request $request)
 
         return [
             'form' => $form->createView(),
+            'user' => $user->getId(),
```
Gugandeep, line 259: I don't think we ended up making use of this, so might be worth dropping.
I could be wrong though

client/config/parameters.yml
```diff
@@ -0,0 +1,69 @@
+# This file is auto-generated during the composer install
```
Gugandeep, line 1: I think this would normally be apart of our .gitignore

Probably shouldn't commit this to be safe.

Gugandeep (commented):
Left a few comments.

I think a few extra things have been committed probably as a result of the nginx container split ticket.

Don't think we want to commit the `public` folder, as I think everything in there is now in the `resources` folder.
You probably also want to drop the `parameters.yml` too.

PR #1478, opened 2023-12-01
DDLS-37 - update RDS certs

xaithe:
Updates our RDS to use the latest AWS CA cert.

This shouldn't require any downtime as can be updated in place - checked in branch env


jamesrwarren (approved):
This looks good to me.

PR #1492, opened 2023-12-18
DDLS-45 Automate CSV Upload

Raffers:
## Purpose
Change the trigger for the CSV job to a schedule to remove the need to manually click a button for it. 

We currently have a process whereby we manually run each CSV asynchronously by clicking buttons to run the CSV on Lay and the combined PA/professional CSV. The Lay CSV takes about ten minutes and the PA/Prof CSV somewhere between 45 minutes to an hour and we have to wait until the first has been completed to do the next one. 

While this is an improvement on the previous way of doing things, it would be beneficial to automate it entirely. 

-     To avoid the manual error of not running it and failing to make updates

-     Not having to plan cover on holidays and remembering to re-allocate the job on unexpected days off 

-     Avoiding the run being unsuccessful e.g. when we run it at the time as a deployment 

Fixes DDLS-45

## Approach
Migrated most aspects of the front-end job over to the API, refactors some of the API classes so the process is shareable between the command & manual approach 

## Learning
We still lack loads of unit tests/require refactoring so they can be tested.

## Checklist

### Frontend


api/app/src/v2/Registration/Controller/LayDeputyshipUploadController.php
```diff
@@ -17,9 +16,8 @@ class LayDeputyshipUploadController
 {
     public function __construct(
         private DataCompression $dataCompression,
-        private LayDeputyshipDtoCollectionAssemblerFactory $factory,
-        private LayDeputyshipUploader $uploader,
-        private LoggerInterface $verboseLogger
+        private CSVDeputyshipProcessing $csvProcessing,
+        private LoggerInterface $logger
```
Gugandeep, line 20: I think we have to keep the `$verboseLogger` variable to let the controller know to use our verbose logger service.

I think we have to use the verbose logger otherwise `notice()` won't work on prod like environments.

api/app/tests/Behat/features-v2/registration/ingest.lay.sirius.feature
```diff
@@ -1,42 +1,44 @@
-@v2 @v2_sequential_1 @registration @ingest
```
Gugandeep, line 1: I would strongly suggest not commenting out these behat tests as they are testing a lot of our business logic around registering and managing cases.

We should refactor the frontend forms or these tests so they align with our CSV upload process before merging in these changes if these tests are failing. 

I do agree with you that we shouldn't have forms or fields just for tests if we can avoid it.

client/app/src/Command/ProcessLayCSVCommand.php
```diff
@@ -54,32 +54,42 @@ protected function execute(InputInterface $input, OutputInterface $output): int
     {
         $bucket = $this->params->get('s3_sirius_bucket');
         $layReportFile = $this->params->get('lay_report_csv_filename');
+        $fileLocation = sprintf('/tmp/%s', $layReportFile);
 
         try {
             $this->s3->getObject([
```
Gugandeep, line 60: If we have the command on the api side now, is this command still required on the client side?

client/app/src/Command/ProcessOrgCSVCommand.php
```diff
@@ -65,32 +65,42 @@ protected function execute(InputInterface $input, OutputInterface $output): int
     {
```
Gugandeep, line 65: If we have the command on the api side now, is this command still required on the client side?

Gugandeep (commented):
Changes look good.

I would hesitate to merge this in with the behat tests commented out as they are vital for testing our business logic. I would recommend refactoring those before merging this in (or doing it as part of this ticket).

May be worth discussing the changes in this ticket in a dev call to understand the overall approach.

.github/workflows/scheduled-cycle-secrets.yml
```diff
@@ -2,8 +2,8 @@ name: "[Scheduled] Cycle Secrets"
 
 on:
   schedule:
-    # 2AM from Monday to Friday
-    - cron: "0 2 * * 1-5"
```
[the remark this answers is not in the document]
Raffers: Suggestion from @jamesrwarren to prevent any possible DB locks

api/app/src/Command/ProcessLayCSVCommand.php
```diff
@@ -0,0 +1,212 @@
...
+    }
+
+    private function process(mixed $data): bool 
+    {
+        $this->preReg->deleteAll();
+
+        if (is_array($data)) {
+            $chunks = array_chunk($data, self::CHUNK_SIZE);
+
+            $this->redis->set('lay-csv-processing', 'processing');
```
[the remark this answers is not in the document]
Raffers: Why hasn't this been refactored out on the FE command if we have this being done elsewhere?

client/app/src/Command/ProcessLayCSVCommand.php
Raffers: Yes, as it's part of the manual process. Which is one of the AC's for the ticket.

api/app/tests/Behat/features-v2/registration/ingest.lay.sirius.feature
Raffers: Will pick up the ticket I've created in the new year then. Off the back of this, as it's the first time I've really looked at our Behat tests. Is there a reason as to why the folder structure for these in buried into the api folder & not in it's own top level tests/behat folder? There was a moment of confusion when looking at this, with it touching the FE & API but is buried inside the API folder?

We shouldn't have them full stop. If we need to consider this then either our approach is wrong or how we are attempting to test is wrong. This should not have gone out once the process was refactored.

api/app/src/v2/Registration/Controller/LayDeputyshipUploadController.php
Raffers: That would be defined in the service.yml & not via a var name. There is no classes referenced in the system as `verbose logger`  The only thing remotely close to this, is `verbose_log_level` which is set to `notice` so long as we are sending our logs via the `notice` level then it's not an issue.

I have just double checked some of the changes I made after refactoring these classes. Seems I changed them from notice to info, so will revert those back.

Raffers:
> May be worth discussing the changes in this ticket in a dev call to understand the overall approach.

It mirror's the manual approach, except with migrating over the logic/commands to the API side of things while maintaining the manual approach. Both the manual & API command, carry out the exact same process.

PR #1521, opened 2024-02-02
DDLS-45 Automate CSV Upload

Raffers:
## Purpose
Change the trigger for the CSV job to a schedule to remove the need to manually click a button for it. 

We currently have a process whereby we manually run each CSV asynchronously by clicking buttons to run the CSV on Lay and the combined PA/professional CSV. The Lay CSV takes about ten minutes and the PA/Prof CSV somewhere between 45 minutes to an hour and we have to wait until the first has been completed to do the next one. 

While this is an improvement on the previous way of doing things, it would be beneficial to automate it entirely. 

-     To avoid the manual error of not running it and failing to make updates

-     Not having to plan cover on holidays and remembering to re-allocate the job on unexpected days off 

-     Avoiding the run being unsuccessful e.g. when we run it at the time as a deployment 

See previous comments - [https://github.com/ministryofjustice/opg-digideps/pull/1492](here)

Fixes DDLS-45 and DDLS-104

## Approach
Migrated most aspects of the front-end job over to the API, refactors some of the API classes so the process is shareable between the command & manual approach 

## Learning
We still lack loads of unit tests/require refactoring so they can be tested.

## Checklist

### Frontend


terraform/environment/region/cloudwatch_scheduled_events.tf
```diff
@@ -1,9 +1,83 @@
...
+
+  ecs_target {
+    task_count          = 1
+    task_definition_arn = aws_ecs_task_definition.api.arn
+    launch_type         = "FARGATE"
+    platform_version    = "1.4.0"
+
+    network_configuration {
+      security_groups  = [module.api_service_security_group.id]
+      subnets          = data.aws_subnet.private.*.id
```
jamesrwarren, line 59: write this as data.aws_subnet.private[*].id

api/app/src/Command/ProcessLayCSVCommand.php
```diff
@@ -0,0 +1,206 @@
...
+
+        $data = $this->csvToArray($fileLocation);
+        if (count($data) >= 1 && $this->process($data) && empty($this->processingOutput["errors"])) {
+            if (!unlink($fileLocation)) {
+                $logMessage = sprintf('Unable to delete file %s.', $fileLocation);
+
+                $this->logger->error($logMessage);
+                $this->cliOutput->writeln(
+                    sprintf(
+                        '%s - partial - %s Output: %s',
```
jamesrwarren, line 113: no such things as partial slack handler. it won't pick it up as a thing to alert on:

`pattern = r"^\w+(?:_\w+)* - (success|failure) - .+$"`

client/app/templates/Admin/Index/uploadOrgUsers.html.twig
```diff
@@ -14,30 +14,6 @@
 {% block pageContent %}
 
 <div class="govuk-grid-row">
-    <div class="govuk-grid-column-one-half hidden">
```
jamesrwarren, line 17: do we actually want to get rid of this? I thought we were keeping the manual process as a backup if overnight job failed

Raffers: This is the OLD process, if you look below the other form is still present, which is what the current manual process uses.

jamesrwarren (approved):

PR #1536, opened 2024-02-15
DDLS-45 - Automate CSV Process Investigation

Raffers:
## Purpose
Fixes various memory issues

Fixes DDLS-45

## Approach
Added various memory management workarounds. Some memory was being consumed by logging, others were because an object that would have gone out of scope if it was called via the api never gets reset. Also needed to boost the size of container and how much memory the process can use because of the size of the files being loaded into memory on the same container.

I have left some basic memory logging in place for the org process to diagnose any prod issue. Will remove in subsequent PR.

## Learning
NA

## Checklist

### Frontend

### DIFF
--- a/api/app/src/Command/ProcessLayCSVCommand.php
+++ b/api/app/src/Command/ProcessLayCSVCommand.php
@@ -65,13 +65,10 @@ class ProcessLayCSVCommand extends Command
         private readonly PreRegistrationRepository $preReg,
     ) {
         parent::__construct();
-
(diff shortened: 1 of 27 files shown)


jamesrwarren (approved):

api/app/src/Command/ProcessLayCSVCommand.php
iqpalm, line 1: Line 124 should it be "some"

jamesrwarren (approved):

PR #1557, opened 2024-03-08
DDLS-144 First name matching setup for preregistration

iqpalm:
## Purpose

Fixes DDLS-144

To be better able to identify which deputy has signed up to the service, it would be useful to make the first names mandatory as part of the registration process, and as a matching criteria. This is so that we can better match the details provided by the deputy with the contents of the Lay CSV.

## Approach
The approach taken is the same as used for Deputy Surname and first name was added to the relevant methods. The allows the Deputy Uid to be uniquely identified in case of matching deputy surnames.

Tests were updated to take into account first name and return an error if incorrect first name is provided. Occasions where a random first name was used was replaced with a method returning the first name.

The ticket was extended to cover the registration process of the co-deputy as well.

An error is also returned when the deputy cannot be uniquely identified using the deputy first name, surname and postcode and client surname.

The code also returns an error when a co-deputy tries to sign in with the same deputy uid and case number combination which has already been registered.

Additionally, this change facilitates the addition of a new column deputy_uid in the user table (dd_user). This column is being populated at the same time the deputy_no column is populated with the Deputy UID of the lay deputy. 

## Learning

## Checklist

### Frontend


Gugandeep (dismissed):
Looks good to go!

Gugandeep (dismissed):

PR #1563, opened 2024-03-20
DDLS-167 Add standards link to declaration section for ndr and reports

iqpalm:
## Purpose
Legal have told us we need to accommodate the new standards within the application and want the deputy to confirm in a declaration that these have been read. We already have a declaration in DigiDeps just ahead of submitting the report so it seems sensible to amend this to include the standards. The standard link will be different depending on the type of user i.e Lay, Public Authority or Professional.

Fixes DDLS-167

## Approach
The approach taken was to split the translation that is currently being used into 3 parts. The pre part would include the text up to the link. The link part would include the link and the post part would include the text after the link. The reason for this approach rather than combining the link into one translation was that the use of raw is not very secure and therefore splitting out the translation into 3 parts is more robust.

Also, the logic used to pick the standards link depending on the deputy type is included in the User class and is then used in the various twig templates.

## Learning
We tried combining the link into one translation and it worked but needed the use of raw in the twig template which was seen as a security issue.

## Checklist

### Frontend
- [X] Translations are used and the profiler doesn't identify any missing
- [X] Any links or buttons added are screen reader friendly and contextually complete


client/app/templates/Ndr/Formatted/_submission_declaration.html.twig
```diff
@@ -1,4 +1,11 @@
 {% set translationDomain = "ndr-declaration" %}
+{% if app.user.isLayDeputy %}
```
Raffers, line 2: Do we need these if statements in 4 different places? Is there no higher level twig file where these values can be set?

ndasmoj (approved):

PR #1570, opened 2024-04-11
DDLS-179 : New Court Order table - Entity classes and migration

ndasmoj:
## Purpose

Fixes DDLS-179

This change creates the client and api entity classes and the migration file for the new court order table.

## Approach

## Learning

## Checklist

### Frontend


Raffers (approved):

api/app/src/Entity/CourtOrder.php
```diff
@@ -0,0 +1,123 @@
...
+
+    public function getId(): int
+    {
+        return $this->id;
+    }
+
+    /**
+     * @return $this
+     */
+    public function setId(int $id)
```
Gugandeep, line 72: I think we don't usually create a setId method for these objects just to remove the chance of someone updating ids for things accidentally.

api/app/src/Entity/CourtOrder.php
```diff
@@ -0,0 +1,123 @@
...
+
+    public function getCourtOrderUid(): int
+    {
+        return $this->courtOrderUid;
+    }
+
+    /**
+     * @return $this
+     */
+    public function setCourtOrderUid(int $courtOrderUid)
```
Gugandeep, line 87: I think we can remove the annotation from the set methods and use the return types syntax like the get methods.

`public function setCourtOrderUid(int $courtOrderUid): CourtOrder`

api/app/src/Entity/CourtOrder.php
ndasmoj: @Gugandeep As discussed, lot of the entities do use it and its mainly for testing purposes. So retaining it.

Gugandeep (approved):

PR #1574, opened 2024-04-17
DDLS-173 Admin Registration Bugs

iqpalm:
## Purpose
1. When a Super Admin adds an OPG Staff as a new user, the system sends out an activation email to the new staff for completing the registration. On completing the registration successfully, the user is able to use the Admin service fine but the user is still marked as ‘Not Active’ and has the registration date set to ‘N/A’. This needs to be corrected wherein the user should be marked as ‘Active’ and the registration date should be set correctly.

2. When a Super Admin adds a PA Named Deputy or PROF Named Deputy as a new user, the concerned user receives an activation email. On clicking the activation email, the system results in an error. 

3. When a Super Admin adds an Admin Manager (OPG Staff) as a new user, the system sends out an activation email to the new staff for completing the registration. On clicking the activation email, the user is presented with tabs that are applicable for non OPG staff user registration. This needs to be corrected wherein the user sees the tabs relevant to OPG staff user registration. 

4. Public Authority and Professional Named Deputies currently have to agree 2 separate terms of use. One set before activating their account and the other after creating a password. Stacey confirmed that the terms of use before activating their account are no longer required and therefore can be deleted. 

5. Admin users active status is not set to true and a registration date is not set when they activate their account. The admin user should be set to true and a registration date set (if available) even if the login context is not available.

Fixes DDLS-173

## Approach
1. The reason for this was that the registration process is not completed and therefore the user is not marked as 'Active' and the registration date is not set. This is due to the login-context not being removed after the user confirms their details. Therefore the login context is now removed as part of the successful user registration.

2. When a Super Admin adds a PA Named Deputy or PROF Named Deputy as a new user, they need to agree terms of use before activating their account but currently this sets the token to null and therefore when registering there is no token that can be used. Therefore a new method was created that took this functionality out of the agreeTermsUse and is run at the end once the general terms of use are agreed by the user.

3. The Admin Manager role was missing for the if statement that decides what steps to show in the progress bar and would default to the steps when a deputy registers. The Admin role was added to the if statement to resolve this issue. Also as part of this the Admin Manager was added to the drop down menu on the homepage.

4. Deletion of the steps, templates and translations specific to pa and prof terms of use as no longer required. 

5. New endpoints were created for Admin users when the user sets a password to set the registration date and the active flag. An admin registration success event is also triggered at this point. Once the admin user logs in the active flag is set to true if false.

## Learning

## Checklist

### Frontend


ndasmoj:
@iqpalm The PR looks good. Just needs handling of the additional scenario of Admin registration using different sessions and not having the login-context across the sessions.

ndasmoj (approved):

ndasmoj (approved):

PR #1599, opened 2024-06-03
DDLS-217 Add date and route to User when pre registration details are validated

iqpalm:
## Purpose
There are a few ways to register to the service. Each of these processes will require the user to validate their details against the Pre-registration table, but at different points compare with each other.

To help us easier identify which process a user has used to register and if they have validated their details against the pre-registration table we will add 2 fields to the user table to populate this data.

Fixes DDLS-217

## Approach
1. Added two fields with pre registration validated date and registration route to dd user table
2. Added new fields to User class on api side
3. Added new fields to User class on client side
4. Set route and pre register validated date for self registration together with updated test
5. If the route is not self registering then the route will be set depending on the logged in user's role
6. Setting the pre register validated date at correct point for the remaining routes, co-deputy, admin and org admin.
7. Setup tests to test the correct status and date are set for each route.

## Learning

## Checklist

### Frontend


api/app/src/Service/UserService.php
```diff
@@ -39,6 +39,15 @@ public function addUser(User $loggedInUser, User $userToAdd, $data)
 
         $userToAdd->recreateRegistrationToken();
         $userToAdd->setCreatedBy($loggedInUser);
+
+        if ($loggedInUser->isLayDeputy()) {
```
Raffers, line 43: You can clean this up a bit more using a `match` - https://www.php.net/manual/en/control-structures.match.php

client/app/src/Entity/User.php
```diff
@@ -78,6 +78,12 @@ class User implements UserInterface, DeputyInterface, PasswordAuthenticatedUserI
     public const ACTIVATE_TOKEN_EXPIRE_HOURS = 48;
     public const PASSWORD_TOKEN_EXPIRE_HOURS = 1;
 
+    public const SELF_REGISTER = 'SELF_REGISTER';
+    public const ADMIN_INVITE = 'ADMIN_INVITE';
+    public const ORG_ADMIN_INVITE = 'ORG_ADMIN_INVITE';
+    public const CO_DEPUTY_INVITE = 'CO_DEPUTY_INVITE';
+    public const UNKNOWN_REGISTRATION_ROUTE = 'UNKNOWN';
```
Raffers, line 85: Are we going to populate all fields in the table? 
If not then why do we have an unknown value?
If we do plan to populate this field for all users, where is the migration script to update these?

client/app/tests/phpunit/Controller/CoDeputyControllerTest.php
```diff
@@ -0,0 +1,71 @@
+<?php
```
Raffers, line 1: Why are we testing an insertion into the DB, do we not trust the DB to do a correct insertion? 
Do we have a tests on the endpoint `/selfregister/verifycodeputy` that would sufficiently cover this.

client/app/tests/phpunit/Controller/UserControllerTest.php
```diff
@@ -0,0 +1,74 @@
+<?php
```
Raffers, line 1: The same comments have been made on `CoDeputyControllerTest`

iqpalm: There were no tests in place on the client side, so I am creating a couple of tickets to get them created/refactored to be proper unit tests.

client/app/tests/phpunit/Controller/CoDeputyControllerTest.php
iqpalm: There were no tests in place on either the client or API side, so I am creating a couple of tickets to get them created/refactored to be proper unit tests.

client/app/src/Entity/User.php
Raffers: Have just seen in the migrations file, that unknown will be the default. So this is required.

Raffers (approved):

PR #1611, opened 2024-06-18
DDLS-208: Update setting of Due Date for reports to account for Bank Holidays & Weekends

MiaGordon91:
## Purpose
Align the setting of the due date with Sirius by ensuring that we don't account for bank holidays (and weekends, which we were already avoiding).  

Fixes DDLS-208

## Approach
- Install Business Days mixin with Carbon to allow the use of the 'addBusinessDays' method which skips over bank holidays and weekends.
- Install PHP ext-calendar extension to allow the businessDays function to work and identify easter holidays. The extension is available locally but needed adding to Dockerfile to work in other environments.
- Wrap loading of Carbon package in service class
- Output GOV.UK bank holiday json file to tmp folder on api container and retrieve in it's own service class. This is required to account for adhoc days that the Business Days package doesn't account for. The dates are then passed in to the config step of the carbon/business days package.
- Abstract 'updateDueDateBasedOnEndDate' method into ReportService to enable injection and use of CarbonBusinessDaysService. Method is set to static and called in the Report Entity constructor.
- The CarbonBusinessDays->load() is also called in the ReportService constructor to make the 'addBusinessDays' method available to the static function.
- Explicitly call method in the ReportController to update due date on active reports.
- Update files where required to load carbon package in order to make the 'addBusinessDays' function available where a new report is instantiated in codebase.
- Unit tests added to test correct due dates are set.

## Learning

## Checklist

### Frontend


api/app/composer.json
```diff
@@ -23,6 +23,7 @@
     "require": {
         "php": "^8.1",
         "aws/aws-sdk-php": "3.299.1",
+        "cmixin/business-day": "^1.18",
```
Raffers, line 26: Why are we using an external library that doesn't pull from the GOV Json list? The referenced library doesn't include all bank holidays either, see - https://github.com/kylekatarnls/business-day/blob/dd2e46ec7e036ccd9b818c49183dcbf0b2a43f10/src/Cmixin/Holidays/gb-national.php

Ref - https://www.gov.uk/bank-holidays.json

Raffers (changes requested):
Why are we using an external library that doesn't pull from the GOV Json list? The referenced library doesn't include all bank holidays either, see - https://github.com/kylekatarnls/business-day/blob/dd2e46ec7e036ccd9b818c49183dcbf0b2a43f10/src/Cmixin/Holidays/gb-national.php

Ref - https://www.gov.uk/bank-holidays.json

Raffers:
Also, can you include a test that attempts to set a report, that would fall on a bankholiday. So we can confirm that it sets the next business day.

MiaGordon91:
This is the package that Sirius use so thought it would be best to use the same one to align our due dates with them. 
The 'gb-engwales' holiday is set, which merges 'gb-national' with additional bank holidays so that they're all accounted for.

Raffers (approved):

PR #1614, opened 2024-06-21
DDLS-180 - Populate Court Order table with active court order data from CSVs

Raffers:
## Purpose
Populate table with data from CSVs when they are ingested.

This would be just populating data into this table only, without creating any links between other tables/rows.

This would also ingest the new data into the pre-registration table and related fixtures.

 

### Acceptance Criteria:

- When the Lay and Org CSV is initially processed, reject the CSV if there is no CourtOrderUid column present.
- When processing a row from the Lay and Org CSV, reject a row if there isn’t a valid value in the CourtOrderUid column.
- When the Lay and Org CSV is ingested, if a new court order uid is present, create a new court order entry.
- When the Lay and Org CSV is ingested, compare our existing court order uids against the court order uids present in said CSVs:
    - If they are present, update the type if it has changed.
    - If they are not present, we should mark them as inactive.

Fixes DDLS-180

## Approach

## Learning

## Checklist

### Frontend


api/app/tests/Behat/features-v2/registration/ingest.lay.sirius.feature
```diff
@@ -1,4 +1,4 @@
-@v2 @v2_sequential_1 @registration @ingest
+@v2 @v2_sequential_1 @registration @ingest @chris
```
jamesrwarren, line 1: missed one ;)

PR #1615, opened 2024-06-24
DDLS-206 block malicious ips

jamesrwarren:
## Purpose
Stop bots/users from IP addresses doing malicious things to our app

Fixes DDLS-206

## Approach
The idea behind this one is to check the logs every 5 minutes and see if any IP address is getting lots of 404s. We have a bit of logic that checks that and also makes sure there's not legit activity for the IP address (in case it's a VPN IP that many ppl use).

We then have a dynamodb table that keeps a state of the addresses so we know how long to ban them for. First strike is half an hour, second is 4 hours and 3rd strike is 12 hours which coincides with the TTL of the dynamo record (time to live).

As such when their big ban is up then the record is removed from dynamo. This is so non static ips don't get perma banned! 

Every record in dynamo that has a expire time greater than current time gets the ban hammer and is added to the BlockedIps IP set which is used as a WAF rule. 

We have to put this through in 2 PRs as some of it lives in the account folder and some in the environment and environment needs account stuff to be there first. 

I also added a fix to the restore process that I noticed as it was restoring with the cross account kms key instead of the default one.

## Learning
Wanted to have decent unit tests for the logic on this one. I had spent quite a few hours failing to work with moto for the aws logs. It seems to have weird behaviour which doesn't pick up the logs properly when doing the insights query. I didn't want to have that issue again so used basic mocks for everything. This also took a long time. It's very slow and hard to work out the dynamo response stuff.. Hopefully I would find it easier next time. 

## Checklist

### Frontend


lambdas/functions/block_ips_lambda/app/block_ips.py
```diff
@@ -0,0 +1,239 @@
...
+      | filter request_uri != "/"
+      | sort @timestamp desc
+      | limit 10000""",
+    )
+
+    # Get the query ID for fetching results
+    query_id = response["queryId"]
+
+    # Wait for the query to complete
+    while True:
```
Raffers, line 42: Should we not include a time out to prevent it running forever?

jamesrwarren: Yeah good call. I'll add that :)

Raffers (approved):

PR #1624, opened 2024-07-11
DDLS-263 handle non existent report pdf documents

jamesrwarren:
## Purpose
Handle failures in htmltopdf better.

Fixes DDLS-263

## Approach
This PR does two things.
1) retries failed generation of htmltopdf docs 3 times. In all the scenarios we have see of this happening (which is not very many), There's nothing wrong with the html and a retry seems to work.

2) In the event that it still fails we now have the option as super admins to regenerate the file. It creates a new pdf in s3 and changes the reference to the current one to the new one.

To do this, we have created this button only visible for super admins on the ReportPdf:

image

When clicked we go to this screen:

image

## Learning
NA

## Checklist

### Frontend


client/app/src/Controller/Admin/ReportSubmissionController.php
```diff
@@ -80,9 +81,12 @@ public function indexAction(Request $request, ParameterStoreService $parameterSt
 
         $currentFilters = self::getFiltersFromRequest($request);
         $ret = $this->restClient->get('/report-submission?'.http_build_query($currentFilters), 'array');
-
         $records = $this->restClient->arrayToEntities(EntityDir\Report\ReportSubmission::class.'[]', $ret['records']);
 
+        foreach ($records as $reportSubmission) {
+            $report = $reportSubmission->getReport();
```
Gugandeep, line 87: Not sure if this change is needed, can't see any use of the $report variable here

jamesrwarren: Ha, yes thank you. I had that in from before for some debugging and removed the printy bit but not that! Removed that block now. Cheers

PR #1638, opened 2024-07-26
DDLS-261: Deleting a user with user research response data gives an error

MiaGordon91:
## Purpose
If a user has been discharged from a client but if the user has previously submitted a satisfaction score and then a user research form (which appears after the satisfaction survey form), then the system throws an error while deleting the user.

Fixes DDLS-261

## Approach
1. Change the user id to null for any rows in the user research table that equals the user id that is to be deleted.
2. Delete these rows in the user response table that have the user id set to NULL, setup as daily job.
3. This does not affect the satisfaction table and therefore the satisfaction scores are unchanged.

## Learning

## Checklist

### Frontend

### DIFF
--- a/api/app/src/Controller/UserResearchController.php
+++ b/api/app/src/Controller/UserResearchController.php
@@ -92,14 +92,4 @@ class UserResearchController extends RestController
     {
         $this->userResearchResponseRepository->findByUserId($id);
     }
-
-    /**
-     * @Route("/user-research/delete", name="delete_user_research", methods={"DELETE"})
-     *
-     * @Security("is_granted('ROLE_SUPER_ADMIN')")
-     */
-    public function deleteNullifiedUserId(): void
-    {
-        $this->userResearchResponseRepository->deleteByNullUserId();
-    }
 }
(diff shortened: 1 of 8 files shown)


api/app/src/Controller/UserResearchController.php
```diff
@@ -82,4 +82,24 @@ public function getAll(Request $request)
...
+    {
+        $this->userResearchResponseRepository->findByUserId($id);
+    }
+
+    /**
+     * @Route("/user-research/delete", name="delete_user_research", methods={"DELETE"})
+     *
+     * @Security("is_granted('ROLE_SUPER_ADMIN')")
+     */
+    public function deleteNullifiedUserId(): void
```
Raffers, line 101: We need to some functional tests around this endpoint. Every release we do, should be confirming that only a super admin can ever access this. We do have controller tests on the user endpoints that also have a deletion action,

I'd go as far as suggesting this is moved to a daily job rather than having an endpoint, as once the FK has been removed. Exactly when the entry is deleted isn't a point of concern long as we do delete the entry.

PR #1640, opened 2024-07-30
DDLS-216 - Expand the Lay Deputy User cleanup job to include incomplete user records

Raffers:
## Purpose
There is an existing Lay Deputy User cleanup Command job in the system 'digideps:delete-zero-activity-users' which deletes lay user(s) that have never had any activity after 30 days of registration. This job can be expanded to include user(s) that have not attempted or completed the registration after a certain period since they were sent a user activation link either through self-registration, admin invite or co-deputy invite. 

The cut off dates for users that have NOT set a password and users that have set a password will be different, so both these scenarios would need to be dealt with separate logic.

These users should not have a registration date but they will have a token_date and a last_logged_in date if they have set the password and then left the process. The criteria to consider the user(s) would be based on either the token_date or last_logged_in date or both. 

There may be cases where active users might not have a registration date (to be confirmed), and so we will need to accommodate for these users to ensure they are not deleted. One check we can do is to make sure they are not associated with a client and do checks on the report (submitted dates etc.)

Fixes DDLS-216

## Approach

## Learning

## Checklist

### Frontend


api/app/src/Service/UserRegistrationService.php
```diff
@@ -146,12 +146,8 @@ public function saveUserAndClient(User $user, Client $client)
             $this->em->persist($user);
             $this->em->flush();
 
-            // Add the user to the client an save it
-            /* @var Client $client */
+            // Add the user to the client then save
             $client->addUser($user);
-            $user->setCreatedBy($user);
```
Raffers, line 152: This was initially added due to wanting to track the route a user had taken. We now have something better in place, as leaving this here for self-registered users would leave them out of scope for the cleanup job I've removed this. Will adjust data so all self referencing created_by are removed.

api/app/src/Migrations/Version282.php
```diff
@@ -0,0 +1,35 @@
...
+    {
+        return 'Adds index for created_by_id column on dd_user';
+    }
+
+    public function up(Schema $schema): void
+    {
+        // this up() migration is auto-generated, please modify it to your needs
+        $this->addSql('ALTER TABLE dd_user DROP CONSTRAINT dd_user_created_by_fk');
+        $this->addSql('ALTER TABLE dd_user ADD CONSTRAINT FK_6764AB8BB03A8386 FOREIGN KEY (created_by_id) REFERENCES dd_user (id) NOT DEFERRABLE INITIALLY IMMEDIATE');
+        $this->addSql('CREATE UNIQUE INDEX UNIQ_6764AB8BB03A8386 ON dd_user (created_by_id)');
```
Gugandeep, line 25: If I'm understanding this correctly, I'm not sure we can do this because an admin user can and should be able to invite multiple users, so the value in the `created_by_id` column won't be unique.

Gugandeep (commented):
Left a few comments.

Is it possible to add a test that checks that only the expected inactive users are returned? I think the best way to do this would be to test the `findInactive()` method in the `UserRepository.php`.

api/app/src/Entity/User.php
```diff
@@ -384,24 +334,18 @@ class User implements UserInterface, PasswordAuthenticatedUserInterface
...
      * @JMS\Groups({"user", "satisfaction", "user-research"})
-     *
      * @ORM\OneToMany(targetEntity="App\Entity\UserResearch\UserResearchResponse", mappedBy="user", cascade={"persist"})
      */
     private $userResearchResponse;
 
     /**
      * @var User|null
      *
-     * @ORM\OneToOne(targetEntity="App\Entity\User")
```
Raffers, line 397: This was the wrong association type, One-to-many self-referencing - https://www.doctrine-project.org/projects/doctrine-orm/en/3.2/reference/association-mapping.html#one-to-many-self-referencing

Raffers:
> Left a few comments.
> 
> Is it possible to add a test that checks that only the expected inactive users are returned? I think the best way to do this would be to test the `findInactive()` method in the `UserRepository.php`.

No, due to the states that accounts need to be in. It would require much more work to allow the fixtures to create the required states. You can manually test this if you wish

MiaGordon91 (approved):

PR #1669, opened 2024-09-12
DDLS-272

Gugandeep:
## Purpose
This change covers 3 tickets that will introduce a new homepage for deputies with multi clients. In the event that a deputy has multi clients they will be redirected to the choose a client homepage rather than the Lay homepage.

Tickets DDLS-272 and DDLS-318 have been created for the backend and frontend changes. 

DDLS-274 is a ticket to ensure that if the deputy inputs an invalid report id and they have multiple clients then they are redirected to the choose a client homepage rather than the Lay homepage. 

Fixes DDLS-272 DDLS-318 DDLS-274

## Approach
DDLS-272

DDLS-318 and DDLS-274
I created a new SQL query to return all the clients based on the deputy uid (ClientRepository). 
A test was also created that had 2 user accounts, each with an active client. Another client was attached to user 2. The same deputy uid was set for both user accounts. The assertion was to return 3 clients, 1 from user 1 and 2 clients from user 2.

The check to make sure the primary account has been used to login was added to the choose a client endpoint. Previously, it was only present in the Lay homepage endpoint. This check is required regardless if the deputy has 1 client or multiple clients.

The redirection is based on Role and there were 2 places in the Redirector class where you would be redirected to the Lay homepage. These were the first page once logged in successfully and the home redirect. I created a new private function getCorrectLayHomepage that would check the number of clients attached to the deputy uid (using above SQL query) and then go to the correct homepage. Currently, this is behind the feature flag in this function.

The behat test for a deputy with multiple clients signing in using their primary user account was updated to land on the choose a client homepage rather than the Lay homepage.

## Learning

## Checklist

### Frontend


client/app/src/Controller/Report/ReportController.php
```diff
@@ -170,6 +170,58 @@ public function indexAction(Redirector $redirector, ParameterStoreService $param
...
+        $user = $this->userApi->getUserWithData(['user-clients', 'client']);
+        $isMultiClientFeatureEnabled = $parameterStoreService->getFeatureFlag(ParameterStoreService::FLAG_MULTI_ACCOUNTS);
+        file_put_contents('php://stderr', print_r('**** Redirected in chooseAClient **** '.$isMultiClientFeatureEnabled, true));
+
+        if ('1' == $isMultiClientFeatureEnabled) {
+            // redirect back to log out page if signing in with non-primary account with primary email
+            if (!$user->getIsPrimary()) {
+                $primaryEmail = $this->userApi->returnPrimaryEmail($user->getDeputyUid());
+
+                $this->addFlash('nonPrimaryRedirect',
```
Gugandeep, line 195: Do we need to add something to check if there is an actual primary address? What happens should be show if there isn't a primary email?

Gugandeep: This also returns an array rather than an email by the looks of it.

client/app/src/Controller/Report/ReportController.php
MiaGordon91, line 1: Once a client is selected it only displays the first client in the header of the lay home page but understand that this will get updated in another ticket.

MiaGordon91 (approved):

PR #1670, opened 2024-09-16
Ddls 302

jamesrwarren:
WORK IN PROGRESS

## Purpose

Fixes DDLS-####

## Approach

## Learning

## Checklist

### Frontend


api/app/scripts/reset_db_fixtures.sh
```diff
@@ -1,6 +1,15 @@
...
 
-# We need below to create the params file on container start
+environment=${1:-development}
+
 confd -onetime -backend env
 php app/console doctrine:fixtures:load --no-interaction
+
+# Only run the test fixtures load if the environment is 'local'
+if [ "$environment" == "local" ]; then
+  php app/console doctrine:fixtures:load --no-interaction --env=test
```
Gugandeep, line 11: These fixtures are used by users on the training environment. Will this affect those users?

terraform/environment/region/ecs_task_reset_database.tf
```diff
@@ -18,7 +18,7 @@ locals {
     {
       name    = "reset-database",
       image   = local.images.api,
-      command = ["sh", "tests/Behat/reset-db.sh"],
+      command = ["sh", "scripts/reset_db_fixtures.sh"],
```
Gugandeep, line 21: Does the `local` argument need to be added for this to actually reset the db now?

terraform/environment/region/lambda_synchronisation.tf
```diff
@@ -1,88 +1,86 @@
-locals {
```
Gugandeep, line 1: Are we re-using a lambda for the audit sql stuff?
Is any of this sync functionality still needed?

jamesrwarren: Nope none of it. I was going to pull it out properly as a separate PR as there's quite a bit of it lying around. Just commented it out here as some of my changes would have involved updating it to get it working as as it's not used, it wasn't worth the effort.

api/app/scripts/reset_db_fixtures.sh
jamesrwarren: Nah. Training environment was just having the usual: php app/console doctrine:fixtures:load --no-interaction previously and this will do the same.

The reason we do additional reset with different flag in local is because the local tests use a different database as defined in the .yml files so we have to reset both.

PR #1685, opened 2024-10-07
DDLS-328b: Update relevant urls to include client id

Gugandeep:
## Purpose

Fixes DDLS-328

## Approach

- Deprecated old endpoints
    - Changed route names
    - Kept original route paths
    - Redirected to default homepage (Redirector.php will redirect as necessary)

- Created new endpoints
    - Used previous route names (other parts of the app redirect to endpoints using these route names so prevents having to update those)
    - Used new route paths which includes the client id
    - Instead of retrieving the first client, now retrieves client using id.
    - Updated a bunch of controllers & templates to pass through the required client id

- Updated default lay user fixtures
    - Lay deputies now have unique deputy uids
    - Created a new set of users to replicate multi-client deputies
    - Created a new set of users to replicate duplicate client deputies (deputies who have accidently registered for the same client multiple times)

- Updated ACL rules for various entities
    - Allowed users to access certain entities if they have a deputy uid that matches the user associated with the original entities.

- Updated ACL api unit tests
    - Updated certain fixtures to be given a deputy uid
    - Added additional fixtures to test ACL exception based on deputy uid works as expected.

- Refactored and update multi-client behat tests
    - Moved logic to create users and clients from a step in the test to tags and fixtures
    - Added steps to test to check user can choose both clients and view their reports.

- Updated smoke tests
    - Updated expect path for lay login to use new route path

## Learning
I tried moving the new route for the lay homepage `clientHomepageAction()` to the `ClientController.php` however, was receiving some weird errors. The first was that when I tried going to the add client page (`/client/add`), it would get stuck in a redirect loop. The add client page would immediately send me to the new lay homepage (`/client/{id}`) without triggering any of the code on the add client page (none of my debug log statements were triggered at all). The new lay homepage would then check to see if the user had a valid client (which it didn't because it wasn't added), and then redirect back to the add client page.

I tried quite a few things like renaming the routes, looking for any custom route config, and even a total rebuild of my local environment, but nothing seemed to work. I decided to try moving the route back to the ReportController and it worked immediately. Didn't fully understand why it broke, and why moving it backed made it work but decided at that point it was enough time wasted so continued with the rest of the ticket.

## Checklist

### Frontend

### DIFF
--- a/orchestration/anonymisation/main.go
+++ b/orchestration/anonymisation/main.go
@@ -94,7 +94,7 @@ func main() {
 	common.CheckError(err)
 
 	// ===== Processing - Additional Complex Updates =====
-	err = processing.CustomSQLScriptUpdates(db)
+	err = processing.CustomSQLScriptUpdates(db, fmt.Sprintf("%sprocessing/sql", path))
 	common.CheckError(err)
 
 	// ===== Processing - Additional Complex Updates =====
--- a/orchestration/anonymisation/processing/customSQLScripts.go
+++ b/orchestration/anonymisation/processing/customSQLScripts.go
@@ -9,10 +9,7 @@ import (
 )
 
 // CustomSQLScriptUpdates reads SQL files from the 'sql' directory and executes them in order.
-func CustomSQLScriptUpdates(db *sql.DB) error {
-	// Directory containing the SQL files
-	sqlDir := "processing/sql"
-
+func CustomSQLScriptUpdates(db *sql.DB, sqlDir string) error {
 	files, err := os.ReadDir(sqlDir)
 	if err != nil {
 		return fmt.Errorf("failed to read directory: %v", err)
@@ -50,7 +47,7 @@ func UpdateAllToPassedInValue(
 	updateString string) error {
 	query := fmt.Sprintf(`UPDATE anon.%s SET %s = '%s';`, tableName, updateColumn, updateString)
 
-	fmt.Println(query)
+	fmt.Printf(`Ran blanket update on %s for column %s`, tableName, updateColumn)
 	// Execute the update query
 	_, err := db.Exec(query)
 	if err != nil {


api/app/src/Repository/UserRepository.php
```diff
@@ -411,4 +409,14 @@ public function findByFiltersWithCounts(
             'count' => $count,
         ];
     }
+
+    public function findDeputyUidsForClient(int $clientId)
+    {
+        $query = $this
+            ->getEntityManager()
+            ->createQuery("SELECT u.deputyUid FROM App\Entity\User u LEFT JOIN u.clients c where c.id = ?1 AND u.deputyUid IS NOT NULL ORDER BY u.id")
```
Gugandeep, line 417: Make inner join

api/app/tests/Unit/Controller/ClientControllerTest.php
```diff
@@ -214,11 +233,23 @@ public function testfindByIdAuth()
...
+        $url = '/client/'.self::$primaryAccountClient->getId();
+        $url2 = '/client/'.self::$nonPrimaryAccountClient->getId();
+
+        $this->assertEndpointAllowedFor('GET', $url, self::$tokenMultiClientPrimaryDeputy);
+        $this->assertEndpointAllowedFor('GET', $url2, self::$tokenMultiClientPrimaryDeputy);
 
-        $this->assertEndpointNotAllowedFor('GET', $url2, self::$tokenDeputy);
+        $this->assertEndpointAllowedFor('GET', $url, self::$tokenMultiClientNonPrimaryDeputy);
+        $this->assertEndpointAllowedFor('GET', $url, self::$tokenMultiClientNonPrimaryDeputy);
     }
```
Gugandeep, line 253: Add test to check a co-deputy not associated with the client can't access the client

Gugandeep: Also add a test for prof/pas (pa's accessing other pa clients, pa's accessing lay clients)

api/app/tests/Behat/bootstrap/v2/Common/AuthTrait.php
```diff
@@ -407,4 +410,37 @@ public function whenTheyLogOutTheyShouldnTSeeANonPrimaryFlashMessage()
...
+     * @When /^they choose their "(primary|non-primary)" Client$/
+     */
+    public function theyChooseTheirFirstClient($isPrimary)
+    {
+        $clientId = 'primary' == $isPrimary ? $this->layPfaHighNotStartedMultiClientDeputyPrimaryUser->getClientId()
+            : $this->layPfaHighNotStartedMultiClientDeputyNonPrimaryUser->getClientId();
+
+        $urlRegex = sprintf('/client\/%d$/', $clientId);
+
+        $this->iClickOnNthElementBasedOnRegex($urlRegex, 0);
```
Gugandeep, line 424: Update index based off of primary/non-primary

jamesrwarren (approved):

PR #1734, opened 2024-11-13
DDLS-387 Pull the NDR report through to the reports page

iqpalm:
## Purpose
As a multi-client deputy with a current NDR, I want to be able to access the NDR from the reports page, so that I can complete or review NDRs

When we implemented the multi-client pathway, we removed the NDR page for these users. Some users in future may need to be able to access and complete NDRs.

Fixes DDLS-387

## Approach
The approach taken has been to implement the change for all Lay users, single and multi clients. This means the "/ndr" endpoint has been deprecated.

1. I deleted the if statement that checks for isNdrEnabled as it is no longer required in getCorrectRouteIfDifferent in Redirector. The lay_home will be accessed whether isNdrEnabled is true or false. The report_create in getLayDeputyHomepage is only required if the NDR has not been enabled else it will show the NDR on the report page. The RedirectorTest file has been amended to delete the tests for ndr_index as the Redirector no longer checks for these scenarios.

2. The functionality for ndr_index has been moved into lay_home page. A check is made against all users attached to a client to find out if isNdrEnabled is set to true (i.e the checkbox has been ticked in admin app). Depending on this a flag is set (ndrEnabled) that determines if a NDR or a report will be shown.

    The twig template includes a reference to the NDR index twig file if ndrEnabled is set to true.

    The ndr fixture createLayNdrNotStarted in FixtureHelper required a change where isPrimary is now set to true. This was 
    required as it was logging out due to the isPrimary check in lay_home endpoint.

3. During the NDR submission journey the app was failing after the post user research step as the client id was required from the NDR by the lay_home endpoint. This was solved by including the JMS groups ['ndr-client', 'client-id'] to ensure the app returns to the correct client.

4. Previously, the pattern was to bring back the client and then access the NDR. However, this has been amended to use the ndrId from the route to access the NDR and then bring back the client. This is cleaner code especially for multi client deputies.

5. Deprecate ndr_index endpoint and redirect to hompage. Amend ndr_index references to lay_home and also include clientId.

6. Create behat tests to test NDR enabled checkbox results in correct report on reports page. I tried to do in 1 test but failed in pipeline so decided to do in 2 seperate tests.

7. Included parameters to breadcrumb for ndr review which was showing as deputyFirstname deputyLastname

8. Added in functionality for multi client breadcrumbs for NDR journey (similar to report journey)

## Learning

## Checklist

### Frontend


client/app/src/Controller/Ndr/ActionController.php
```diff
@@ -36,39 +37,50 @@ class ActionController extends AbstractController
...
+     * @var ClientApi
+     */
+    private $clientApi;
+
     public function __construct(
         ReportApi $reportApi,
         RestClient $restClient,
-        StepRedirector $stepRedirector
+        StepRedirector $stepRedirector,
+        ClientApi $clientApi
```
Raffers, line 49: It doesn't really matter for this, but if you do come across/need to add to the constructor, refactor it & use [constructor promotion](https://www.php.net/manual/en/language.oop5.decon.php#language.oop5.decon.constructor.promotion)

client/app/src/Controller/Ndr/IncomeBenefitController.php
```diff
@@ -38,39 +39,47 @@ class IncomeBenefitController extends AbstractController
...
-     * @param $ndrId
+     * @Template("@App/Ndr/IncomeBenefit/start.html.twig")
      *
      * @return array|RedirectResponse
      */
     public function startAction($ndrId)
     {
+        /** @var User $user */
+        $user = $this->getUser();
+        $isMultiClientDeputy = 'ROLE_LAY_DEPUTY' == $user->getRoleName() ? $this->clientApi->checkDeputyHasMultiClients($user->getDeputyUid()) : null;
```
Raffers, line 67: I thought there might have been a simpler way for this, but it would require refactoring `checkDeputyHasMultiClients` as this function is only for lay's we should have done the checking if they are a lay inside the function rather than have a ternary check. 

With the above in mind, I will suggest changing the null to a false just for type consistency.

client/app/src/Controller/Ndr/NdrController.php
```diff
@@ -106,7 +79,7 @@ public function indexAction(Redirector $redirector)
      *
      * @return array|RedirectResponse
      */
-    public function overviewAction(Redirector $redirector, ParameterStoreService $parameterStore)
+    public function overviewAction(Redirector $redirector, $ndrId)
```
Raffers, line 82: Missing type information on `$ndrId`

Raffers (changes requested):

Raffers (approved):

PR #1738, opened 2024-11-21
DDLS-406: Allow a multi-client deputy to send a co-deputy invitation from a client linked to their secondary account

MiaGordon91:
## Purpose
A bug was identified whilst investigating another bug ticket (DDLS-396) where a multi-client deputy cannot send a co-deputy invitation from any client attached to their secondary accounts - they can only successfully invite a co-deputy from the client linked to their primary account. 

Reason being, is because the deputy account that’s sending the invite is not attached to that particular client on a database level. 

Fixes DDLS-406

## Approach
- Update logic to use the clientId from the URL as part of the co-deputy invitation process
- Move function from the OrgService to the UserService as it did not apply to organisations
- Add a unit test for the updated endpoint in the CoDeputyController
- Add a behat test to cover the user journey

- Added in condition in Redirector that co-deputy verification only required when the user has been invited through the co-deputy invite route. The reason this was required is as follows:

 1. Multi-client deputy sends co-deputy invitation from any client attached to their secondary account
 2. Co-deputy then signs up
 3. Multi-client deputy can sign in and see the co-deputy is attached to the client
 4. However, the secondary account was not logging out and showing the login with your primary account error message. It was redirecting to the co-deputy verification page. I think this is to do with the user->getCoDeputyClientConfirmed() which would not be set for the secondary account as the invitation came from the primary account.

## Learning

## Checklist

### Frontend


api/app/src/Controller/CoDeputyController.php
```diff
@@ -43,10 +44,11 @@ public function countMld(Request $request)
     }
 
     /**
-     * @Route("add", methods={"POST"})
+     * @Route("add/{clientId}", methods={"POST"})
+     *
      * @Security("is_granted('ROLE_DEPUTY')")
      */
-    public function add(Request $request)
+    public function add(Request $request, $clientId)
```
Raffers, line 51: Missing type information from `$clientId`

api/app/src/Service/OrgService.php
```diff
@@ -47,13 +41,11 @@ class OrgService
     public function __construct(
         EntityManagerInterface $em,
         UserRepository $userRepository,
-        ClientRepository $clientRepository,
```
Raffers, line 50: You've gone ahead & used constructor promotion elsewhere in this PR, small little refactor on this section as well?

Raffers (changes requested):

Raffers (approved):

PR #1760, opened 2024-12-09
DDLS-371: Show primary email in admin for client assigned to non primary

Gugandeep:
## Purpose
When an admin user views the details of a client, it also shows the details for the user account associated with that client.
This change will now show the primary user account details if the original associated user account was a non-primary account.

Fixes DDLS-371

## Approach

- Change to frontend to check if the user associated with client is a primary account
- Call to new API endpoint to retrieve the primary account for a deputy uid
- Updated transformer and dto in API to bring back deputy_uid and is_primary
- Added api unit test
- Added behat test

## Learning

## Checklist

### Frontend


iqpalm (approved):

api/app/src/Controller/UserController.php
```diff
@@ -651,4 +651,25 @@ public function getPrimaryEmail(int $deputyUid): string
 
         return $userEmail;
     }
+
+    /**
+     * Endpoint for getting the primary user account associated with a deputy uid.
+     *
```
Raffers, line 657: As this is an admin action do we not want to limit this endpoint using a `@security` tag?

Gugandeep: So on the API side we don't tend to add the `@security` tag. The reason for this is that the requests made to these endpoints on the API side could be coming from a deputy request or an admin request (the API endpoints don't care).
However, they do have the some level of security checks and that's to do with the key that is passed in with the API request. It does this to make sure that the request is legitimate and has come from the frontend, and not from anywhere else.

If you look at `client/app/src/Controller/Admin/Client/ClientController.php` you'll notice that there is a `@security` tag on that `detailsAction` endpoint.

Raffers (approved):

PR #1775, opened 2024-12-17
DDLS-402 - Migrate existing data from dd_user table to deputy table

Raffers:
## Purpose
Allows us to separate account information from deputy entity and move one step closer to aligning Lay deputy data to Prof and PA.

Fixes DDLS-402

## Approach
Adds migration script to add missing deputies to deputies table from user table. 

## Learning

## Checklist

### Frontend

### DIFF
--- a/api/app/src/Migrations/Version286.php
+++ b/api/app/src/Migrations/Version286.php
@@ -19,7 +19,7 @@ final class Version286 extends AbstractMigration
 
     public function up(Schema $schema): void
     {
-        if (getenv('RUN_ONE_OFF_MIGRATIONS')) {
+        if (!!getenv('RUN_ONE_OFF_MIGRATIONS')) {
             $sql = <<<SQL
             INSERT INTO deputy 
             (user_id, deputy_uid, firstname, lastname, email1, address1, address2, 
--- a/api/app/src/Migrations/Version286.php
+++ b/api/app/src/Migrations/Version286.php
@@ -10,7 +10,7 @@ use Doctrine\Migrations\AbstractMigration;
 /**
  * Auto-generated Migration: Please modify to your needs!
  */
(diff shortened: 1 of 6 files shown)


api/app/src/Migrations/Version286.php
```diff
@@ -0,0 +1,50 @@
...
+final class Version286 extends AbstractMigration
+{
+    public function getDescription(): string
+    {
+        return 'Adds mirgration script';
+    }
+
+    public function up(Schema $schema): void
+    {
+        if (getenv('RUN_ONE_OFF_MIGRATIONS')) {
```
jamesrwarren, line 22: I'd check this is doing what you expect it to be doing locally by trying it as false. I have a feeling the will resolve to true even if it's set as false because it's checking on the existence of the var and the var comes in as a string. 

You might have to convert to boolean on this line. Not 100% certain but I have had issues with this sort of thing in the past

terraform/environment/region/ecs_cluster.tf
```diff
@@ -120,6 +120,10 @@ locals {
       name  = "LAY_REPORT_CSV_FILENAME",
       value = local.lay_report_csv_file
     },
+    {
+      name  = "RUN_ONE_OFF_MIGRATIONS",
+      value = "true"
```
jamesrwarren, line 125: So there's a bit more to do here so we can make it environment specific so we can release to pre and test then update the var for prod.

What you need to do is add the variable run_one_off_migrations into the environment folders terraform.tfvars.json for every environment block in there (follow example of the other vars in there). Next add it to the mapping which is in variables.tf.

Finally update the value on line 125 to be var.account.run_one_off_migrations.

This way we can stagger out the changes and test it in pre without having to intervent with manual terraform changes etc..

PR #1797, opened 2025-01-21
DDLS-451: Associate deputies with court orders (database changes)

MiaGordon91:
## Purpose
As part of the database change work, this PR is to add a link table between court orders and deputies so that one deputy can have many court orders, and one court order can have many deputies.

Fixes DDLS-451

## Approach
- Create CourtOrderDeputy entity class (which acts as the link table) 
- Getter method added to Deputy entity to retrieve court order objects and discharged property
- Define ManytoOne relationship with CourtOrder and Deputy entities
- Generate Doctrine migration 
- Add unit and integration test 

Note: associateWithCourtOrder method has been added to the Deputy entity class temporarily to show how the association would work between the link table and associated entities, this logic would eventually move into a factory class. The integration test would also get replaced with a more suitable test as it's currently in place to test how the new link table is populated. 

## Learning

## Checklist

### Frontend

### DIFF
--- a/api/app/src/Entity/CourtOrderDeputy.php
+++ b/api/app/src/Entity/CourtOrderDeputy.php
@@ -1,0 +1,62 @@
+<?php
+
+namespace App\Entity;
+
+use Doctrine\ORM\Mapping as ORM;
+
+/**
+ * Link table between court orders and deputies.
+ *
+ * @ORM\Table(name="court_order_deputy")
+ *
+ * @ORM\Entity()
+ *
+ * @ORM\HasLifecycleCallbacks()
+ */
+class CourtOrderDeputy
+{
+    /**
+     * @ORM\Id
+     *
+     * @ORM\ManyToOne(targetEntity="App\Entity\CourtOrder")
+     *
+     * @ORM\JoinColumn(name="court_order_id", referencedColumnName="id", nullable=false)
+     */
+    private $courtOrder;
+
+    /**
+     * @ORM\Id
+     *
+     * @ORM\ManyToOne(targetEntity="App\Entity\Deputy")
+     *
+     * @ORM\JoinColumn(name="deputy_id", referencedColumnName="id", nullable=false)
+     */
+    private $deputy;
+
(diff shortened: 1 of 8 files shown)


api/app/src/Entity/CourtOrderDeputy.php
```diff
@@ -0,0 +1,76 @@
...
+class CourtOrderDeputy
+{
+    /**
+     * @ORM\Id
+     *
+     * @ORM\ManyToOne(targetEntity="App\Entity\CourtOrder", inversedBy="courtOrderDeputyRelationships", cascade={"persist"})
+     *
+     * @ORM\JoinColumn(name="court_order_id", referencedColumnName="id")
+     */
+    private CourtOrder $courtOrder;
```
jamesrwarren, line 25: Think the 2 ids should be non nullable as they would always both be populated if the row exists right?

jamesrwarren: that is if nullable if false by default which i think it is

api/app/src/Entity/CourtOrderDeputy.php
```diff
@@ -0,0 +1,76 @@
...
+     * @ORM\JoinColumn(name="deputy_id", referencedColumnName="id")
+     */
+    private Deputy $deputy;
+
+    /**
+     * @ORM\Column(name="discharged", type="boolean", nullable=false)
+     */
+    private bool $discharged;
+
+    public function getDeputy(): Deputy
```
jamesrwarren, line 41: is there a way to set up false as the default for this? maybe in constructor or here. Not sure where :)

townxelliot:
@jamesrwarren: @MiaGordon91 is addressing your comments

api/app/src/Entity/CourtOrderDeputy.php
MiaGordon91: Good point, I've updated the entity class

jamesrwarren (approved):

PR #1807, opened 2025-02-03
DDLS-467 upgrade redis - part 1

jamesrwarren:
## Purpose
Upgrade redis to 7.1

Fixes DDLS-467

## Approach
Doing straight upgrade takes ages.. Would be better to create another instance and change across where the env vars are pointing to avoid downtime and disruption.

image

Not sure how long it would have taken as I killed it at this point after 16 mins..

As such approach is to make another one and then switch it over which I've tried in this env and it seems to work fine and would be a few seconds downtime instead of however long the upgrade would have taken. This is also easier as we use shared redis so we can update it one for each env and then switch them over in next PR. So basically this will be 3 little PRs. This one, one to update to new redis and finally a cleanup one.

## Learning
NA

## Checklist

### Frontend


townxelliot (approved):
Looks good, but would it make sense to do the switchover out of hours? I'm assuming the keys from the old Redis won't be migrated to the new one?

PR #1819, opened 2025-02-18
DDLS-485 check submission valid for NDR before submitting

jamesrwarren:
## Purpose
When important data for the submission is missing for whatever reason, create exception before creating the report pdf and doing other updates

Fixes DDLS-485

## Approach
This is the first in a number of submission related bug fixes. This does a check early on for casenumber and court_date and if they're missing it raises an exception that breaks the request early. I'm also logging instances of missing address and postcode as although this would not break the submit (and hence I don't raise exception), it would be useful to see how often we're missing the address and postcode.

## Learning
For the unit tests I'm using createMock as the prophesise was giving deprecation warnings. 

## Checklist

### Frontend

### DIFF
--- a/client/app/src/Controller/Ndr/NdrController.php
+++ b/client/app/src/Controller/Ndr/NdrController.php
@@ -17,6 +17,7 @@ use App\Service\File\S3FileUploader;
 use App\Service\HtmlToPdfGenerator;
 use App\Service\NdrStatusService;
 use App\Service\Redirector;
+use Psr\Log\LoggerInterface;
 use Sensio\Bundle\FrameworkExtraBundle\Configuration\Template;
 use Symfony\Component\HttpFoundation\RedirectResponse;
 use Symfony\Component\HttpFoundation\Request;
@@ -55,7 +56,8 @@ class NdrController extends AbstractController
         private ClientApi $clientApi,
         private SatisfactionApi $satisfactionApi,
         private NdrApi $ndrApi,
-        private HtmlToPdfGenerator $htmlToPdf
+        private HtmlToPdfGenerator $htmlToPdf,
+        private LoggerInterface $logger
     ) {
     }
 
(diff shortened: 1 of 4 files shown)


client/app/src/Controller/Ndr/NdrController.php
```diff
@@ -229,6 +231,17 @@ public function declarationAction(Request $request, int $ndrId, S3FileUploader $
...
         $form->handleRequest($request);
         if ($form->isSubmitted() && $form->isValid()) {
+            $validForSubmission = $ndr->validForSubmission();
+            if (count($validForSubmission['msg']) > 0) {
+                foreach ($validForSubmission['msg'] as $message) {
+                    $this->logger->warning(sprintf('Client id %s has submission issue: %s', $clientId, $message));
+                }
+
+                if (!$validForSubmission['valid']) {
+                    throw new \RuntimeException(sprintf('Submission validation failed for client id '.$clientId));
```
townxelliot, line 241: Are we handling validation using exceptions? Should we be giving the user information about why their submission failed? Also, I don't think the client ID will mean much to them. (Assuming the runtime exception will cause something to be displayed to the user.)

I think this is OK short term, but we should try to give the validation errors back to the user. Perhaps put that in a bug ticket?

client/app/src/Entity/Ndr/Ndr.php
```diff
@@ -636,4 +641,39 @@ public function setClientBenefitsCheck(?ClientBenefitsCheck $clientBenefitsCheck
...
+    /**
+     * Checks if NDR is in submittable state.
+     *
+     * @return array
+     */
+    public function validForSubmission()
+    {
+        $valid = true;
+        $msg = [];
+        if ($this->nullOrEmptyString($this->getClient()->getCourtDate())) {
```
townxelliot, line 654: `empty` does the same as your method.

townxelliot (changes requested):

client/app/src/Controller/Ndr/NdrController.php
jamesrwarren: Yeah no this will get swallowed by symfony so user will just see the generic application error screen. As discussed, the actual fix for the root cause of this will be further upstream before they get to this point. This simply protects against us creating the report pdf and other db updates if the app isn't in fit state to submit. I mentioned doing validation but it would really require redirects and messaging around what user needs to do etc. It's easier for us to simply throw this error and fix the upstream (another bug ticket).

In terms of what we see, this gives us a CRITICAL php error in the logs which is nice for us as we get alerted and the message will make sense to us.

townxelliot (approved):

PR #1825, opened 2025-02-24
DDLS-498 build for different architectures

jamesrwarren:
## Purpose
For cost savings and better performance

Fixes DDLS-498

## Approach
To get those sweet cost savings and more efficient more eco friendly cpu cycles, we should use graviton. To do that we need to use arm64 architecture and muck around with some task definition settings.

So at first I tried multi-arch builds but my github actions runner didn't like that. Did a lot of debugging and looking through how the JS docker action handles it and I'm not convinced it does. In any case, I can still specify a single platform as part of the build which is fine as the images we upload are only used in AWS. The ones on our machine can just use amd64.

The other problems were that confd is a amd64 binary so had to put an arg in for that... The other issue was waitforit didn't have a arm64 binary so I ditched it and created a shell script that did the same thing. Less dependencies to worry about too.

## Learning
[Announcing AWS Graviton2 Support for AWS Fargate – Get up to 40% Better Price-Performance for Your Serverless Containers | Amazon Web Services](https://aws.amazon.com/blogs/aws/announcing-aws-graviton2-support-for-aws-fargate-get-up-to-40-better-price-performance-for-your-serverless-containers/)

## Checklist

### Frontend


jogilder (approved):
LGTM

jay-whitwell (approved):

PR #1834, opened 2025-03-03
DDLS-498 build for graviton architecture (with more testing)

jamesrwarren:
Tested building the images without the cache and they build much more respectable time now. I've used native runners instead of QEMU and that seems to help nicely as it doesn't emulate anymore and just builds for the environment that we want.


charlesmmarshall (approved):

townxelliot (approved):
lgtm; my only question is why are we building ARM64 images? Just interested in if/why they are better. Is it because that's the arch on our laptops?

PR #1852, opened 2025-03-13
DDLS-482: Add CourtOrder demo data

MiaGordon91:
## Purpose
As part of the work towards a court order centric model, populate the multi-join court order tables with demo data when the make-reset-fixtures command is ran.

Fixes DDLS-482

## Approach
- Generate a migration file to update the data type of the court order uid field to varchar(36)
- Associate court orders with reports
- Update the LayUserFixtures class to populate the court order and linked tables with demo data for Lay deputies including hybrids, NDRs, multi-client deputies and co-deputies

## Learning

## Checklist

### Frontend


api/app/src/DataFixtures/LayUserFixtures.php
```diff
@@ -149,6 +161,11 @@ class LayUserFixtures extends AbstractDataFixture
     ];
 
     private Deputy $deputy;
+    private ?Client $client2 = null;
```
townxelliot: I think I would prefer it if we stuck to how the code was previously, where entities aren't stored in member variables. We can just use local variables for these, and pass them between methods where needed. Where you are adding new variables, these can just be appended to the existing method signatures.

townxelliot:
The `__construct` method for LayUserFixtures can be removed, as it has no effect.

townxelliot:
I've added a commit to show how I would convert the member variables you've added into local variables. We should avoid creating member variables in classes for transient variables which aren't going to be used later. The issue with adding member variables is that it can sometimes be difficult to untangle where their values are set, so the preference should be to use local variables where possible (as was done in the existing version of the code).

The process for fixing this one was:

1. Remove the member variable declaration `private Client $client2`
2. Find an instance of the member variable `$this->client2`
3. Refactor it with Refactor > Introduce Variable (in phpstorm)
4. Replace all instances of the member variable
5. Check where the new local variable is not accessible in private methods (because it used to be referred to via `$this->client2`) and fix the signatures and calls to those methods

I can take you through this in detail if it's useful.

townxelliot (approved):

PR #1859, opened 2025-03-18
DDLS-517 remove confd from client side and nginx

jamesrwarren:
## Purpose
Remove confd from client side and nginx

Fixes DDLS-517

## Approach
confd is horribly out of date, is confusing for devs and means we can't make our container filesystems readonly because it templates out files at runtime.

Instead for nginx the config only changes between local build and AWS build so we can set all the variables at build time rather than runtime. There aren't a huge number of vars to replace so the simplest to understand and maintain is just using envsubst as it's part of the nginx image anyway.

On the app side one of the files didn't have variables in it anyway so we can just straight copy it across. We can do a simple replace for the meta.json.

This just leaves the parameters.yml. Now this has a lot of env specific env vars in it. We can just have these as env vars and they will be loaded in on startup. 

However, there is also some config differences between local and AWS config. As such I have gone for a simplistic templating approach with no external tools that need maintaining. I thought about using jinja or incenteev or something but why have the overhead of maintaining them? In the end I just went for a simple shell script to generate the parameters.yml. If we get anything more complicated to do in the future then a sheel script should be able to handle that just fine.

So as part of the build not the shell script builds the parameters.yml based on an arg passed to the build.

## Learning
NA

## Checklist

### Frontend


.github/workflows/_build-and-push.yml
```diff
@@ -136,6 +137,7 @@ jobs:
             --platform linux/${PLATFORM} \
             --load \
             --build-arg PLATFORM=${PLATFORM} \
+            --build-arg TAG=${TAG} \
```
jamesrwarren, line 140: if TAG isn't an arg for a container then it just does nothing.

.gitignore
```diff
@@ -42,6 +42,7 @@ log/pact.log
 **/client/resources/public/images
 **/client/resources/public/*.png
 **/client/resources/public/*.ico
+**/parameters.yml
```
jamesrwarren, line 45: we don't want to ever add the parameters.yml as it's generated at build time

client/app/frontend.env
```diff
@@ -65,3 +65,6 @@ SSM_LOCALSTACK=true
 SECRETS_MANAGER_LOCALSTACK=true
 
 WORKSPACE=local
+HTMLTOPDF_ADDRESS=http://htmltopdf:8080
```
jamesrwarren, line 68: might as well have these as vars instead of logic

scripts/custom_sql_query/run_custom_query.py
```diff
@@ -197,6 +197,8 @@ def main(
     expected_before=None,
     expected_after=None,
 ):
+    sql_file = f"/function/{sql_file}"
```
jamesrwarren, line 200: This is a fix to our custom sql script that i thought i might as well add to this PR. It's not part of the functionality of this PR

api/docker/web/Dockerfile
```diff
@@ -8,20 +12,17 @@ WORKDIR /www/data
...
-# Add Confd to configure nginx on start
-ENV CONFD_VERSION="0.16.0"
-ADD https://github.com/kelseyhightower/confd/releases/download/v${CONFD_VERSION}/confd-${CONFD_VERSION}-linux-${PLATFORM} /usr/local/bin/confd
-RUN chmod +x /usr/local/bin/confd
+COPY --chown=nginx api/docker/web/config/default.conf /etc/nginx/conf.d/default.conf.tmpl
+COPY --chown=nginx api/docker/web/config/nginx.conf /etc/nginx/nginx.conf
 
-COPY --chown=nginx client/docker/web/confd /etc/confd
-COPY --chown=nginx client/resources/public/ /www/data/public/
+RUN envsubst '$APP_HOST $APP_PORT' < /etc/nginx/conf.d/default.conf.tmpl > /etc/nginx/conf.d/default.conf
```
townxelliot, line 18: The `X < Y > Z` syntax is a bit unintuitive. I personally prefer something more like `cat Y | envsubst X > Z` as it feels a bit more intuitive (left to right). But maybe that's just me.

api/docker/web/health-check.sh
```diff
@@ -0,0 +1,59 @@
+#!/bin/sh
```
townxelliot, line 1: Is it not possible to just use docker health checks and curl for this, or is it because we're doing multi-endpoint health checking and you're trying to combine outputs?

client/docker/app/config/generate_parameters_yml.sh
```diff
@@ -0,0 +1,129 @@
+#!/bin/sh
+
```
townxelliot, line 2: Do we need to have all the parameters inlined here? Couldn't we just use envsubst again and concatenate files together which contain placeholders? Not a deal breaker, but this isn't as clear as it might be.

townxelliot (approved):
It looks OK to me. Have you checked stuff still works locally?

api/docker/web/health-check.sh
jamesrwarren: The second option. This is just moving this health-check over to the api side as it was pulling from the client side so this has been running for a long time as is. I can explain the full reasoning behind this healthcheck in a convo after standup if you like as it has a bit of history.

api/docker/web/Dockerfile
jamesrwarren: Hmm yeah you might be right there. I'll probably change it when I do the API side next

jay-whitwell (approved):

PR #1873, opened 2025-03-20
Realigns Database; sets text type on description field

Raffers:
## Purpose
Realigns database

Fixes DDLS-####

## Approach

## Learning

## Checklist

### Frontend


api/app/src/Migrations/Version292.php
```diff
@@ -0,0 +1,49 @@
...
+
+    public function up(Schema $schema): void
+    {
+        // this up() migration is auto-generated, please modify it to your needs
+        $this->addSql('ALTER TABLE court_order ALTER court_order_uid TYPE BIGINT USING court_order_uid::bigint');
+        $this->addSql('ALTER TABLE deputy DROP CONSTRAINT FK_28FA6B9FA76ED395');
+        $this->addSql('ALTER TABLE deputy ADD CONSTRAINT FK_28FA6B9FA76ED395 FOREIGN KEY (user_id) REFERENCES dd_user (id) ON DELETE CASCADE NOT DEFERRABLE INITIALLY IMMEDIATE');
+        $this->addSql('ALTER TABLE document ALTER sync_attempts SET DEFAULT 0');
+        $this->addSql('ALTER TABLE pre_registration ALTER hybrid TYPE VARCHAR(6)');
+        $this->addSql('ALTER TABLE pre_registration ALTER client_firstname TYPE VARCHAR(100)');
```
townxelliot, line 28: This potentially truncates fields. Have you checked whether there are any client_firstname values longer than 100 chars?

api/app/src/Migrations/Version292.php
```diff
@@ -0,0 +1,49 @@
...
+{
+    public function getDescription(): string
+    {
+        return '';
+    }
+
+    public function up(Schema $schema): void
+    {
+        // this up() migration is auto-generated, please modify it to your needs
+        $this->addSql('ALTER TABLE court_order ALTER court_order_uid TYPE BIGINT USING court_order_uid::bigint');
```
townxelliot, line 23: I'm not sure why the type of this field changed. It should be a varchar I think?

townxelliot (changes requested):

api/app/src/Migrations/Version292.php
Raffers: It was, but the type was updated on doc hint to bigint

api/app/src/Migrations/Version292.php
```diff
@@ -0,0 +1,49 @@
...
+        $this->addSql('ALTER TABLE deputy DROP CONSTRAINT FK_28FA6B9FA76ED395');
+        $this->addSql('ALTER TABLE deputy ADD CONSTRAINT FK_28FA6B9FA76ED395 FOREIGN KEY (user_id) REFERENCES dd_user (id) ON DELETE CASCADE NOT DEFERRABLE INITIALLY IMMEDIATE');
+        $this->addSql('ALTER TABLE document ALTER sync_attempts SET DEFAULT 0');
+        $this->addSql('ALTER TABLE pre_registration ALTER hybrid TYPE VARCHAR(6)');
```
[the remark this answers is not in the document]
Raffers: Yer, the max will only ever be six currently (SINGLE, DUAL, HYBRID)

api/app/src/Migrations/Version292.php
Raffers: Will double check, but there really shouldn't be anything that long as a first name.

Raffers: Current top is 21 characters

townxelliot (approved):

townxelliot (changes requested):
I'm just unapproving Chris as I want to double check the cascades, if you'll give me a few minutes.

api/app/src/Migrations/Version292.php
```diff
@@ -0,0 +1,47 @@
...
+    {
+        // this up() migration is auto-generated, please modify it to your needs
+        $this->addSql('ALTER TABLE deputy DROP CONSTRAINT FK_28FA6B9FA76ED395');
+        $this->addSql('ALTER TABLE deputy ADD CONSTRAINT FK_28FA6B9FA76ED395 FOREIGN KEY (user_id) REFERENCES dd_user (id) ON DELETE CASCADE NOT DEFERRABLE INITIALLY IMMEDIATE');
+        $this->addSql('ALTER TABLE document ALTER sync_attempts SET DEFAULT 0');
+        $this->addSql('ALTER TABLE pre_registration ALTER hybrid TYPE VARCHAR(12)');
+        $this->addSql('ALTER TABLE pre_registration ALTER client_firstname TYPE VARCHAR(100)');
+        $this->addSql('ALTER TABLE user_research_response DROP CONSTRAINT FK_3B9FE71A5423F28F');
+        $this->addSql('ALTER TABLE user_research_response DROP CONSTRAINT FK_3B9FE71AA76ED395');
+        $this->addSql('ALTER TABLE user_research_response ADD CONSTRAINT FK_3B9FE71A5423F28F FOREIGN KEY (researchType_id) REFERENCES research_type (id) ON DELETE CASCADE NOT DEFERRABLE INITIALLY IMMEDIATE');
```
townxelliot, line 30: We probably want to remove this cascade delete, otherwise deleting a research type deletes the related user_research_response records.

api/app/src/Migrations/Version292.php
```diff
@@ -0,0 +1,47 @@
+<?php
```
townxelliot, line 1: There is a Version292.php file now, so this will have to move to Version293.

api/app/src/Migrations/Version292.php
```diff
@@ -0,0 +1,47 @@
...
+    public function getDescription(): string
+    {
+        return '';
+    }
+
+    public function up(Schema $schema): void
+    {
+        // this up() migration is auto-generated, please modify it to your needs
+        $this->addSql('ALTER TABLE deputy DROP CONSTRAINT FK_28FA6B9FA76ED395');
+        $this->addSql('ALTER TABLE deputy ADD CONSTRAINT FK_28FA6B9FA76ED395 FOREIGN KEY (user_id) REFERENCES dd_user (id) ON DELETE CASCADE NOT DEFERRABLE INITIALLY IMMEDIATE');
```
townxelliot, line 24: Just to confirm: deleting a dd_user record will also delete the deputy record, providing the deputy isn't referenced in the court_order_deputy table or elsewhere.

api/app/src/Migrations/Version292.php
Raffers: Looking at the research type table they are intrinsically linked to the UR submission.  This table contains the UID for the submission along with how the Survey was carried out, so if we're removing a row on this table we would want to remove the corresponding row as not to orphan it.

api/app/src/Migrations/Version292.php
Raffers: Correct, it would error out. Assuming all other locations do not have cascade on them as well, in which case if there is a link to that particular record it would error out.

api/app/src/Migrations/Version292.php
townxelliot: Right, but my understanding is that removing a row in the research_type table would remove all the user_research_response rows which reference it. Is that the right way round? If we remove the cascade, it would prevent removing a research_type row if it was referenced from a user_research_response, I think.

townxelliot (approved):

PR #1874, opened 2025-03-20
DDLS-524b use event subscriber for logout

jamesrwarren:
## Purpose
Since symfony/security-http 5.1: The "Symfony\Component\Security\Http\Logout\LogoutSuccessHandlerInterface" interface is deprecated, create a listener for the "Symfony\Component\Security\Http\Event\LogoutEvent" event instead

Fixes DDLS-524

## Approach
Initially I tried with a basic listener but I couldn't get it to work as intended. The documentation has a event subscriber model so I went with that and it works perfectly. Added logging to check it was being called properly etc.. 

## Learning
https://symfony.com/doc/6.3/security.html#logging-out

## Checklist

### Frontend


client/app/src/EventListener/AuthenticationHandler.php
```diff
@@ -16,11 +15,4 @@ public function onAuthenticationFailure(Request $request, AuthenticationExceptio
 
         return new RedirectResponse($referer);
     }
-
-    public function onLogoutSuccess(Request $request)
```
jamesrwarren, line 20: I added logging and ran all the end to end tests and this never got called.

jackgoodby (approved):

PR #1888, opened 2025-03-27
Ddls 554 - use RedisSessionHandler

jamesrwarren:
## Purpose
To fix the following warning:

Since snc/redis-bundle 3.6: Use Symfony built-int RedisSessionHandler instead.

Fixes DDLS-554

## Approach
This involved a bit of googling and trial and error but it turns out we only need predis now so we can completely ditch snc_redis bundle. 

Wasn't too many changes to make in the end but had some issues with getting it wired up as we're still using predis. I think this simplifies things for sure and fixes this deprecation warning without changing too much.

## Learning
[symfony sessions docs](https://symfony.com/doc/current/session.html#store-sessions-in-a-key-value-database-redis)

## Checklist

### Frontend


townxelliot:
Presumably this has to be applied out of hours, otherwise it will trash everyone's active sessions?

jamesrwarren:
That's an interesting point @townxelliot. I didn't think it would as it's merely changing the connections around but it's not changing the underlying connection string to redis or redis itself so everything in redis will still be the same but yes I'm not totally sure actually as we're changing the session handler... Hmm. Probably for safety we should do it super early morning so I won't merge it in now.

townxelliot:
> That's an interesting point @townxelliot. I didn't think it would as it's merely changing the connections around but it's not changing the underlying connection string to redis or redis itself so everything in redis will still be the same but yes I'm not totally sure actually as we're changing the session handler... Hmm. Probably for safety we should do it super early morning so I won't merge it in now.

It may do something weird like change the keys used to store the session data, making the two "drivers" incompatible with each other.

townxelliot (approved):

PR #1900, opened 2025-04-01
DDLS-536: Update client count to reflect active, archived and discharged clients

MiaGordon91:
## Purpose
The client count at the top of the admin org view page calculates the number of active and archived clients. Whereas the paginated client list include all clients - active, archived and discharged. 

OPG have asked for the client count to match the number of paginated clients. 
Agreed to show the count of active & archived clients and conditionally show the count of discharged clients (if there are any)

Fixes DDLS-536

## Approach
- Call API to retrieve full client count for the client tab to retrieve discharged clients
- Compare result with total client count on the organisation object to identify any discharged clients
- Pass discharged client count count to twig and conditionally render if not null

## Learning

## Checklist

### Frontend


client/app/src/Controller/Admin/OrganisationController.php
```diff
@@ -66,6 +70,8 @@ public function viewAction(Request $request, int $id)
         $tab = $request->get('tab', 'users');
         $currentFilters = self::getFiltersFromRequest($request);
 
+        $clientCount = $this->restClient->get('/v2/organisation/'.$id.'/clients', 'array')['count'];
```
townxelliot, line 73: I think we will be making nearly the same API call twice if we are on the clients tab. Is that the case? Does the request which includes filters (line 75) return the total client count? If it does, we don't need to do the call in 73, as `$result` already contains the data.

If two calls are required, we should only do the line 73 call inside the conditional (line 77), as the data is otherwise not going to be used.

townxelliot (changes requested):

townxelliot:
@MiaGordon91 we should also have some tests around this to ensure that the fix makes a difference.

client/app/src/Controller/Admin/OrganisationController.php
```diff
@@ -81,12 +87,15 @@ public function viewAction(Request $request, int $id)
             'currentTab' => $tab,
             'tabData' => $tabData,
             'count' => $result['count'],
+            'clientCount' => $clientCount,
```
townxelliot, line 90: Thinking about this some more, should the client count be the count of the results anyway? Presumably `$clientCount` contains the count of all the clients, but that might be misleading if the results have been filtered.

townxelliot (changes requested):

client/app/src/Controller/Admin/OrganisationController.php
MiaGordon91: The client count currently includes archived and active clients whereas the paginated list includes all clients (active, discharged and archived) which confuses admin users as the counts don't match.
Phil has confirmed that OPG wants the client count to include all clients to match the number of clients listed in the paginated list. 
I noted a content suggestion in the ticket to separate out the counts like this:
- ‘Active clients: xxx’ 
- ‘Archived clients: xxx’ - which is conditionally displayed if an organisation has archived clients
Or I could just include a note that the client count includes discharged clients.

I'll check with Phil...

townxelliot: Actually, is `$result['count']` just wrong? I would expect it to be the count of results, but it sounds like it's the count of all the clients? There's definitely a discrepancy somewhere which we need to iron out.

townxelliot (changes requested):

client/app/src/Controller/Admin/OrganisationController.php
MiaGordon91: The $result['count'] returns the total count of results depending on which tab you select. Yes, for clients it returns all clients (active, archived and discharged).

townxelliot (approved):

PR #1930, opened 2025-04-17
Ddls 597c lambda move - part 3

jamesrwarren:
## Purpose
Last PR to move from environment to account

Fixes DDLS-597

## Approach
Was making destroy jobs really slow so moving it to account where it doesn't get destroyed as part of environment teardown will really help. 

## Learning
NA

## Checklist

### Frontend


scripts/custom_sql_query/run_custom_query.py
```diff
@@ -226,6 +236,18 @@ def get_current_user():
         sys.exit(1)
 
 
+def get_db_endpoint(environment):
+    instance_id = f"api-{environment}-0"
+    rds = boto3.client("rds")
+    try:
+        response = rds.describe_db_instances(DBInstanceIdentifier=instance_id)
+        db_instance = response["DBInstances"][0]
```
townxelliot, line 244: Is it always going to be the first instance in the list? Is there any way to be more specific in finding the db?

townxelliot (approved):
I've left a comment, but if you have a good explanation you can just ignore it.

scripts/custom_sql_query/run_custom_query.py
jamesrwarren: It's based on the instanceid which is always 0 for the writer instance. It's set here :) 
`instance_id = f"api-{workspace}-0"`

PR #1939, opened 2025-04-23
DDLS-627 Migrate API controller annotations to attributes

ramchale:
## Purpose
Attributes have several benefits over annotations

Fixes DDLS-627

## Approach
Run Rector, manually review changes

## Checklist

### Frontend

### DIFF
--- a/api/app/config/routes/attributes.yaml
+++ b/api/app/config/routes/attributes.yaml
@@ -1,28 +1,28 @@
 app:
     resource: "../../src/Controller/"
-    type: annotation
+    type: attribute
     prefix: /
 
 app_v2:
     resource: "../../src/v2/Controller/"
-    type: annotation
+    type: attribute
     prefix: /v2
 
 app_v2_registration:
     resource: "../../src/v2/Registration/Controller/"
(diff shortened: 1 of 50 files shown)


townxelliot:
Have you checked that the controllers are still functional? I think you may need to alter some config in api/app/config/routes/annotations.yaml so that the correct format for annotations is used. Alternatively, if that config file doesn't need changing or otherwise has no effect, we should remove it.

ramchale:
> > Have you checked that the controllers are still functional? I think you may need to alter some config in api/app/config/routes/annotations.yaml so that the correct format for annotations is used. Alternatively, if the controllers are working and that config file doesn't need changing/has no effect, we should remove it.
> 
> Good point. I've checked they're still working. I'll try removing that config and see if it has a detrimental effect

If you remove the config file entirely it no longer works. I've updated it and checked before and after with `docker compose -f docker-compose.yml -f docker-compose.override.yml run api-app php app/console debug:router` (after updating the mounts to include the config dir), and they match.

townxelliot (approved):

PR #1950, opened 2025-04-30
DDLS-529 - Create new API Endpoint that pulls reports for a particular deputy

Raffers:
## Purpose
As part of our Court Order data restructure, the front end app will need to accommodate the fact that a client may now have more than one active Court Order within the app (Court Orders are exposed as 'reports').

As well as selecting a client, the deputy will now have to select which court order they wish to work with.

Rather than a two step process 'Choose your client' -> 'Now choose the order', we could accomplish both steps on one page using a UI interface that is visually identical to the 'Choose a Client' page that our multi-client-deputies currently see (screenshot 1). The only difference here is that where a client has a dual order (and those orders are not a single hybrid) then there would be two individual rows for the same client (H&W and PFA, see screenshot 2).

Fixes DDLS-529

## Approach

## Learning

## Checklist

### Frontend


api/app/tests/Integration/Entity/DeputyTest.php
```diff
@@ -9,7 +9,7 @@
 use Faker\Factory;
 use Symfony\Bundle\FrameworkBundle\Test\KernelTestCase;
 
-class DeputyIntegrationTest extends KernelTestCase
```
Raffers, line 12: This test was bugged

api/app/tests/Unit/Entity/DeputyTest.php
```diff
@@ -1,39 +0,0 @@
-<?php
```
Raffers, line 1: This entire file was duplicated under the integration test. Removed this test file & renamed the other

api/app/src/Controller/DeputyController.php
```diff
@@ -19,7 +18,7 @@ class DeputyController extends RestController
     public function __construct(
         private readonly DeputyService $deputyService,
         private readonly RestFormatter $formatter,
-        EntityManagerInterface $em
+        protected readonly EntityManagerInterface $em
```
townxelliot, line 21: could this be private?

api/app/src/TestHelpers/ClientTestHelper.php
```diff
@@ -11,21 +11,24 @@
 use Doctrine\ORM\EntityManager;
 use Faker\Factory;
 use PHPUnit\Framework\TestCase;
+use Prophecy\Prophet;
```
townxelliot, line 14: Can we get rid of prophet from the unit tests? I'd prefer to use vanilla phpunit for new code.

api/app/src/TestHelpers/CourtOrderTestHelper.php
```diff
@@ -0,0 +1,57 @@
+<?php
```
townxelliot, line 1: We've got a createCourtOrder method on the Fixtures class which could be used for this.

There's also a method on the Deputy class to associate a deputy with a court order (associateWithCourtOrder()), which would replace the second method.

api/app/src/TestHelpers/DeputyTestHelper.php
```diff
@@ -5,18 +5,16 @@
 namespace App\TestHelpers;
```
townxelliot, line 5: The Fixtures class has a createDeputy() method you could use instead of adding a new test helper.

api/app/src/v2/Controller/DeputyController.php
```diff
@@ -2,38 +2,69 @@
...
         return $this->buildSuccessResponse($transformedDto);
     }
+
+    #[Route(path:'/{uid}/reports', name:'deputy_find_reports_by_uid', requirements:['uid' => '\d+'], methods:['GET'])]
+    public function getAllDeputyReports(Request $request, int $uid): JsonResponse
+    {
+        /** @var User $user */
+        $user = $this->getUser();
+        
+        if ($uid !== $user->getDeputyUid()) {
```
townxelliot, line 50: Do we need to check this? Putting this constraint on the endpoint would mean that it's not usable by a superuser, for example.

townxelliot (changes requested):
I made some comments about duplication of test helper methods, but then realised that was already there, so can be ignored. There were a couple of other points I noticed.

We should add a ticket to remove the duplicated test helpers.

townxelliot:
I made DDLS-646 to cover rationalising the integration test helpers.

PR #1951, opened 2025-05-01
DDLS 493b: Postcodes too long - failing upload

MiaGordon91:
## Purpose
Ticket was re-raised due to a deputy postcode issue, it was confirmed that the deputy issue wasn't due to their postcode, but we still need to increase the database field to accommodate non UK postcodes. 
This ticket was [reverted ](https://github.com/ministryofjustice/opg-digideps/pull/1839)a while ago due to the migration autogenerating unwanted sql.

Fixes DDLS-493b

## Approach
- Generated new migration to increase database field length
- The migration still autogenerated unwanted sql which has been removed, however it did generate a change to the staging.deputyships table that has been applied in this ticket. The change is to add an id primary key and remove any reference to it's old composite key (order_uid & deputy_uid)
- Remove whitespace from postcodes before it gets persisted into the pre reg table.

## Learning

## Checklist

### Frontend


api/app/src/Entity/StagingSelectedCandidate.php
```diff
@@ -40,7 +40,7 @@ class StagingSelectedCandidate
      *
      * @ORM\SequenceGenerator(sequenceName="candidate_id_seq", allocationSize=1, initialValue=1)
      */
-    public int $id;
+    public ?int $id = null;
```
townxelliot, line 43: should this ever be null, as it's autogenerated?

townxelliot (approved):

api/app/src/Entity/StagingSelectedCandidate.php
MiaGordon91: I had to make the property nullable as it was failing on an integration test because the csv/serializer was expecting a value when converting csv data into an object. I set it as nullable so that the serializer doesn't expect id to be set and id can still be auto-generated by Doctrine once the object is persisted.

PR #1955, opened 2025-05-06
Api unit config

ramchale:
## Purpose
Alternative config that might simpler for local development

## Approach
- Reduce docker-compose files to docker_compose.yml for setup and docker_compose.override.yml for any volume mounting or additional env setup that will make local dev easier
- Rely on volume mounts rather than rebuilds so that running in the IDE is more likely to reflect changes to the code
- Run all tests, rather than stopping on the first warning or failure so that the test run gives a more complete picture of issues

## Checklist

### Frontend


townxelliot:
lgtm

jamesrwarren (approved):

PR #1969, opened 2025-05-14
DDLS-645: Allow multi-client deputies to view all clients when user account is odr enabled

MiaGordon91:
## Purpose
If a multi-client deputies user account is marked as odr_enabled, then the deputy is unable to view any clients attached to their user account that doesn't have a record in the odr (ndr) table. 

Fixes DDLS-645

## Approach
- Expand on conditional logic to also check if an ndr exists for that client before trying to retrieve the ndr's id as this currently breaks the app and prevents deputies from accessing a client that doesn't have an ndr. 

## Learning

## Checklist

### Frontend

### DIFF
--- a/api/app/src/Controller/Ndr/NdrController.php
+++ b/api/app/src/Controller/Ndr/NdrController.php
@@ -5,6 +5,7 @@ namespace App\Controller\Ndr;
 use App\Controller\RestController;
 use App\Entity as EntityDir;
 use App\Entity\Report\Document;
+use App\Repository\NdrRepository;
 use App\Service\Formatter\RestFormatter;
 use App\Service\ReportService;
 use Doctrine\ORM\EntityManagerInterface;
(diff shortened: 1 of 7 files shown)


client/app/src/Controller/Report/ReportController.php
```diff
@@ -208,7 +208,9 @@ public function clientHomepageAction(Redirector $redirector, string $clientId)
             'clientHasCoDeputies' => $this->preRegistrationApi->clientHasCoDeputies($clientWithCoDeputies->getCaseNumber()),
         ];
 
-        if ($ndrEnabled) {
+        $ndrExistsOnClient = $this->restClient->get('/ndr/client/'.$clientId, 'array');
```
townxelliot, line 211: I think we should be able to test this on the Client instance without requiring a separate API call.

api/app/src/Controller/Ndr/NdrController.php
```diff
@@ -228,4 +229,12 @@ public function update(Request $request, $id)
 
         return ['id' => $ndr->getId()];
     }
+
+    #[Route(path: '/ndr/client/{clientId}', methods: ['GET'])]
```
townxelliot, line 233: The API endpoint for Client already has an ndr field which is either an NDR or null. This data is included with the client API endpoint response, and available on the Client object on the client/ (confusing!) side of the application.

What I'd suggest is checking whether this field is null, rather than making another request to the API to find out whether the client has an NDR.

townxelliot (changes requested):

townxelliot (approved):

PR #1973, opened 2025-05-19
Updated version of Golang and Alpine

MurlockCadi:
To fix the last vulnerabilities found.


api/app/composer.json
```diff
@@ -27,7 +27,7 @@
     "require": {
         "php": "^8.2",
         "aws/aws-sdk-php": "3.329.0",
-        "doctrine/doctrine-bundle": "^2.13",
+        "doctrine/doctrine-bundle": "2.13.3",
```
townxelliot, line 30: Are we pinning this for a specific reason?

api/app/composer.json
```diff
@@ -46,6 +46,9 @@
         "symfony/console": "^6.0",
         "symfony/dependency-injection": "^6.0",
         "symfony/dotenv": "^6.0",
+        "symfony/doctrine-bridge": "6.4.21",
```
townxelliot, line 49: Same question: should these be `^` or do we need specific versions?

townxelliot (changes requested):

api/app/composer.json
MurlockCadi: @townxelliot When trying to compose them, other packages would complain that it needed a specific version and fail. It was some juggling to get the packages working together while also fixing any vulnerabilities. Me and Jim talked about this last week, and we intend to clean this up in the future.

jamesrwarren (approved):

api/app/composer.json
jamesrwarren: Yes we have added a reminder to unpin this but we had to for now to get it working as latest version doesn't work properly with other upgrade.

townxelliot (approved):

townxelliot:
Please add a ticket to sort out the dependency issues later.

PR #1988, opened 2025-05-21
DDLS-529 - Issue with SQL

Raffers:
## Purpose
While working on DDLS-576 I've noticed an issue with the SQL query behind the endpoint. For retrieving reports via court orders, returning too many results. This PR fixes the query, concats court orders if it's a hyrbird & fixes some logic in the controller itself as it was incorrectly returning an error if it was a superadmin

Fixes DDLS-529

## Approach

## Learning

## Checklist

### Frontend


api/app/src/Repository/DeputyRepository.php
```diff
@@ -44,16 +44,27 @@ public function getUidToIdMapping(): array
...
-        LEFT JOIN client c ON c.deputy_id = d.id
-        LEFT JOIN court_order co ON co.client_id = c.id
-        LEFT JOIN report r ON r.client_id = c.id
-        LEFT JOIN court_order_deputy cod ON cod.deputy_id = d.id
+        SELECT DISTINCT
+        c.firstname AS "firstName",
+        c.lastname AS "lastName",
+        c.case_number AS "caseNumber",
+        (
+            SELECT DISTINCT string_agg(co.court_order_uid, ', ')
```
townxelliot, line 52: If we're doing the concatenating in this SQL, we should order by court order UID ASC, otherwise it might look a bit odd in the UI. I'd suggest at least calling the field "court order UIDs".

Also, aren't we going to be relying on these to link to the next page? If so, it would be nicer if they were an array when returned by the API, so we can easily create the links by loop.

api/app/src/Repository/DeputyRepository.php
```diff
@@ -44,16 +44,27 @@ public function getUidToIdMapping(): array
...
+            LEFT JOIN deputy d ON cod.deputy_id = d.id
+            LEFT JOIN client c ON d.id = c.deputy_id
+            LEFT JOIN report re ON c.id = re.client_id
+            WHERE d.deputy_uid = :deputyUid
+            AND r.id = re.id
+        ) AS "courtOrderUid",
+        r.type AS "type",
+        r.id
+        FROM report r
+        LEFT JOIN client c ON r.client_id = c.id
```
townxelliot, line 64: If we're left joining, we could potentially return rows where the join returns a null. I think at least some of these should be inner joins (some of the left joins will have to not be nulls as they are in the WHERE condition, but we should probably use inner joins so we don't accidentally return rows with nulls by missing a condition).

I would guess it's not a problem because of the WHERE, but I can't see a reason why they can't be inner joins anyway.

townxelliot (changes requested):

api/app/src/Repository/DeputyRepository.php
Raffers: With how the data should be structured (excluding our current data issues)
Every report will be linked to a client
Every client will be linked to court order
Every court order will be linked to a deputy via the link table

I have just noticed that I need to include `AND cod.deputy_id = d.id` so we're targeting the correct deputy in the link table. So the other left joins are still required so we get the correct chain to the specific deputy

api/app/src/Repository/DeputyRepository.php
```diff
@@ -62,7 +73,7 @@ public function findReportsInfoByUid(int $uid, bool $includeInactive = false): ?
             // Possibly need to be changed when we have all applicable status
```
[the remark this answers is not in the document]
Raffers: Yes, we will have to change it to an OR. But I think we're going to be receiving more status than `ACTIVE` `INACTIVE`

api/app/src/Repository/DeputyRepository.php
townxelliot: > It's my understanding we're going to directly link to the report from this page & not the court order itself.
> 
> The only reason I've had to include the concat is due to hybrid's which have two court orders associated with each report, which I failed to consider when initially doing this ticket & happened across incidentally while testing [DDLS-576](https://opgtransform.atlassian.net/browse/DDLS-576)

If we're going to be linking to the report, I think we're going to need the report ID in the response.

api/app/src/Repository/DeputyRepository.php
townxelliot: Would it be better to enforce those assumptions with inner joins? There could potentially not be a court_order_deputy entry for a court order, and an inner join would exclude that court order.

Also, you wouldn't need the extra AND clause you mentioned if you use an inner join, because there would be no chain from court_order_deputy to the deputy table. At the moment there could be, as you'll get all the rows for the court order, regardless of deputy.

api/app/src/Repository/DeputyRepository.php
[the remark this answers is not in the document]
townxelliot: Right, I suppose the question is what we want to include. If we want all records, then just negating it would do the trick. I am reading "inactive" as "not active", rather than the more specific "include the INACTIVE status as well".

townxelliot (changes requested):

api/app/src/Repository/DeputyRepository.php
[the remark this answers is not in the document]
Raffers: Yer, having looked at it again. It's poor naming that's causing the confusion. As the default is set to `false`
Have the query it so it's more clear & specific

api/app/src/Repository/DeputyRepository.php
Raffers: I would expect them to be excluded on this endpoint. We're trying to retrieve reports that are linked to deputies via court orders. So there should be an entry on `court_order_deputy` 

I can't think of a situation where we don't have the chain of 
```
reports > court order > court_order_deputy > deputy
                      > client
```
If we have a deputy signed up & starting to complete reports this chain should always exist. Although looking at it again, I realised I've gone of the deputy id on the client table rather than going via court order

townxelliot: I'm still not sure why we need left joins instead of inner joins, though. Perhaps we should have a chat about it tomorrow.

api/app/src/Repository/DeputyRepository.php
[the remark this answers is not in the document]
townxelliot: Yes, originally we were going to store the actual status, but we're now storing whether the deputy is active or not as a boolean. We should probably rename the field for clarity.

townxelliot (changes requested):

api/app/src/TestHelpers/DeputyTestHelper.php
```diff
@@ -6,11 +6,12 @@
 
 use App\Entity\Deputy;
 use App\Entity\User;
+use Doctrine\ORM\EntityManager;
 use Faker\Factory;
 
 class DeputyTestHelper
 {
-    public static function generateDeputy(?string $email = null, ?string $deputyUid = null, ?User $user = null): Deputy
+    public static function generateDeputy(?string $email = null, ?string $deputyUid = null, ?User $user = null, ?EntityManager $em = null): Deputy
```
townxelliot, line 14: Looking at this in more detail, I would just get the returned Deputy and save it, rather than passing an entity manager around.

townxelliot (changes requested):

api/app/tests/Integration/Controller/DeputyControllerTest.php
```diff
@@ -190,8 +192,9 @@ public function testGetDeputyReportsReturnsResults()
         self::$fixtures->flush();
 
         // generate courtOrder and set client and deputy
-        $courtOrder = self::$fixtures->createCourtOrder(7055555550, 'pfa', true);
+        $courtOrder = self::$fixtures->createCourtOrder(7055555550, 'pfa', 'ACTIVE');
         $courtOrder->setClient($client);
+        $courtOrder->setStatus('ACTIVE');
```
townxelliot, line 197: Is this line redundant, as you're setting the status when the court order is created?

api/app/tests/Integration/Entity/Repository/DeputyRepositoryTest.php
```diff
@@ -91,4 +91,10 @@ public function testFindReportsInfoByUidIsNull()
 
         self::assertNull($results);
     }
+
```
townxelliot, line 94: I'd put this at the top of the class.

townxelliot (changes requested):

townxelliot (approved):

PR #2002, opened 2025-06-02
New SSM EC2 Instance

MurlockCadi:


terraform/account/region/modules/ssm_ec2_instance/boot.sh
```diff
@@ -0,0 +1,231 @@
...
+}
+
+list_databases() {
+  printf "%-30s %-70s %-35s %-35s\n" "Database" "Endpoint" "digidepsmaster password" "readonly password"
+  printf "%-30s %-70s %-35s %-35s\n" "---------" "---------" "------------------------" "------------------"
+
+  dbs=$(aws rds describe-db-clusters --query "DBClusters[?Engine=='aurora-postgresql'].[DBClusterIdentifier,Endpoint]" --output text)
+
+  while read -r identifier endpoint; do
+    if [[ "$identifier" == *ddls* ]]; then
```
jamesrwarren, line 27: might not always be called ddls

PR #2041, opened 2025-07-02
DDLS-767: Update Org Dashboard to show report type

MiaGordon91:
## Purpose
Update org dashboard to accommodate the rare scenario where an org might be involved with both sides of a dual order.

Fixes DDLS-767

## Approach
- Add report type definitions for orgs
- Change 'court order number' to 'case number'
- Update repository query - reports are still ordered by end_date however dual reports linked to the same client are grouped together. Screenshot of an example is in the [ticket](https://opgtransform.atlassian.net/jira/software/c/projects/DDLS/boards/206?selectedIssue=DDLS-767)
- Added integration test for updated ordering within report repository 
- Noticed some unit tests were sitting in reportRepository integration test file, so moved to unit test directory.

## Learning

## Checklist

### Frontend


api/app/src/Repository/ReportRepository.php
```diff
@@ -131,8 +131,17 @@ public function getAllByDeterminant(mixed $orgIdsOrUserId, int $determinant, Par
...
+                // join report table to itself to calculate the earliest due date for dual reports if they exist
+                ->innerJoin(
+                    'App\Entity\Report\Report',
+                    'r2',
+                    'WITH',
+                    'r2.client = r.client'
+                )
+                // allows sorting by the earliest due date for each grouped set of reports linked to the same client
+                ->addSelect('MIN(r2.dueDate) AS HIDDEN minDueDate')
                 ->where('o.isActivated = true AND o.id in ('.implode(',', $orgIdsOrUserId).')');
```
townxelliot, line 145: You need to be careful with implode in SQL. You should use a parameter instead, and pass an array as the value of the parameter, something like:
```
->where('o.isActivated = true AND o.id in :orgIdsOrUserId')
->setParameter('orgIdsOrUserId', $orgIdsOrUserId)
```

You might need to experiment with this a bit, as I'm not sure how Doctrine works with parameter values which are arrays.

townxelliot (changes requested):

townxelliot:
I think this is fine how you have it, but could probably be compressed to:

```
->andWhere($qb->expr()->in('o.id', $orgIdsOrUserId))
```

I would expect Doctrine to do the right thing in this situation and properly construct the IN clause from an array.

townxelliot (approved):

PR #2061, opened 2025-07-14
DDLS-568 API integration tests should not be order-dependent

iqpalm:
## Purpose

Fixes DDLS-568

## Approach
1. Be able to run individual test suites

## Learning

## Checklist

### Frontend

### DIFF
--- a/api/app/src/Repository/DeputyRepository.php
+++ b/api/app/src/Repository/DeputyRepository.php
@@ -102,7 +102,6 @@ class DeputyRepository extends ServiceEntityRepository
                     'caseNumber' => $line['caseNumber'],
                 ],
                 'report' => [
-                    'id' => $line['reportId'] ?? '',
                     'type' => $line['type'],
                 ],
                 'courtOrder' => [


api/app/src/Model/DeputyshipProcessingRawDbAccess.php
```diff
@@ -77,6 +77,15 @@ public function insertOrder(array $insertOrder): DeputyshipProcessingRawDbAccess
         try {
             $qb = $this->ingestWriterEm->getConnection()->createQueryBuilder();
 
+            $qb->insert('client')
```
townxelliot, line 80: The test needs a client with ID 1 to be inserted so that it will pass, but the client **MUST NOT** be inserted here, only in the test.

The candidate should have a reference to an existing client already. We should not be adding the client when we insert the order, as this will cause chaos. Each insert/update function on this class should only insert/update a court_order, court_order_deputy, or court_order_report row. If it inserts anything else, it potentially interferes with or disrupts existing data.

api/app/tests/Integration/Model/DeputyshipProcessingRawDbAccessIntegrationTest.php
```diff
@@ -64,7 +64,7 @@ public function testInsertOrderAndFindOrderId(): void
         $this->sut->endTransaction();
 
         // use SUT to find the ID of the order just inserted
-        $orderId = $this->sut->findOrderId($uid);
+        $orderId = $this->sut->findOrderId($uid)->data;
```
townxelliot, line 67: I'm not sure how this test is passing atm, but probably because the client exists.

PR #2094, opened 2025-08-05
DDLS-819 Split the Report submission page into separate Contact details and Declaration pages

iqpalm:
## Purpose

Fixes DDLS-819

## Approach
1. Add in new contact details page
2. Change existing declaration page to only include declaration part pf the submission process
3. Delete breadcrumbs and use buttons for navigation instead
4. Ensure edit in contact details page will then return the user back to the contact details page

## Learning

## Checklist

### Frontend


api/app/tests/Behat/bootstrap/v2/Common/IShouldBeOnFrontendTrait.php
```diff
@@ -36,6 +36,14 @@ public function iAmOnNdrReviewPage(): bool
         return $this->iAmOnPage('/ndr\/.*\/review$/');
     }
 
+    /**
+     * @Then I should be on the confirm your details page
+     */
+    public function iAmOnReportConfirmDetailsPage(): bool
+    {
+        return $this->iAmOnPage('/report\/.*\/confirm-details$/');
```
Raffers, line 44: You need to be specific here. `.*` is a greedy quantifier you want to use `\d` in there at the very least. I know we have been using `.*` everywhere, but it's a bad pattern and if something goes wrong it makes it harder to debug, specially if we somehow pass a string instead of an int to the url.

api/app/tests/Behat/bootstrap/v2/Common/ReportTrait.php
```diff
@@ -61,7 +61,20 @@ private function submitSteps(string $ndrOrReport, int $reportId)
             }
         }
 
-        $this->clickLink('Continue');
+        if ('ndr' == $ndrOrReport) {
+            $this->clickLink('Continue');
+        } else {
+            $this->clickLink('Confirm contact details');
```
Raffers, line 67: If this is now a multiple step process then we should be separating them out into different functions and behat steps.

We should also be confirming what it is we expect to see on these pages and not just relying on there being a button

jackgoodby (approved):

PR #2128, opened 2025-09-02
Bring API Integration tests up to current standards - Part 1

ramchale:
## Purpose
Easier to maintain and stricter type checking should help reduce potential bugs

## Checklist

### Frontend


townxelliot (changes requested):
I made a few specific comments, but there are some general notes which apply across a lot of these changes:

1. We have previously used `\DateTime`, `\Exception` etc. in a lot of places in the code. If we're now moving to import these via `use` instead, we're going to need to make a lot of changes to make things consistent.
2. I personally prefer using `test*` method names in test cases rather than an explicit `[# Test]` annotation. This is what we do fairly consistently across the code base, so I am reticent to move to annotations instead.

I'm a bit surprised phpstan let these through, as when I try to commit `use Exception`...`catch (Exception $e)...` and similar, phpstan complains and stops me doing that. Perhaps something has changed in the phpstan config while I was away?

I'll do another review once we've addressed the above, to catch the specific stuff.

PR #2167, opened 2025-10-02
DDLS-947 Error summary links do not move focus to the relevant field - admin app, org users

iqpalm:
## Purpose

Fixes DDLS-947

## Approach
For input fields focus on the input field when error
For checkbox focus on checkbox group when error

## Learning

## Checklist

### Frontend

### DIFF
--- a/client/app/composer.json
+++ b/client/app/composer.json
@@ -5,9 +5,6 @@
     "description": "The \"Symfony Standard Edition\" distribution",
     "config": {
         "sort-packages": true,
-        "platform": {
-            "php": "8.2.28"
-        },
         "allow-plugins": {
             "php-http/discovery": true
         }
@@ -19,7 +16,7 @@
         }
     },
     "require": {
-        "php": "^8.2",
+        "php": "^8.4",
         "alphagov/notifications-php-client": "^4.0.1",
         "aws/aws-sdk-php": "^3.337.3",
         "doctrine/collections": "^2.0.0",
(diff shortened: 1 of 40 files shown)


client/app/templates/Components/Alerts/_validation-summary.html.twig
```diff
@@ -10,7 +10,7 @@
                 {% endfor %}
             {% endif %}
             {% for formErrorMessage in formErrorMessages %}
-                <li><a href="#form-group-{{ formErrorMessage.elementId }}">{{ formErrorMessage.message | nl2br}}</a></li>
+                <li><a href="#{{ formErrorMessage.elementId }}" aria-invalid="true" aria-describedby="{{ formErrorMessage.elementId }}-error">{{ formErrorMessage.message | nl2br}}</a></li>
```
townxelliot, line 13: I think this should be on the element linked to, not on the link. "The aria-describedby attribute lists the ids of the elements that describe the object." In this case, the object is the form element, and the element that describes it is the error message.

client/app/templates/Components/Form/_checkboxgroup.html.twig
```diff
@@ -55,7 +55,10 @@
                     {% set attr = attr | merge({'data-aria-controls': choiceElement.vars.id ~ '-conditional'}) %}
                 {% endif %}
 
-                {{ form_widget(choiceElement, {'attr': attr}) }}
+                <div id="{{ element.vars.id }}">
```
townxelliot, line 58: I think we should be putting the ID on the form widget, not on a div around it. Also note my previous comment: I think aria-described by should be on the form widget, not on the link to it.

townxelliot (changes requested):

PR #2410, opened 2026-03-20
DDLS-1235 Timing out of org fixtures

iqpalm:
## Purpose
The org fixtures are not being created when the number of users selected is high . This also sometimes occurs when the number of clients selected is high as well. 

Fixes DDLS-1235

## Approach
The approach was to create users in batches
The creation of additional clients has been moved to a new endpoint for organisations.
It has been agreed to reduce the number of org client fixtures to 100 and org users to 50 until we investigate a more robust way of creating a huge number of fixtures as they are timing out

## Learning

## Checklist

### Frontend


client/app/src/Controller/Admin/FixtureController.php
```diff
@@ -150,26 +150,64 @@ public function orgCourtOrdersAction(Request $request): array
...
+                        'reportType' => $submittedFormData['reportType'],
+                        'reportStatus' => $submittedFormData['reportStatus'],
+                        'courtDate' => $formAndRequestData['courtDate']->format('Y-m-d'),
+                        'activated' => $submittedFormData['activated'],
+                        'orgSizeUsers' => $numberOfUsersToCreate,
+                        'deputyUid' => $formAndRequestData['deputyUid'],
+                    ]),
+                    ['timeout' => 1000]
+                );
+                $remainingUsersToCreate -= $numberOfUsersToCreate;
```
townxelliot: Shouldn't this be `-= 1` (reduce the number of remaining users to create by 1 as we just created one user)?

client/app/src/Controller/Admin/FixtureController.php
```diff
@@ -150,26 +150,64 @@ public function orgCourtOrdersAction(Request $request): array
...
+            }
 
-            $query = ['query' => ['filter_by_ids' => implode(',', $response['deputyIds'])]];
-            $deputiesData = $this->restClient->get('/user/get-all', 'array', [], $query);
-            $sanitizedDeputyData = $this->removeNullValues($deputiesData);
+            $numberOfAdditionalClientsToCreate = min($submittedFormData['orgSizeClients'] - 1, 9);
+            $remainingAdditionalClientsToCreate = $submittedFormData['orgSizeClients'] === 1 ? 0 : $submittedFormData['orgSizeClients'] - 1;
```
townxelliot: Couldn't this just be set to `$numberOfAdditionalClientsToCreate`? Alternatively, why not just use `$numberOfAdditionalClientsToCreate` and reduce it by 1 each time you create a client, until it reaches 0.

Same goes for the previous loop: I don't think you need two variables for this.

api/app/src/v2/Fixture/Controller/FixtureController.php
```diff
@@ -132,11 +170,14 @@ private function createSingleClientDeputy(mixed $fromRequest, Client $client): U
             $this->createNdr($fromRequest, $client);
             $user->setNdrEnabled(true);
         } else {
-            $report = $this->generateReport($fromRequest, $client);
-            $this->em->persist($report);
-            if (User::TYPE_LAY === $fromRequest['deputyType']) {
-                $courtOrder->addReport($report);
-                $this->em->persist($courtOrder);
+            if (!$this->reportRepository->findOneBy(['client' => $client])) {
```
townxelliot: Could be collapsed into an else if.

api/app/src/v2/Fixture/Controller/FixtureController.php
```diff
@@ -335,51 +376,30 @@ private function createOrgAndAttachParticipants($fromRequest, User $deputy, Clie
         $uniqueOrgNameSegment = (preg_match('/\d+/', $fromRequest['deputyEmail'], $matches)) ? $matches[0] : rand(0, 9999);
         $orgName = sprintf('Org %s Ltd', $uniqueOrgNameSegment);
 
-        if (null === ($organisation = $this->organisationRepository->findOneBy(['name' => $orgName]))) {
+        if (null === ($this->organisationRepository->findOneBy(['name' => $orgName]))) {
```
townxelliot: Brackets are no longer needed around the right hand side of this condition.

townxelliot (dismissed):
lgtm, just made a couple of suggestions but they are not vital

api/app/src/FixtureFactory/UserFactory.php
```diff
@@ -96,11 +96,11 @@ public function createAdmin(array $data): User
     /**
      * @return User|void
      */
-    public function createGenericOrgUser(Organisation $organisation)
+    public function createGenericOrgUser(Organisation $organisation, int $number)
     {
         $faker = Factory::create();
 
-        $email = sprintf('%s.%s@%s', $faker->firstName(), $faker->lastName(), $organisation->getEmailIdentifier());
+        $email = sprintf('%s.%s.%s@%s', $faker->firstName(), $faker->lastName(), $number, $organisation->getEmailIdentifier());
```
townxelliot: You'll also need to remove these references to Faker.

client/app/src/Controller/Admin/FixtureController.php
```diff
@@ -150,26 +150,62 @@ public function orgCourtOrdersAction(Request $request): array
...
-                'reportStatus' => $submittedFormData['reportStatus'],
-                'courtDate' => $formAndRequestData['courtDate']->format('Y-m-d'),
-                'activated' => $submittedFormData['activated'],
-                'orgSizeClients' => $submittedFormData['orgSizeClients'],
-                'orgSizeUsers' => $submittedFormData['orgSizeUsers'],
-                'deputyUid' => $formAndRequestData['deputyUid'],
-            ]));
+            $remainingUsersToCreate = $submittedFormData['orgSizeUsers'];
+            $response = [];
+            while ($remainingUsersToCreate !== 0) {
```
townxelliot: Reading through this again, we've moved from sending one request to create 500 users to sending 500 requests creating one user each. We should find a better way to batch this, as doing 500 requests will timeout the response.

I suggest batching them so we create 50 or 100 users at a time, and make the appropriate amount of requests to create the total number of users we want.

townxelliot (changes requested):

client/app/src/Controller/Admin/FixtureController.php
iqpalm: Thanks will revert it back as that was what I originally implemented

townxelliot (dismissed):

api/app/src/v2/Fixture/Controller/FixtureController.php
```diff
@@ -91,6 +101,34 @@ public function createCourtOrder(Request $request): JsonResponse
...
+
+        /** @var Organisation $organisation */
+        $organisation = $this->organisationRepository->findByEmailIdentifier($fromRequest['deputyEmail']);
+
+        foreach (range(1, $fromRequest['orgSizeClients']) as $ignored) {
+            $orgClient = $this->clientFactory->createGenericOrgClient($deputy, $organisation, $fromRequest['courtDate']);
+            $this->em->persist($orgClient);
+
+            $report = $this->generateReport($fromRequest, $orgClient);
+            $this->em->persist($report);
```
townxelliot, line 128: This is still likely to cause memory issues as you only flush once all the entities are created. They will be held in memory in the mean time. You need to flush occasionally to prevent this from happening, but not on every entity as that will make the request time out.

iqpalm: But it is based on batching of 10 which i thought would be ok to flush once or do you mean I need to flush more than once in each batch

townxelliot (dismissed):
lgtm, but may need some more testing by product team as the algorithm is different

olivier-nicodemi (dismissed):

townxelliot (approved):

PR #2488, opened 2026-04-16
DDLS-1198 - Set/update the report type via the deputyship ingest 

Raffers:
## Purpose
As we move towards supporting dual reporters we will aim to have one ingest that sets deputy-court-order-report associations.

As a step-change towards this, we will update report types during the deputyship ingest. 

By the end of this ticket the deputyship ingest should be setting the report type correctly.

Fixes DDLS-1198

## Approach

## Learning

## Checklist

### Frontend

### DIFF
--- a/api/app/src/Domain/Report/ReportType.php
+++ b/api/app/src/Domain/Report/ReportType.php
@@ -30,7 +30,7 @@ final readonly class ReportType implements \Stringable
         $parts[] = '10' . $suffix;
 
         if ($this->courtOrderKind === CourtOrderKind::Hybrid) {
-            $parts[] = substr(CourtOrderReportType::OPG104->getSuffix(), -1);
+            $parts[] = CourtOrderReportType::OPG104->getSuffix();
         }
 
         if ($this->deputyType !== DeputyType::LAY) {
--- a/api/app/src/Domain/Report/ReportType.php
+++ b/api/app/src/Domain/Report/ReportType.php
@@ -40,10 +40,13 @@ final readonly class ReportType implements \Stringable
         return implode('-', $parts);
     }
 
-    public static function tryFrom(string $value): ReportType
+    public static function tryFrom(string $value): ?ReportType
     {
-        $courtOrderReportType = CourtOrderReportType::tryFrom('OPG' . substr($value, 0, 3)) ??
-            CourtOrderReportType::OPG102;
+        $courtOrderReportType = CourtOrderReportType::tryFrom('OPG' . substr($value, 0, 3));
+
+        if ($courtOrderReportType === null) {
+            return null;
+        }
 
         $courtOrderType = match ($courtOrderReportType) {
             CourtOrderReportType::OPG102, CourtOrderReportType::OPG103 => CourtOrderType::PFA,
--- a/api/app/src/Domain/Report/ReportType.php
+++ b/api/app/src/Domain/Report/ReportType.php
@@ -30,7 +30,7 @@ final readonly class ReportType implements \Stringable
         $parts[] = '10' . $suffix;
 
         if ($this->courtOrderKind === CourtOrderKind::Hybrid) {
-            $parts[] = CourtOrderReportType::OPG104->getSuffix();
+            $parts[] = substr(CourtOrderReportType::OPG104->getSuffix(), -1);
         }
 
         if ($this->deputyType !== DeputyType::LAY) {
@@ -38,5 +38,27 @@ final readonly class ReportType implements \Stringable
         }
 
         return implode('-', $parts);
+    }
+
+    public static function tryFrom(string $value): ReportType
+    {
+        $courtOrderReportType = CourtOrderReportType::tryFrom('OPG' . substr($value, 0, 3)) ??
+            CourtOrderReportType::OPG102;
+
+        $courtOrderType = match ($courtOrderReportType) {
+            CourtOrderReportType::OPG102, CourtOrderReportType::OPG103 => CourtOrderType::PFA,
+            CourtOrderReportType::OPG104 => CourtOrderType::HW,
+        };
+
+        // Default to single as we can not determine Dual or Single at this stage
+        $courtOrderKind = str_contains($value, '-4') ? CourtOrderKind::Hybrid : CourtOrderKind::Single;
+
+        $deputyType = match (true) {
+            str_contains($value, '-5') => DeputyType::PRO,
+            str_contains($value, '-6') => DeputyType::PA,
+            default => DeputyType::LAY,
+        };
+
+        return new ReportType($courtOrderReportType, $courtOrderType, $courtOrderKind, $deputyType);
     }
 }
--- a/api/app/src/Entity/Report/Report.php
+++ b/api/app/src/Entity/Report/Report.php
@@ -10,7 +10,9 @@ use OPG\Digideps\Backend\Entity\User;
 use OPG\Digideps\Backend\Service\ReportService;
 use OPG\Digideps\Backend\Service\ReportStatusService;
 use Doctrine\Common\Collections\ArrayCollection;
+use Doctrine\Common\Collections\Collection;
 use Doctrine\ORM\Mapping as ORM;
+use Doctrine\ORM\PersistentCollection;
 use JMS\Serializer\Annotation as JMS;
 
 /**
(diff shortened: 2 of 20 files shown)


api/app/src/Entity/Report/Report.php
```diff
@@ -14,9 +15,12 @@
 use DateTime;
 use DateTimeInterface;
 use Doctrine\Common\Collections\ArrayCollection;
+use Doctrine\Common\Collections\Collection;
 use Doctrine\ORM\Mapping as ORM;
+use Doctrine\ORM\PersistentCollection;
 use JMS\Serializer\Annotation as JMS;
 use phpDocumentor\Reflection\Types\This;
+use Ramsey\Collection\CollectionInterface;
```
olivier-nicodemi: We don't seem to need this (nor DateTime, Doctrine\ORM\PersistentCollection, or phpDocumentor\Reflection\Types\This for that matter)

api/app/src/Entity/Report/Report.php
```diff
@@ -1655,4 +1660,12 @@ public function isHw(): bool
             ]
         );
     }
+
+    /**
+     * @return Collection<int, CourtOrder>
+     */
+    public function getCourtOrders(): Collection
```
olivier-nicodemi, line 1653: Do we expect this to ever not be an ArrayCollection? Or is there a concrete benefit here of keeping the type vague?

api/app/src/Service/ReportTypeService.php
```diff
@@ -0,0 +1,29 @@
...
+            array_map(function (CourtOrder $courtOrder) {
+                return $courtOrder->getDesiredReportType();
+            }, $courtOrders)
+        );
+
+        if (count($reportTypes) > 1) {
+            return null;
+        }
+
+        return $reportTypes[0];
```
olivier-nicodemi, line 31: What happens if `count($reportTypes) === 0`? (I.e. an empty array was passed in). If you just want to also return `null`, then it should be  `$reportTypes[0] ?? null`.

api/app/src/Domain/Report/ReportType.php
```diff
@@ -39,4 +39,26 @@ public function __toString(): string
 
         return implode('-', $parts);
     }
+
+    public static function tryFrom(string $value): ReportType
```
olivier-nicodemi: The precedent if following PHP's enums is `::from(): T //throws on error` and `tryFrom(): ?T //null on error`. This currently always returns a non null value even if absolute garbage is passed in. Which may or may not be a good thing.

api/app/src/Domain/Report/ReportType.php
```diff
@@ -30,7 +30,7 @@ public function __toString(): string
         }
 
         if ($this->courtOrderKind === CourtOrderKind::Hybrid) {
-            $parts[] = CourtOrderReportType::OPG104->getSuffix();
+            $parts[] = substr(CourtOrderReportType::OPG104->getSuffix(), -1);
```
olivier-nicodemi: Since `->getSuffix()` only exists to support this function, it may be worth considering to only return `'2' | '3'  | '4'` from `CourtOrderReportType::getSuffix` and prefixing a `'10'` at the end here before returning the imploded string.

olivier-nicodemi (changes requested):

api/app/src/Entity/Report/Report.php
Raffers: Doctrine, when retrieving this will set it as `PersistentCollection` We need to set it as an `ArrayCollection` so the class is correctly initialized when retrieving the record.

api/app/src/Domain/Report/ReportType.php
Raffers: It's currently only supporting this function, but there are a few other places that would look at this value. So long as we're mindful of updating these in future I think we should keep the current implementation

Also I had assumed this was a bug. As otherwise when we have a hybrid we would end up with a report type of `xxx-104` where we would expect `xxx-4`

olivier-nicodemi: It was a bug, but as it is brand new it still has a malleability it might not have in the future. Keep in mind that `CourtOrderReportType::OPG104->value` exists in a pinch for other places

api/app/src/v2/Registration/DeputyshipProcessing/Report/ReportTypeUpdate.php
```diff
@@ -0,0 +1,109 @@
...
+
+    public function run(): DataFactoryResult
+    {
+        $count = 0;
+        $errors = [];
+        foreach ($this->getChangedReports() as $report) {
+            /** @var CourtOrder[] $courtOrders */
+            $courtOrders = $report->getCourtOrders()->toArray();
+
+            /** @var ReportType $currentReportType */
```
olivier-nicodemi: Lying to PHPStan to shut it up is exactly not what one should do. If the existing one is garbage this will throw latter on. I recommend handling this on line 93

olivier-nicodemi (changes requested):

api/app/src/v2/Registration/DeputyshipProcessing/Report/ReportTypeUpdate.php
Raffers: If we're getting a `null` from `ReportType::tryFrom` at this point, then we have much larger issues with the system overall. But sure.

olivier-nicodemi (dismissed):

townxelliot:
Waiting for dry-run flag and testing on preprod.

api/app/src/v2/Registration/DeputyshipProcessing/Report/ReportTypeUpdate.php
```diff
@@ -0,0 +1,118 @@
...
+
+        foreach ($reportIds->toIterable() as $reportId) {
+            /** @var Report $report */
+            $report = $this->reportRepository->find($reportId);
+
+            yield $report;
+        }
+    }
+
+    public function run(bool $dryRun = false): DataFactoryResult
```
olivier-nicodemi: No need for a default

api/app/src/v2/Registration/DeputyshipProcessing/Report/ReportTypeUpdate.php
```diff
@@ -0,0 +1,118 @@
...
+use OPG\Digideps\Backend\Factory\DataFactoryResult;
+use OPG\Digideps\Backend\Repository\ReportRepository;
+use OPG\Digideps\Backend\Service\ReportTypeService;
+use OPG\Digideps\Backend\v2\Registration\Enum\DeputyshipCandidateAction;
+use Doctrine\ORM\EntityManagerInterface;
+use Doctrine\ORM\Query\QueryException;
+use Doctrine\ORM\Query\ResultSetMappingBuilder;
+use Psr\Log\LoggerInterface;
+
+class ReportTypeUpdate
```
olivier-nicodemi: Should there not be a `implements DataFactoryInterface`?

api/app/src/Factory/ReportTypeUpdateFactory.php
```diff
@@ -0,0 +1,118 @@
...
+            }
+
+            if (
+                $currentReportType !== null && $currentReportType->courtOrderKind === CourtOrderKind::Hybrid ||
+                $possibleReportType->courtOrderKind === CourtOrderKind::Hybrid
+            ) {
+                $errors[] = 'Possible dangerous change to or from Hybrid on report: ' . $reportId;
+            }
+
+            if (!$dryRun) {
```
olivier-nicodemi, line 101: Should we not consider logging independent of whether this is a dry run or not?

api/app/src/v2/Registration/DeputyshipProcessing/Report/ReportTypeUpdate.php
```diff
@@ -0,0 +1,118 @@
+<?php
+
+declare(strict_types=1);
+
+namespace OPG\Digideps\Backend\v2\Registration\DeputyshipProcessing\Report;
```
olivier-nicodemi: Should this maybe be in `OPG\Digideps\Backend\Factory`?

olivier-nicodemi (dismissed):

api/app/src/Factory/ReportTypeUpdateFactory.php
Raffers: Yes, their should be as we have no other logging around any of the post/pre dataFactory processing. The only reason I've currently implemented this, is to prove to product that this has worked correctly & provide them with information to double check themselves.

olivier-nicodemi (approved):

PR #2502, opened 2026-04-20
DDLS-1395 Fix rendering of required/optional fields: form_known_date

iqpalm:
## Purpose

Fixes DDLS-1395

## Approach

## Learning

## Checklist

### Frontend


client/app/src/Twig/FormFieldsExtension.php
```diff
@@ -403,6 +375,91 @@ private function getFormComponentTwigVariables(FormView $element, string $elemen
...
+     *
+     * @param string $translationKey The translation key prefix
+     * @param string $domain The translation domain
+     * @param string|null $customHint Optional custom hint text to use instead of translation
+     * @return string The hint text or default date hint if no translation found
+     */
+    private function getDateHintText(string $translationKey, string $domain, ?string $customHint = null): string
+    {
+        // Use custom hint if provided
+        if (!empty($customHint)) {
```
olivier-nicodemi: This will prevent an explicit `''` to remove the hint. Consider testing for `null` instead unless this is intentional. In which case consider making the parameter non-nullable and using an empty string as sentinel for "no custom hint". (That empty string is special should be called out in the doc's either in the latter case or if you do not change the functionality.)

client/app/src/Twig/FormFieldsExtension.php
```diff
@@ -403,6 +375,91 @@ private function getFormComponentTwigVariables(FormView $element, string $elemen
...
+    }
+
+    /**
+     * Extract translation key and domain from FormView and parameters.
+     * Handles optional transaction index for nested translations.
+     *
+     * @param FormView $element The form element
+     * @param string $elementName The element name
+     * @param int|null $transIndex Optional transaction index for nested translations
+     * @return array Array containing [translationKey, domain]
```
olivier-nicodemi: `@return array{string, string} containing [translationKey, domain]`

olivier-nicodemi: Or even better type as `array{'translationKey': string, 'domain': string}`
Return as `return ['translationKey' => $translationKey, 'domain' => $domain];`
And destructure as ['translationKey' => $translationKey, 'domain' => $domain] = ...` on line 179

client/app/templates/Components/Form/_known-date.html.twig
```diff
@@ -1,31 +1,39 @@
...
+
 <div id="form-group-{{ element.vars.id }}" class="govuk-form-group {% if not element.vars.valid %}govuk-form-group--error{% endif %}">
     <fieldset class="govuk-fieldset" aria-describedby="{{ element.vars.id }}-hint">
         <legend class="govuk-fieldset__legend {{ legend.isPageHeading ? "govuk-fieldset__legend--xl" : "govuk-fieldset__legend--m" }} {% if legendClass is defined %} {{legendClass}}{% endif %}">
             {% if legend.isPageHeading %}
                 {% if legend.caption %}
                     <span class="govuk-caption-xl">{{ legend.caption }}</span>
                 {% endif %}
-                <h1 class="govuk-fieldset__heading">{{ legend.text }}</h1>
+                <h1 class="govuk-fieldset__heading">{{ legend.text }}{% if not required %} {{ optionalLabel }}{% endif %}</h1>
```
olivier-nicodemi: if required then optionalLabel is `''`. Is the extra if really need ed here and at line 16?

If legend text is not used anywhere else on its own, consider updating it  (in a new variable) at the beginning of the file instead of creating  optionalLabel

client/app/src/Twig/FormFieldsExtension.php
```diff
@@ -387,7 +387,7 @@ private function getFormComponentTwigVariables(FormView $element, string $elemen
     private function getDateHintText(string $translationKey, string $domain, ?string $customHint = null): string
     {
         // Use custom hint if provided
-        if (!empty($customHint)) {
+        if (null !== ($customHint)) {
```
olivier-nicodemi: Superfluous parentheses

olivier-nicodemi (dismissed):

jamesrwarren (dismissed):

olivier-nicodemi (changes requested):
Why are the client-unit-tests committed? They would have been floating around because they where ignored while the folder structure changed

olivier-nicodemi (approved):

PR #2530, opened 2026-04-30
DDLS-1392 Fix rendering of required/optional fields: form_select

iqpalm:
## Purpose

Fixes DDLS-1392

## Approach

## Learning

## Checklist

### Frontend


client/app/src/Form/Report/MoneyTransactionType.php
```diff
@@ -34,11 +35,14 @@ class MoneyTransactionType extends AbstractType
...
      * @return array where keys and values are the categoriesID. e.g. [broadband=>null, fees=>null]
      */
-    private function getCategories()
+    private function getCategories(): array
     {
         $ret = [];
 
-        foreach (MoneyTransaction::$categories as $cat) {
+        /** @var array[] $categories */
+        $categories = MoneyTransaction::$categories;
```
olivier-nicodemi, line 30: It is usually preferable to fix type issues at the source. I.e. `public static array $categories = ...`

You might even consider putting the shape of the array in an annotation there. I believe that would be `@var array<array{string, bool, string, string}> $categories` but double check that.

client/app/src/Form/Report/MoneyTransactionType.php
```diff
@@ -34,11 +35,14 @@ class MoneyTransactionType extends AbstractType
...
+        foreach ($categories as $cat) {
+            /** @var int $categoryId */
             $categoryId = $cat[0];
```
olivier-nicodemi, line 33: Are we sure that is an int?

client/app/src/Form/Report/MoneyTransactionType.php
```diff
@@ -58,23 +62,36 @@ private function getCategories()
     /**
      * @return bool
      */
-    private function isDescriptionMandatory()
+    private function isDescriptionMandatory(): bool
     {
-        foreach (MoneyTransaction::$categories as $row) {
+        /** @var array[] $categories */
+        $categories = MoneyTransaction::$categories;
```
olivier-nicodemi, line 54: See above

client/app/src/Form/Report/MoneyTransactionType.php
```diff
@@ -58,23 +62,36 @@ private function getCategories()
...
     }
 
     public function buildForm(FormBuilderInterface $builder, array $options)
     {
-        $this->authorizationChecker = $options['authChecker'];
-        $this->step = (int) $options['step'];
-        $this->type = $options['type'];
-        $this->selectedCategory = $options['selectedCategory'];
+        /** @var int $step */
+        $step = $options['step'];
```
olivier-nicodemi: Careful. You dropped the cast here which may be needed

olivier-nicodemi (changes requested):

client/app/src/Form/Report/MoneyTransactionType.php
iqpalm: Changed to string

client/app/src/Form/Report/MoneyTransactionType.php
iqpalm: I replaced the cast with the @var

olivier-nicodemi: The var tells PHPStan what the type allegedly is. The cast actually changes the type (unless it already was the type).

These are not the same.

If `$options['step']` was a `string` then you now have a `string` in `$step` masquerading as `int`. Your change is only correct if you are absolutely certain that `$options['step']` is always already an `int`

iqpalm: Not sure I get what you mean but added the cast back in

olivier-nicodemi (dismissed):

olivier-nicodemi (approved):

PR #2535, opened 2026-05-01
Remove ReportInterface from client and make self::$entitymanager not nullable in ApiTestCase (extra cleanup following DDLS-1316 and DDLS-641)

olivier-nicodemi:


townxelliot:
@olivier-nicodemi Which ticket does this relate to?

olivier-nicodemi:
@townxelliot 

1st commit is something I should on reflection have done in [DDLS-](https://opgtransform.atlassian.net/browse/DDLS-1316 as part of reconfiguring PHPStan because it causes trouble
2nd commit is what I raised in the stand-up.
3rd commit relates to https://opgtransform.atlassian.net/browse/DDLS-641
4th commit, I declare it out of scope to have to fix the typing here.

townxelliot (approved):

PR #2551, opened 2026-05-06
DDLS-1284

olivier-nicodemi:
## Purpose

Fixes DDLS-1284

## Approach

## Learning

## Checklist

### Frontend

### DIFF
--- a/api/app/src/Entity/Deputy.php
+++ b/api/app/src/Entity/Deputy.php
@@ -51,6 +51,10 @@ class Deputy
     #[ORM\Column(name: 'deputy_type', type: 'string', length: 3, nullable: false)]
     private string $deputyType;
 
+    #[ORM\ManyToOne(targetEntity: Organisation::class)]
+    #[ORM\JoinColumn(onDelete: "SET NULL")]
+    private ?Organisation $organisation;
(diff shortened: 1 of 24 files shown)


api/app/src/Entity/Deputy.php
```diff
@@ -51,6 +51,10 @@ class Deputy
     #[ORM\Column(name: 'deputy_type', type: 'string', length: 3, nullable: false)]
     private string $deputyType;
 
+    #[ORM\ManyToOne(targetEntity: Organisation::class)]
+    #[ORM\JoinColumn(onDelete: "SET NULL")]
+    private ?Organisation $organisation;
```
townxelliot, line 56: Would it be safer to set this to null by default, in case someone calls getOrganisation() before it has a value?

townxelliot (approved):
I made a small comment but I don't think it makes enough of a difference to block merging it.

api/app/src/Entity/Deputy.php
olivier-nicodemi: That is not behaviour we should encourage. In a better world this would be a mandatory constructor parameter.

PR #2660, opened 2026-06-03
DDLS-1472 DoctrineListener

olivier-nicodemi:
First in Series DDLS-1472


api/app/src/Entity/Report/Traits/AssetTrait.php
```diff
@@ -50,7 +50,7 @@ public function removeAsset(Asset $assets)
     }
 
     /**
-     * @return Collection<int, AssetInterface>|AssetInterface[]
+     * @return Collection<int, AssetInterface>
      */
     public function getAssets(): Collection|array
```
olivier-nicodemi, line 55: Intentionally not correcting this in this PR

api/app/src/Entity/Report/Traits/DebtTrait.php
```diff
@@ -4,19 +4,20 @@
...
 {
     /**
-     * @var Debt[]
+     * @var Collection<int, Debt>
      */
     #[JMS\Groups(['debt'])]
     #[ORM\OneToMany(mappedBy: 'report', targetEntity: Debt::class, cascade: ['persist', 'remove'])]
     #[ORM\OrderBy(['id' => 'ASC'])]
-    private $debts;
+    private Collection $debts;
```
townxelliot: do we need to think about initialising this somehow?

api/app/src/Entity/Report/Traits/DebtTrait.php
```diff
@@ -71,13 +72,13 @@ public function setDebtManagement($debtManagement)
...
     public function getDebtByTypeId($typeId)
     {
         return $this->getDebts()->filter(function (Debt $debt) use ($typeId): bool {
-            return $debt->getDebtTypeId() == $typeId;
-        })->first();
+            return $debt->getDebtTypeId() === $typeId;
+        })->first() ?: null;
     }
 
     public function addDebt(Debt $debt)
```
townxelliot, line 84: ...otherwise this method will fail unless the Collection has been initialised already.

api/app/src/Entity/Report/Contact.php
```diff
@@ -350,4 +350,12 @@ public function getCountry()
     {
         return $this->country;
     }
+
+    #[ORM\PreRemove]
+    public function onPreRemove(PreRemoveEventArgs $_): void
+    {
+        if ($this->getReport()->getContacts()->count() === 1) {
```
townxelliot, line 357: Please remind me why we are no longer comparing this to 0 and setting the field to "no" if it is 0. Same goes for other fields which follow this pattern.

I think I get why it would be nulled rather than set to "no", but I'm not sure why the comparison is with 1 rather than 0.

townxelliot (changes requested):
I'd like some clarification about the change to setting a value to "no" when count === 0,  to setting it to null when count === 1. Apart from that, lgtm.

api/app/src/Entity/Report/Traits/DebtTrait.php
olivier-nicodemi: Yes we will need to initialise it when we type it. That shouldn't have been in this PR yet

api/app/src/Entity/Report/Traits/DebtTrait.php
olivier-nicodemi: Reverted this for now

townxelliot (approved):

PR #2670, opened 2026-06-05
DDLS-1318

olivier-nicodemi:
DDLS-1318

### DIFF
--- a/api/app/src/Controller/Report/ReportController.php
+++ b/api/app/src/Controller/Report/ReportController.php
@@ -329,10 +329,6 @@ class ReportController extends RestController
             ]);
         }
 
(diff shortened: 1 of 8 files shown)


api/app/src/Controller/Report/ReportController.php
```diff
@@ -344,6 +340,10 @@ public function update(Request $request, int $id): array
             $report->updateDueDateBasedOnEndDate();
         }
 
+        if (array_key_exists('due_date', $data)) {
```
olivier-nicodemi, line 343: If due date is set, then we don't wan't to immediately overwrite it based on end date?

townxelliot (approved):

PR #2675, opened 2026-06-08
Remove superfluous package

olivier-nicodemi:


olivier-nicodemi:
:( We actually "need" it

PR #2678, opened 2026-06-09
DDLS-1472 api - entities

olivier-nicodemi:
4th in the DDLS-1472 series

### DIFF
--- a/api/app/src/Entity/Deputy.php
+++ b/api/app/src/Entity/Deputy.php
@@ -27,7 +27,7 @@ class Deputy
     #[ORM\Id]
     #[ORM\GeneratedValue(strategy: 'IDENTITY')]
     #[ORM\SequenceGenerator(sequenceName: 'deputy_id_seq', allocationSize: 1, initialValue: 1)]
-    private int $id;
+    private ?int $id = null;
 
     /**
      * Holds the deputy the client belongs to
@@ -53,7 +53,7 @@ class Deputy
 
     #[ORM\ManyToOne(targetEntity: Organisation::class)]
     #[ORM\JoinColumn(onDelete: "SET NULL")]
-    private ?Organisation $organisation;
+    private ?Organisation $organisation = null;
 
     #[JMS\Type('string')]
     #[JMS\Groups(['report-submitted-by', 'deputy'])]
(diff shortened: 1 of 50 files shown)


api/app/src/Entity/Report/Traits/AssetTrait.php
```diff
@@ -4,98 +4,71 @@
...
 
-    /**
-     * Add assets.
-     *
-     * @return Report
-     */
-    public function addAsset(Asset $assets)
+    public function addAsset(Asset $assets): static
     {
         $this->assets[] = $assets;
```
townxelliot, line 32: Does this work on Collection objects?

api/app/src/Entity/Report/Traits/ContactTrait.php
```diff
@@ -12,69 +12,50 @@
...
 
-    /**
-     * Add contacts.
-     *
-     * @return Report
-     */
-    public function addContact(Contact $contacts)
+    public function addContact(Contact $contacts): Report
     {
         $this->contacts[] = $contacts;
```
townxelliot, line 33: Will this work now it's a Collection?

api/app/src/Entity/Report/Traits/StatusTrait.php
```diff
@@ -14,52 +14,44 @@ trait StatusTrait
...
      * Set by endpoints hooks on section CRUD operations that call `updateSectionsStatusCache` below
      *
      * Note: Manually Json-serialised. `json_array` type not working properly in the unity of work in this doctrine version
-     *
-     * @var string
      */
     #[JMS\Exclude]
     #[ORM\Column(name: 'status_cached', type: 'text', nullable: true)]
-    private $sectionStatusesCached;
+    private ?string $sectionStatusesCached;
```
townxelliot, line 25: Would it be safer to set this to null by default?

api/app/src/Entity/Deputy.php
```diff
@@ -137,19 +137,34 @@ class Deputy
     #[ORM\OneToMany(mappedBy: 'deputy', targetEntity: CourtOrderDeputy::class, cascade: ['persist', 'remove'])]
     private Collection $courtOrderDeputyRelationships;
 
-    public function __construct()
-    {
+    public function __construct(
```
townxelliot, line 140: Are there other places where we're doing `new Deputy()` which will break if this constructor changes signature?

townxelliot (approved):
Looks like you've fixed all the `new Deputy()` calls, so ignore that comment if so.

I had a couple of questions, but happy for it to be merged as-is.

api/app/src/Entity/Report/Traits/AssetTrait.php
olivier-nicodemi: Collection implements ArrayAccess

api/app/src/Entity/Report/Traits/ContactTrait.php
olivier-nicodemi: Collection implements ArrayAccess

api/app/src/Entity/Report/Traits/StatusTrait.php
olivier-nicodemi: It is set in the constructor

api/app/src/Entity/Deputy.php
olivier-nicodemi: Went through all of them (I hope)

PR #2684, opened 2026-06-10
DDLS-1495 "Application error" when attempting to submit report

iqpalm:
## Purpose
Multiple users have reported receiving an “Application error” when attempting to submit their report. 
All the cases are attempting to re-submit incomplete reports with expired documents attached to them.

Fixes DDLS-1495

## Approach
The error is an unique violation on court_order_report table where the incomplete report when submitted is trying to create next year's report which already exists.

The solution is check to see if the report being submitted is an incomplete one and therefore do not add it to the court order again.

## Learning

## Checklist

### Frontend

### DIFF
--- a/api/app/src/Service/ReportService.php
+++ b/api/app/src/Service/ReportService.php
@@ -89,17 +89,12 @@ class ReportService
             if (!is_null($newYearReport)) {
                 $this->clonePersistentResources($newYearReport, $currentReport);
             }
-
-            $addReportToCourtOrder = false;
         } else {
             // first-time submission
             $this->logger->warning("Creating next year report for client $clientId (NO existing report) at $now");
 
             $newYearReport = $this->createNextYearReport($currentReport);
-            $addReportToCourtOrder = true;
-        }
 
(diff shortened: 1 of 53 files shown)


api/app/src/Service/ReportService.php
```diff
@@ -89,14 +89,17 @@ public function submit(Report $currentReport, User $user, \DateTime $submitDate)
...
         } else {
             // first-time submission
             $this->logger->warning("Creating next year report for client $clientId (NO existing report) at $now");
 
             $newYearReport = $this->createNextYearReport($currentReport);
+            $addReportToCourtOrder = true;
         }
 
-        if (!is_null($newYearReport)) {
+        if (!is_null($newYearReport) && $addReportToCourtOrder) {
```
Raffers: If this is only needed in the else statement, you can move this if block to there. You can also remove the `null` check as within the else statement `ReportService::createNextYearReport` will always return a report object

Raffers (changes requested):

Raffers (approved):

PR #2685, opened 2026-06-10
DDLS-1496 prototype for playwright end to end tests

jamesrwarren:
After discussions, converted from python version to typescript.

The readme gives a good explanation of the change.

Some future work:
- although there's nothing that runs this in prod, we should probably selectively not create the task definitions in prod for this or some of our other images. Will do that as a separate little PR for all our non prod task definitions.
- We could probably make the image a lot more secure. We don't run as root but slim images are pretty awful. It was suggested that it might not work well with alpine though so would have to try and imrove this as another PR.
- Doing single threaded. It's just config to update but we need to work out what best settings are when we have more tests.


playwright/pages/login_page.py
```diff
@@ -0,0 +1,23 @@
+import re
```
townxelliot: I'm on the fence about whether we should use classes, or just write tests as a series of function calls.

playwright/Dockerfile
```diff
@@ -0,0 +1,40 @@
...
+    libasound2 \
+    libpangocairo-1.0-0 \
+    libatk1.0-0 \
+    libcups2 \
+    libdrm2 \
+    --no-install-recommends \
+    && rm -rf /var/lib/apt/lists/*
+
+# Install Python deps
+RUN pip install --no-cache-dir \
```
townxelliot: I'd quite like the deps in a requirements.txt file, so I could easily install them locally.

jamesrwarren: Oh yeah they definitely should be. In fact I have a requirements.txt, I just didn't bother installing from it for the purposes of this prototype.

playwright/pages/login_page.py
jamesrwarren: Me too. We should have conversation about it after standup. Just going to get the aws hardware bit working for now. Not going to work on any further test internals till we make some decisions on things.

playwright/tests/fixtures/fixtures.ts
```diff
@@ -0,0 +1,25 @@
+export const testPassword = "DigidepsPass1234";
```
olivier-nicodemi: Why are we exporting the constants?

playwright/tests/pages/LoginPage.ts
```diff
@@ -0,0 +1,23 @@
+import { Page, expect } from "@playwright/test";
+
+export class LoginPage {
+  constructor(private page: Page) {}
```
olivier-nicodemi, line 6: If you want privacy use `#` (See https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes/Private_elements)

playwright/tests/pages/login-page.ts
```diff
@@ -0,0 +1,23 @@
+import { Page, expect } from "@playwright/test";
+
+export class LoginPage {
```
olivier-nicodemi: We probably should have a discussion with the team if this file should be LoginPage.ts and the export a `default export`

playwright/eslint.config.mjs
```diff
@@ -0,0 +1,12 @@
+import { defineConfig, globalIgnores } from "eslint/config";
+import tseslint from "typescript-eslint";
+import globals from "globals";
+
+export default defineConfig([
```
olivier-nicodemi, line 5:
```ts
export default defineConfig([
  globalIgnores(["**/playwright-report/"]),
  {
    files: ["**/*.ts", "**/*.cts", "**/*.mts"],
    languageOptions: {
      globals: globals.browser,
      parserOptions: {
        projectService: true,
      },
    },
    extends: [tseslint.configs.strictTypeChecked],
  },
]);
```

playwright/tests/fixtures/fixtures.ts
```diff
@@ -0,0 +1,25 @@
...
+  },
+};
+
+export function createFixture(type: keyof typeof fixtureUsers) {
+  // A call to the fixtures API that would create the fixture and return details
+  // like login email goes here.
+  const user = fixtureUsers[type];
+
+  if (!user) {
+    throw new Error(`Unknown fixture_type: ${type}`);
```
olivier-nicodemi: Dead code

playwright/package.json
```diff
@@ -0,0 +1,22 @@
+{
+  "name": "opg-digideps-end-to-end-tests",
+  "version": "1.0.0",
+  "description": "End-to-end tests for Digideps: Managed by opg-org-infra &amp; Terraform",
+  "scripts": {
+    "lint": "eslint .",
+    "check-format": "prettier --check .",
+    "format": "prettier --write .",
```
olivier-nicodemi, line 9: Opinionated: Eslint can also do the formatting if configured accordingly.

playwright/tsconfig.json
```diff
@@ -0,0 +1,10 @@
+{
+  "compilerOptions": {
```
olivier-nicodemi, line 2: This configuration feels... archaic.

playwright/playwright.config.ts
```diff
@@ -0,0 +1,13 @@
+import { defineConfig } from "@playwright/test";
+
+export default defineConfig({
+  testDir: "./tests",
+  retries: 1,
+  reporter: [["html", { outputFolder: "playwright-report", open: "never" }]],
+  use: {
+    baseURL: process.env.FRONT_URL || "http://frontend-webserver",
```
olivier-nicodemi: Not a fan of hardcoding the container config

playwright/package.json
```diff
@@ -0,0 +1,22 @@
...
+    "show-report": "playwright show-report --host 0.0.0.0"
+  },
+  "devDependencies": {
+    "@eslint/js": "^10.0.1",
+    "@playwright/test": "^1.58.2",
+    "@types/node": "^25.5.0",
+    "eslint": "^10.0.3",
+    "globals": "^17.4.0",
+    "prettier": "3.8.3",
+    "typescript-eslint": "^8.57.1"
```
olivier-nicodemi: I would expect `tsc` as direct dependency

playwright/tests/pages/login-page.ts
```diff
@@ -0,0 +1,23 @@
...
+  async goto() {
+    await this.page.goto("/login");
+  }
+
+  async login(email: string, password: string) {
+    await this.page.fill("#login_email", email);
+    await this.page.fill("#login_password", password);
+
+    await Promise.all([
+      this.page.waitForNavigation(),
```
olivier-nicodemi: waitForNavigation is deprecated see its doc string for replacment

olivier-nicodemi (changes requested):
I generally prefer as much as possible to be explicitly typed. (And those types to be declared since in larger projects they would be reused)

E.g.
```ts
export type UserFixture = Readonly<{email: string, password: string}>;

export function createFixture(type: keyof typeof fixtureUsers): UserFixture {
    //...
}
```

playwright/package.json
```diff
@@ -0,0 +1,22 @@
+{
+  "name": "opg-digideps-end-to-end-tests",
+  "version": "1.0.0",
+  "description": "End-to-end tests for Digideps: Managed by opg-org-infra &amp; Terraform",
+  "scripts": {
```
olivier-nicodemi, line 5: Missing `"check-type": "tsc --noEmit",`

playwright/tests/fixtures/fixtures.ts
jamesrwarren: Fair point

playwright/tests/pages/LoginPage.ts
jamesrwarren: Read that but not sure if it needs the additional privacy. One for the team to talk about. Maybe we need a standard doc

playwright/tests/pages/login-page.ts
jamesrwarren: Changed it but yeah again as long as we're consistent. Worth a quick chat and setting some standards

playwright/eslint.config.mjs
jamesrwarren: Yeah strict type checking seems like a good idea. Changed.

playwright/package.json
jamesrwarren: Fair point but prettier separates concerns and less config. Does the eslint one also correct the formatting? I don't mind either way. Haven't changed it in this PR but feel free to if you prefer.

playwright/tsconfig.json
jamesrwarren: I've had a look and changed some of the settings to be more apt as I had some that shouldn't be being used from what I can tell but otherwise you might have to update it to they config you believe it should be.

playwright/playwright.config.ts
jamesrwarren: Yeah it was a bit hacky. Will just pass it in via the docker env var.

playwright/package.json
jamesrwarren: from what I can tell it's just typescript that we need? Added that (though it was dependency of eslint) and I've added a typecheck in the npm commands too which In will put in the pipeline

playwright/package.json
jamesrwarren: added

playwright/tests/pages/login-page.ts
jamesrwarren: Good point

playwright/tests/fixtures/fixtures.ts
jamesrwarren: yeah maybe it's me not understanding how it would react with key lookup. removed.

playwright/tests/pages/LoginPage.ts
olivier-nicodemi: More of a since js now has a real private why are we using the "old" typescript way which does not enforce anything at runtime

olivier-nicodemi (dismissed):

olivier-nicodemi (approved):

PR #2699, opened 2026-06-12
DDLS-1510 Remove jwt-core as a direct dependency

townxelliot:
## Purpose

I have retained jwt-framework (which has jwt-core as a component), as it's an up-to-date library (it's just that using jwt-core as a direct dependency is deprecated). I considered using lcobucci/jwt but it doesn't support RSA signatures without some faff, which is what we're using.

I then removed lcobucci/jwt from the client side of the code.

We still have lcobucci/jwt on the API side, but it's less of a problem there as we are generating signatures, not trying to read them, and that seems to work OK.

Fixes DDLS-1510

## Approach

## Learning

## Checklist

### Frontend


api/app/composer.json
```diff
@@ -72,6 +72,7 @@
         "symfony/yaml": "^6.4.41"
     },
     "require-dev": {
+        "ext-openssl": "*",
```
olivier-nicodemi: I think that goes in (prod) require

client/app/composer.json
```diff
@@ -31,8 +31,8 @@
     "require": {
         "ext-dom": "*",
         "ext-zip": "*",
-        "alphagov/notifications-php-client": "^7.0.0",
-        "aws/aws-sdk-php": "^3.384.10",
+        "alphagov/notifications-php-client": "^7.1.0",
+        "aws/aws-sdk-php": "^3.379.3",
```
olivier-nicodemi: This looks backwards

client/app/composer.json
```diff
@@ -112,6 +112,7 @@
         }
     ],
     "require-dev": {
+        "ext-openssl": "*",
```
olivier-nicodemi: As in api

client/app/composer.json
townxelliot: Good spot.

client/app/composer.json
townxelliot: This is an interesting one, as we only use the openssl functions directly in the tests. They are likely to be referred to from other packages, but I'm not sure if that requirement will be enforced if it's a transitive requirement of a third party package.

olivier-nicodemi (dismissed):

olivier-nicodemi (approved):

PR #2714, opened 2026-06-18
DDLS-1494

olivier-nicodemi:
Fixes DDLS-1494


api/app/src/Entity/Report/BankAccount.php
```diff
@@ -32,356 +32,232 @@ class BankAccount implements BankAccountInterface
...
+        return $this->id ?? 0;
     }
 
-    /**
-     * Set id.
-     *
-     * @return int
-     */
-    public function setId($id)
+    public function setId(int $id): static
```
townxelliot, line 124: What do you think of having some sort of IdTrait so we don't have this repeated code everywhere? I'm not 100% sure traits are a great idea, but would be interested to hear your opinion.

api/app/src/Entity/Setting.php
```diff
@@ -11,99 +11,54 @@
...
-        $this->id = $id;
-        $this->content = $content;
-        $this->enabled = $enabled;
+    public function __construct(
+        #[JMS\Type('string')]
+        #[JMS\Groups(['setting'])]
+        #[ORM\Column(name: 'id', type: 'string', length: 64, nullable: false)]
+        #[ORM\Id]
+        private string $id,
+        #[JMS\Type('string')]
```
townxelliot, line 20: My personal preference would be to have a blank line between lines 19 and 20, to make this more readable.

api/app/src/Entity/User.php
```diff
@@ -1319,12 +1002,14 @@ public function populate(?array $data): static
 
     public function deriveDeputyType(): ?DeputyType
     {
-        if (str_contains($this->roleName, 'LAY')) {
-            return DeputyType::LAY;
-        } elseif (str_contains($this->roleName, 'PROF')) {
-            return DeputyType::PRO;
-        } elseif (str_contains($this->roleName, 'PA')) {
-            return DeputyType::PA;
+        if ($this->roleName !== null) {
```
townxelliot, line 1005: I'd probably do an early return if $this->roleName is null here. Though then you might still need to return null at the end, so maybe inefficient. Maybe a match?

townxelliot (approved):
lgtm, very pleasing to see! Just a few comments/questions; as the feature branch has broken e2e tests at the moment, do you want to hold this until I've fixed them, or fix them yourself?

api/app/src/Entity/Report/BankAccount.php
olivier-nicodemi: I am on the fence on that. I didn't want to introduce yet another trait (we are using them decidedly too much) plus to be kept in mind is that I do not expect this to be ever used outside of tests.

olivier-nicodemi: That said an Id trait would actually be used as a trait is supposed to be and not as a stand in for this mess https://learn.microsoft.com/en-us/dotnet/csharp/programming-guide/classes-and-structs/partial-classes-and-methods like a lot of our traits are.

PR #2730, opened 2026-06-24
DDLS-1533 upgrade the postgres DB

jamesrwarren:
Release note for 15:
https://www.postgresql.org/docs/release/15.0/

Release notes for 16:
https://www.postgresql.org/docs/release/16.0/

Possible risks explored in ticket.

Don't merge as I want to manually do the upgrade on preprod first and check for any issues.


olivier-nicodemi (approved):
Is there a way to have this on preprod and dev for a week or two without having it live?

jamesrwarren:
@olivier-nicodemi yeah that's what this does, it does all envs except production where I would leave it to bed in for a couple of weeks. However, before even merging this bit I want to do the upgrade manually on pre to check for issues with migrations.

PR #2731, opened 2026-06-24
DDLS-1535 add alb access logs bucket

jamesrwarren:
This fullfils PR1 of DDLS-1535

What it does in a line: 
It adds a bucket in line with out bucket requirements that can then be used to take the ALB access logs. 

ALB access logs are interesting in that they go to a bucket rather than cloudwatch. I have set a 13 month time frame as that is our default min for log retention of app logs.

### DIFF
--- a/terraform/account/region/s3_alb_access_logs.tf
+++ b/terraform/account/region/s3_alb_access_logs.tf
@@ -1,0 +1,142 @@
+locals {
+  s3_alb_log_account_names = {
+    "development" : "dev",
+    "preproduction" : "pre",
+    "production" : "prod",
+  }
+  s3_alb_log_account_name    = local.s3_alb_log_account_names[var.account.name]
+  s3_access_log_account_name = var.account.name == "development" ? "dev" : var.account.name
+}
+
+resource "aws_s3_bucket" "alb_access" {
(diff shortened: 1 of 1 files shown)


terraform/account/region/s3_alb_access_logs.tf
```diff
@@ -0,0 +1,142 @@
+locals {
+  s3_alb_log_account_names = {
```
jamesrwarren, line 2: need this because of max bucket lengths (63 chars max)

terraform/account/region/s3_alb_access_logs.tf
```diff
@@ -0,0 +1,142 @@
...
+    "development" : "dev",
+    "preproduction" : "pre",
+    "production" : "prod",
+  }
+  s3_alb_log_account_name    = local.s3_alb_log_account_names[var.account.name]
+  s3_access_log_account_name = var.account.name == "development" ? "dev" : var.account.name
```
jamesrwarren, line 8: I'm guessing this was set this way because of the max bucket lengths too.

olivier-nicodemi (approved):

townxelliot (approved):

PR #2735, opened 2026-06-26
DDLS-1519 - Stop creating reports during registration

Raffers:
## Purpose
The current self-serve registration and admin invite pathways create a report.

However, under the new branch all reports will be created during the ingest, so we just want to remove the create report functionality at go live.

This ticket is to remove unnecessary report creation from the registration paths, because this will be handled by the ingest in future.

Fixes DDLS-1519

## Approach

## Learning

## Checklist

### Frontend


client/app/src/Controller/Report/ReportController.php
```diff
@@ -143,57 +142,6 @@ public function editAction(Request $request, int $reportId): RedirectResponse|ar
...
-        $form->handleRequest($request);
-
-        if ($form->isSubmitted() && $form->isValid()) {
-            /** @var array $data */
-            $data = $form->getData();
-
-            $this->restClient->post('report', $data);
-
-            $user = $this->userApi->getUserWithData();
-            $this->eventDispatcher->dispatch(new RegistrationSucceededEvent($user), RegistrationSucceededEvent::DEPUTY);
```
townxelliot, line 186: Do we need to trigger this event in the previous step if it's no longer being triggered when the report is created? We have some hooks for this event which do various things when it's handled.

Raffers: It already exists - 
https://github.com/ministryofjustice/opg-digideps/blob/50fbec2cef0cc5762fd9ee19f986b3db41dcb9a0/client/app/src/Controller/ClientController.php#L164

PR #2743, opened 2026-06-29
Fix PDF generation of new preview

olivier-nicodemi:
Fixing errors before they are noticed.

Note that there is still something wrong with the css. At least locally.


townxelliot (approved):

olivier-nicodemi:
Will be obsolete with components

PR #2764, opened 2026-07-09
Current smoke and future Stan

olivier-nicodemi:
- calm snyk down
- fix some typings PhpStan complains about if it is aware of the future


jackgoodby (approved):
LGTM

PR #2767, opened 2026-07-15
DDLS-978 Replace our behat mink extension fork and rewrite broken e2e tests with playwright

townxelliot:
## Purpose

Fixes DDLS-978

## Approach

## Learning

## Checklist

### Frontend


api/app/src/Fixture/FixtureService.php
```diff
@@ -134,13 +159,37 @@ public function instantiateScenario(
         ];
     }
 
+    public function addSupportingDocumentWithoutS3Object(Report $report, string $filename): Document
```
olivier-nicodemi: This does not feel like it should be public

api/app/src/Fixture/FixtureService.php
```diff
@@ -280,7 +335,7 @@ private function makeDeputy(DeputyDescriptor $descriptor, Client $client, ?Organ
             ->setAddress3('')
             ->setAddress4('')
             ->setAddress5('')
-            ->setAddressCountry(null)
+            ->setAddressCountry('GB')
```
olivier-nicodemi, line 348: I set it explicitly to `null` for completeness. We seem to only ever set this to 'GB' inside the other fixtures.

api/app/src/Controller/FixtureController.php
```diff
@@ -0,0 +1,233 @@
...
+        // TODO refactor into a ReportsSet on a court order
+        // for each report on each court order, add all uploaded files to it, none with a backing S3 object
+        foreach ($details['orders'] as $orderPair) {
+            foreach (['pfa', 'hw'] as $orderType) {
+                $order = $orderPair[$orderType] ?? null;
+
+                if ($order !== null) {
+                    foreach ($order['reports'] as $report) {
+                        foreach ($supportingDocumentNames as $uploadedFile) {
+                            $this->fixtureService->addSupportingDocumentWithoutS3Object($report, $uploadedFile);
```
olivier-nicodemi: My 50 cents is that `$supportingDocumentNames` should be passed as a field inside `Scenario` and not hot patched afterwards

api/app/src/Fixture/FixtureService.php
```diff
@@ -414,4 +565,6 @@ private function refreshCounter(): void
         $this->entityManager->persist($counter);
         $this->counter = $counter;
     }
+
+
```
olivier-nicodemi, line 584: ?

api/app/src/Controller/FixtureController.php
townxelliot: > My 50 cents is that `$supportingDocumentNames` should be passed as a field inside `Scenario` and not hot patched afterwards

I agree. I think the scenario should have a report set, maybe a document set inside each report. But for the purposes of getting this done, I went for a patch approach. I think I put a todo in for it. Would you be satisfied with a follow-up ticket to extend Scenario so it can take more complex arrangments of reports and documents on those reports?

api/app/src/Controller/FixtureController.php
```diff
@@ -0,0 +1,257 @@
...
+    #[Route('/fixtures/scenarios/laysimple', name: 'fixtures_scenarios_laysimple', methods: ['POST'])]
+    #[IsGranted(attribute: 'ROLE_SUPER_ADMIN')]
+    public function scenarioLaySimple(Request $request): array
+    {
+        $this->checkIfAccessible();
+
+        $payload = new ValidatingArray($request->getPayload()->all());
+
+        $deputyReference = $payload->getStringOrThrow('deputyReference');
+        $reportTypeStr = $payload->getStringOrNull('reportType') ?? '';
```
olivier-nicodemi, line 100: getStringOrDefault

api/app/src/Controller/FixtureController.php
```diff
@@ -0,0 +1,257 @@
...
+    #[IsGranted(attribute: 'ROLE_SUPER_ADMIN')]
+    public function scenarioLayReadyToSubmit(Request $request): array
+    {
+        $this->checkIfAccessible();
+
+        $payload = new ValidatingArray($request->getPayload()->all());
+
+        $deputyReference = $payload->getStringOrThrow('deputyReference');
+        $reportTypeStr = $payload->getStringOrNull('reportType') ?? '';
+        $reportType = CourtOrderReportType::tryFrom($reportTypeStr) ?? CourtOrderReportType::OPG102;
```
olivier-nicodemi, line 134: Undocumented default inconsistent with the previous route where it was documented

api/app/src/Fixture/FixtureService.php
```diff
@@ -260,10 +309,11 @@ private function makeClient(): Client
             ->setCreatedAt(null);
     }
 
-
     private function makeCourtOrder(CourtOrderDescriptor $descriptor, Client $client, CourtOrderType $type): CourtOrder
     {
-        $madeDate = new \DateTime()->sub(new \DateInterval('P6M'))->sub(new \DateInterval("P{$descriptor->submittedReports}Y"));
+        $madeDate = $descriptor->madeDate ??
+            new \DateTime()->sub(new \DateInterval('P6M'))->sub(new \DateInterval("P{$descriptor->submittedReports}Y"));
```
olivier-nicodemi, line 315:
```
$madeDate = ($descriptor->madeDate ??
            new \DateTime()->sub(new \DateInterval('P6M'))
)->sub(new \DateInterval("P{$descriptor->submittedReports}Y"));
```
If there are 5 reports then the order must be older than 5 years in a "normal" scenario

olivier-nicodemi: That said I think I spotted an issue with multiple sequential court orders in a scenario would have the same dates here. But that has nothing to do with this PR

api/app/src/Fixture/FixtureService.php
```diff
@@ -316,9 +366,9 @@ private function makeOrganisation(DeputyDescriptor $descriptor): Organisation
         $organisation = new Organisation()
             ->setId($this->counter->nextInt())
             ->setIsActivated(true)
-            ->setDeletedAt(null)
+            ->setDeletedAt()
             ->setName($descriptor->organisation);
-        $organisation->setEmailIdentifier("@{$organisation->getId()}.{$descriptor->emailDomain}");
+        $organisation->setEmailIdentifier("@{$organisation->getId()}.$descriptor->emailDomain");
```
olivier-nicodemi, line 371: I know the default ide setting suggests making interpolation less readable...

api/app/src/Fixture/FixtureService.php
```diff
@@ -359,13 +409,16 @@ private function makeUser(DeputyDescriptor $descriptor, ?Deputy $deputy = null,
                 UserType::Admin => User::ROLE_ADMIN,
                 UserType::AdminManager => User::ROLE_ADMIN_MANAGER,
                 UserType::SuperAdmin => User::ROLE_SUPER_ADMIN,
+                UserType::PaNamedUser => User::ROLE_PA_NAMED,
```
olivier-nicodemi, line 412: UserType::Deputy + DeputyType::PA
Why the extra UserType?

api/app/src/Fixture/FixtureService.php
```diff
@@ -404,7 +457,102 @@ private function makeReportSubmitted(Report $report, ?User $submitter): void
...
+        $document = new Document($report);
+        $document->setIsReportPdf(true);
+        $document->setCreatedBy($submitter);
+        $document->setStorageReference("dd_doc_{$report->getId()}_" . time());
+        $document->setFileName("DigiRep-{$this->counter->nextString(8)}.pdf");
+        $this->persist($document);
+
+        $reportSubmission = new ReportSubmission($report, $submitter);
+        $reportSubmission->setUuid($this->counter->nextString(20));
+        $this->persist($reportSubmission);
```
olivier-nicodemi, line 470: Do not persist unless you created the object with new (triggers less flushes potentialy). I.e. persist on report is redundant.

And yes I am also not 100% comfortable with doctrines behaviour allowing persist to affect changes which have not happened yet)

api/app/src/Fixture/FixtureService.php
olivier-nicodemi, line 452: Needs a persist. Probably my bug though. (This should not rely on the one in makeReportSubmitted)

api/app/src/Fixture/FixtureService.php
```diff
@@ -404,7 +457,102 @@ private function makeReportSubmitted(Report $report, ?User $submitter): void
...
+        // supporting documents
+        $report->setWishToProvideDocumentation('no');
+
+        $this->persist($action);
+        $this->persist($account);
+        $this->persist($benefitsCheck);
+        $this->persist($visitsCare);
+        $this->persist($mentalCapacity);
+        $this->persist($report);
+        $this->flush();
```
olivier-nicodemi, line 551: You should not need to persist report here (assuming make report calls persist). Not sure if the persist after refresh is needed.

api/app/src/Fixture/FixtureService.php
```diff
@@ -427,4 +575,19 @@ private function refreshCounter(): void
         $this->entityManager->persist($counter);
         $this->counter = $counter;
     }
+
+    private function addSupportingDocumentWithoutS3Object(Report $report, string $filename): void
+    {
+        // required so that the system recognises the report has documents
+        $report->setWishToProvideDocumentation('yes');
+        $this->persist($report);
```
olivier-nicodemi, line 583: Idem to all of the above

api/app/src/Fixture/FixtureService.php
```diff
@@ -427,4 +575,19 @@ private function refreshCounter(): void
...
+
+        $document = new Document($report);
+        $document->setIsReportPdf(false);
+        $document->setCreatedBy($report->getSubmittedBy());
+        $document->setStorageReference("dd_doc_{$report->getId()}_" . time());
+        $document->setFileName($filename);
+        $this->persist($document);
+        $this->flush();
```
olivier-nicodemi, line 591: do not flush explicitly unless you need to

api/app/src/Fixture/UserType.php
```diff
@@ -4,12 +4,16 @@
...
-    case Admin;
-    case AdminManager;
-    case SuperAdmin;
+    case Deputy = User::ROLE_DEPUTY;
+    case OrgAdmin = User::ROLE_ORG_ADMIN;
+    case OrgTeamMember = User::ROLE_ORG_TEAM_MEMBER;
+    case Admin = User::ROLE_ADMIN;
+    case AdminManager = User::ROLE_ADMIN_MANAGER;
+    case SuperAdmin = User::ROLE_SUPER_ADMIN;
+    case PaNamedUser = User::ROLE_PA_NAMED;
```
olivier-nicodemi, line 17: These should not exist. This enum does not intend to mirror the `User::ROLE_*` constants

api/app/src/Fixture/UserType.php
```diff
@@ -4,12 +4,16 @@
...
+    case ProNamedUser = User::ROLE_PROF_NAMED;
```
olivier-nicodemi, line 18: Idem

api/app/tests/Behat/features-v2/report-submission/report-submission.feature
```diff
@@ -36,27 +36,24 @@ Feature: Report submissions dashboard
         And I manually 'archive' the client that has one submitted report
         Then I should see the client row under the Synchronised tab
 
-    @super-admin
+    @super-admin @THIS
```
olivier-nicodemi, line 39: ?

playwright/tests/fixtures/fixtures.ts
```diff
@@ -1,30 +1,137 @@
-type UserType = "lay_user" | "pro_user";
+const apiURL = process.env.API_URL ?? "";
+const clientSecret = process.env.ADMIN_API_CLIENT_SECRET ?? "";
 
-interface FixtureUser {
-  email: string;
+type UserType = "ROLE_DEPUTY" | "ROLE_PROF_NAMED";
```
olivier-nicodemi, line 4: See comments on PHP enum

playwright/tests/fixtures/fixtures.ts
```diff
@@ -1,30 +1,137 @@
...
+  const users = fixture.data as UserDetails[];
 
   return {
-    email: user.email,
+    email: users[0].email,
     password: testPassword,
   };
 }
+
+export function getAdminUserFixture(): TestUser {
```
olivier-nicodemi, line 109: `getSuperAdminUser` maybe? its not really a fixture (despite being created as one).

playwright/tests/pageComponents/PageBanner.ts
```diff
@@ -0,0 +1,13 @@
+// checks text in the flash message banner
+import { expect, Page } from "@playwright/test";
+
+export default class PageBanner {
+  private selector: string = "div.moj-banner.moj-banner--success";
```
olivier-nicodemi, line 5: Here and all the other js classes: consider using https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes/Private_elements instead of the ts `private` keyword as the latter does not enforce any runtime behaviour

playwright/tests/test-attaching-further-documents.spec.ts
```diff
@@ -0,0 +1,89 @@
...
+    );
+
+    // attach two files
+    const uploadedFiles = [];
+    const filesToUpload = ["testimage1.png", "testimage2.png"];
+    for (const fileToUpload of filesToUpload) {
+      await documentsUploadPage.goto();
+
+      const fileToUploadFullPath = path.join(
+        __dirname,
```
olivier-nicodemi, line 41: Check if `import.meta.dirname` is available else `dirname(fileURLToPath(import.meta.url))` should do the trick. (We want to avoid env globals except import.meta in es6)

olivier-nicodemi (changes requested):
A good chunk of the comments in FixtureService are only performance related and don't need addressing necessarily.

I am definitely not happy about the change in UserType. (What is a PaNamedUser with deputy type Lay?) If you want to store Deputy plus User type in one enum on the ts side that is fine but you need to map it in the controller then.

On the ts side I generally would prefer interpolation using ``` syntax rather than with `+` but that is minor. otherwise see my comments in the ts files. Note that I flagged recurring "issues" only once not each time they are in the code.

PR #2782, opened 2026-07-23
DDLS-1569 "Edit" forms contain "Add another" option

iqpalm:
## Purpose

Fixes DDLS-1569

## Approach

## Learning

## Checklist

### Frontend


client/app/src/Controller/Report/AssetController.php
```diff
@@ -211,14 +216,7 @@ public function propertyStepAction(Request $request, int $reportId, ?int $assetI
                         $request->getSession()->getFlashBag()->add('notice', 'Asset edited');
                     }
                 }
-
```
townxelliot, line 214: If you remove this, does this prevent "Add another" being shown for the add case? The switch statement also ignores the null case: is this how it copes with "Add another" being absent? We may want to deal explicitly with the null case here, or default to "no" if addAnother is unset (getStringOrDefault exists, I think).

client/app/src/Form/Report/Asset/AssetTypeProperty.php
```diff
@@ -68,8 +67,7 @@ public function buildForm(FormBuilderInterface $builder, array $options)
                 'grouping' => true,
                 'scale' => 2,
                 'invalid_message' => 'asset.property.rentIncomeMonth.type',
-            ])
-            ->add('addAnother', AddAnotherThingType::class);
```
townxelliot, line 72: Do we never show the "Add another" on this page?

client/app/src/Form/Report/ClientBenefitsCheckType.php
```diff
@@ -72,22 +72,22 @@ public function buildForm(FormBuilderInterface $builder, array $options): void
...
                 'entry_type' => MoneyReceivedOnClientsBehalfType::class,
                 'entry_options' => ['label' => false, 'empty_data' => null],
-                'delete_empty' => function (MoneyReceivedOnClientsBehalfInterface $money) use ($options) {
+                'delete_empty' => function (MoneyReceivedOnClientsBehalf $money) use ($options) {
                     return $money->getAmount() === null && $money->getMoneyType() === null && $money->getAmountDontKnow() === false && $options['allow_delete_empty'];
                 },
                 'allow_delete' => true,
             ]);
-
-            $builder->add('addAnother', SubmitType::class);
```
townxelliot, line 81: I think we may want to add this in some circumstances.

townxelliot (dismissed):
This looks OK, but my concern is that the "Add another" is just being removed from some of these pages, rather than being displayed conditionally. I think in some of the cases we should be checking whether we need the "Add another" choices, rather than just removing it.

client/app/src/Controller/Report/AssetController.php
iqpalm: No this is only deleted for edit case i.e where there is an asset id present. The add another is still present for add case on line 226. The add another option can only be yes or no for the add case else it returns an error for it requiring to be filled in

client/app/src/Form/Report/Asset/AssetTypeProperty.php
iqpalm: This is deleted from the AssetTypeProperty class and is separately added to the form for the add case on line 202 (if there is no asset id) in AssetController. If there is an asset id then it is an edit case and therefore no add another option is included on the form

client/app/src/Form/Report/ClientBenefitsCheckType.php
iqpalm: This is added in ClientsBenefitCheckController where it checks for moneyTypeId equalling null, meaning a new money received item is being added rather than edited so add another option is included on the form.

townxelliot (approved):

PR #2800, opened 2026-07-31
DDLS-1543 Rare redis timeout error

iqpalm:
## Purpose

Fixes DDLS-1543

## Approach
Retry approach

## Learning

## Checklist

### Frontend

### DIFF
--- a/api/app/src/Service/Client/TokenStorage/RedisStorage.php
+++ b/api/app/src/Service/Client/TokenStorage/RedisStorage.php
@@ -54,7 +54,7 @@ class RedisStorage extends TokenStorage
                     throw new \RuntimeException("Operation failed after {$maxRetries} retries.", 0, $e);
                 }
 
-                usleep(1000000); // Delay in microseconds
+                usleep(500000); // Delay in microseconds
             }
         }
     }
--- a/api/app/src/Service/Client/TokenStorage/RedisStorage.php
+++ b/api/app/src/Service/Client/TokenStorage/RedisStorage.php
@@ -22,9 +22,13 @@ class RedisStorage extends TokenStorage
     ) {
     }
 
-    public function get($id): mixed
+    public function get($id): ?string
     {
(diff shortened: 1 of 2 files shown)


api/app/src/Service/Client/TokenStorage/RedisStorage.php
```diff
@@ -23,17 +24,38 @@ public function __construct(
...
+            try {
+                return $operation();
+            } catch (ConnectionException $e) {
+                $attempt++;
+
+                if ($attempt >= $maxRetries) {
+                    throw new \RuntimeException("Operation failed after {$maxRetries} retries.", 0, $e);
+                }
+
+                usleep($retryDelay * 1000); // Delay in milliseconds
```
olivier-nicodemi: This is a bit confusing. I suggest you rename `$retryDelay` to `$retryDelayInMilliseconds` and if deemed necessary state that `usleep` takes microseconds to explain the times thousand.

olivier-nicodemi: On the other hand there is limited benefit of making this configurable as it is a private method in a small class so the cost of having to explain the parameters might outweigh future benefit of a generic method interface.

olivier-nicodemi (dismissed):
Please check delay amount with webobs before merge

api/app/src/Service/Client/TokenStorage/RedisStorage.php
```diff
@@ -21,19 +22,41 @@ public function __construct(
     ) {
     }
 
-    public function get($id): ?string
+    public function get($id): mixed
```
olivier-nicodemi: If the return wasn't a `?string` then it would have crashed in the past. So you can just make that explicit.
```
$value = ...
if ($value !== null && !is_string($value) {
   throw new ...
}
return $value;
```

olivier-nicodemi (approved):

PR #2805, opened 2026-08-05
Resolve Database Connection Issues

MurlockCadi:
Currently there is an issue in connecting to read only database user, as well as in some exotic envs its not able to find the password or the host to connect.

The script needs editing as well as some permissions and how the read only user is created. This PR should resole that.


jamesrwarren (approved):

api/docker/app/Dockerfile
```diff
@@ -164,6 +165,7 @@ CMD ["sh", "-c", "/var/www/check_processes.sh & \
     /var/www/wait-for-db.sh -h $DATABASE_HOSTNAME -p $DATABASE_PORT -t $TIMEOUT \
     && php app/console doctrine:migrations:migrate --allow-no-migration --no-interaction \
     && php app/console doctrine:migrations:up-to-date \
+    && ./scripts/setup_readonly_sql_query.sh \
```
jamesrwarren: I would try this on env where the readonly user is already created. I'm also wondering if we should just skip production by changing the .sh script to exit if on prod as we this seems like something where we could end up making a mistake in the future and we don't need it to run that sql every time the api container spins up on prod.

jamesrwarren (changes requested):

jamesrwarren (approved):

PR #2877, opened 2026-09-03
Ddls 1670f ecr vpc endpoint policy

jamesrwarren:
ECR endpoint policy.

Assumptions made: 

We don't need any write actions to ECR from within the VPC as that could be used as an attack vector.


olivier-nicodemi (approved):

terraform/account/region/variables.tf
```diff
@@ -18,9 +18,10 @@ data "aws_region" "current" {}
 data "aws_caller_identity" "current" {}
 
 locals {
-  current_main_region = data.aws_region.current.name
-  s3_bucket           = var.account.name
-  default_allow_list  = concat(module.allow_list.palo_alto_prisma_access, module.allow_list.moj_sites)
+  current_main_region   = data.aws_region.current.name
+  s3_bucket             = var.account.name
+  default_allow_list    = concat(module.allow_list.palo_alto_prisma_access, module.allow_list.moj_sites)
+  management_account_id = 311462405659
```
olivier-nicodemi, line 24: I assume this is not secret?

jamesrwarren: No the accounts are not considered secrets. There's an argument that they should be but all our OPG repos have the account identifiers in them and we decided as a group that they don't constitute a secret value.

PR #2878, opened 2026-09-04
DDLS-1427 PHPUnit - Upgrade to latest version in Digideps (Back end)

iqpalm:
## Purpose

Fixes DDLS-1427

## Approach

## Learning

## Checklist

### Frontend


api/app/tests/Integration/Command/IngestDeputyshipsCSVCommandTest.php
```diff
@@ -75,9 +81,12 @@ public function testExecuteWithFailedFilePullS3Error(): void
 
     public function testExecuteWithFailedCSVProcessing(): void
     {
-        $this->s3->expects($this->once())
-            ->method('getObject')
-            ->willReturn(new Result());
+        $this->s3->method('__call')
```
townxelliot, line 84: Does this cause an issue because there is no real getObject method on the S3Client?

api/app/tests/Integration/Controller/AuthControllerTest.php
```diff
@@ -209,8 +210,10 @@ private function resetAttempts($key): void
         ])['data'];
         $this->assertEquals('deputy@example.org', $data['email']);
 
+        /** @var Response $response */
+        $response = self::$frameworkBundleClient->getResponse();
```
townxelliot, line 214: I'm surprised this needs an `@var` definition before it. Is that necessary or does `getResponse` always return a Response (in which case `@var` isn't needed).

townxelliot (dismissed):
lgtm, just a couple of questions to clarify

townxelliot:
Co-pilot made a good point: https://github.com/ministryofjustice/opg-digideps/pull/2913/changes#r3987883396

I forgot we have a S3ClientMock, specifically because the real one is difficult to mock due to its magic methods.

api/app/tests/Integration/Command/IngestDeputyshipsCSVCommandTest.php
iqpalm: Yes it gives a PHPUnit\Framework\MockObject\MethodCannotBeConfiguredException: Trying to configure method "getObject" which cannot be configured because it does not exist, has not been specified, is final, or is static

api/app/tests/Integration/Controller/AuthControllerTest.php
iqpalm: No longer required must have needed it for one of the initial upgrades in PHPUnit. I have taken this out

olivier-nicodemi (approved):

PR #2907, opened 2026-09-10
DDLS-1686

olivier-nicodemi:
Fixes DDLS-1686


client/app/src/Form/Admin/ManageActiveReportType.php
```diff
@@ -35,15 +36,11 @@ public function buildForm(FormBuilderInterface $builder, array $options): void
                         'message' => 'report.endDate.beforeStart',
                         'groups' => ['startEndDates'],
                     ]),
-                    new Constraints\Expression([
-                        'expression' => 'this.getParent().get("startDate").getData().modify("+15 months") > this.getData()',
+                    new AugmentedExpression([
```
townxelliot, line 39: We do have OPG\Digideps\Frontend\Validator\Constraints\EndDateNotGreaterThanFifteenMonthsValidator which I think does something similar already.

olivier-nicodemi: The reason why this shenanigans were introduced is that we need to access different form fields at once during validation. Rewriting the form to have an object containing both start and end is outside of this tickets scope (and kind of was out of expected of the ticket introducing this validation).

townxelliot (approved):

PR #2917, opened 2026-09-14
DDLS-1654 add workflow changes requested by zizmor

jamesrwarren:
This PR is doing a few cleanup things:
a) a script to help use data-access with s3 temporarily when applied with breakglass
b) fix an issue with renovate where it wasn't grouping python uv dependencies
c) fix the zizmor warnings to improve workflow security
d) make preproduction2 deploy the image that was already on preproduction2


renovate.json
```diff
@@ -94,18 +94,14 @@
     },
     {
       "groupName": "Python dependency updates",
-      "matchManagers": ["uv", "pip_requirements", "pip-compile"],
+      "matchManagers": ["pep621", "pip_requirements", "pip-compile"],
```
jamesrwarren, line 97: uv isn't a real manager... think I was guessing a bit there... pep621 handles uv files

scripts/github_actions_audit/check_actions.sh
```diff
@@ -4,14 +4,14 @@ echo """
...
+actionlint -ignore 'at "uses" is not following the format "owner/' /github/workflows/*.yml
 
 echo """
 
 ===== RUNNING ZIZMOR =====
 
 """
 
-zizmor --collect=all /github
+zizmor --collect=all /github/workflows/*.yml
```
jamesrwarren, line 15: this wasn't actually working as intended.

scripts/github_actions_audit/check_actions.sh
```diff
@@ -4,14 +4,14 @@ echo """
 
 """
 
-actionlint /github/workflows/*.yml
+actionlint -ignore 'at "uses" is not following the format "owner/' /github/workflows/*.yml
```
jamesrwarren, line 7: actionlint doesn't recognise some of the newer standards that zizmor points out. In this case the documentation agress with zizmor

olivier-nicodemi (approved):

PR #2923, opened 2026-09-15
DDLS-1659 - Update previous report definition

Raffers:
## Purpose

Fixes DDLS-####

## Approach

## Learning

## Checklist

### Frontend


api/app/tests/Behat/features-v2/report-checklist/checklist.lay.feature
```diff
@@ -4,7 +4,6 @@ Feature: Reporting Checklists - Lay reporting checklist
     @admin @lay-combined-high-submitted
     Scenario: An admin submits the checklist form - applies to all admin roles
         Given a Lay Deputy has submitted a Combined High Assets report
-        And the deputies 'previous' report ends and is due 'more' than 60 days after the client benefits check feature flag date
```
Raffers, line 7: This has nothing associated with it, while running the tests thought best to remove.

api/app/src/Entity/Report/Report.php
```diff
@@ -969,25 +969,49 @@ public function setReviewChecklist(ReviewChecklist $reviewChecklist): void
     #[JMS\Type('array')]
     public function getPreviousReportData(): array
     {
+        $uidPopulateCallback = function (CourtOrder $courtOrder): string {
```
townxelliot: This can be inlined into array_map.

api/app/src/Entity/Report/Report.php
```diff
@@ -969,25 +969,49 @@ public function setReviewChecklist(ReviewChecklist $reviewChecklist): void
...
+            return $courtOrder->getCourtOrderUid();
+        };
+
+        $latestCourtOrderUids = array_map($uidPopulateCallback, $this->getActiveCourtOrders());
```
townxelliot:
```suggestion
        $latestCourtOrderUids = array_map(fn (CourtOrder $courtOrder) => $courtOrder->getCourtOrderUid(), $this->getActiveCourtOrders());
```

api/app/src/Entity/Report/Report.php
```diff
@@ -969,25 +969,49 @@ public function setReviewChecklist(ReviewChecklist $reviewChecklist): void
...
+
         $orderedSubmittedClientReports = $this->getClient()->getSubmittedReports();
+        $latestStartDate = $this->getStartDate();
+
+        $filteredReports = $orderedSubmittedClientReports->filter(function (Report $clientReport) use (
+            $latestStartDate,
+            $latestCourtOrderUids,
+            $uidPopulateCallback,
+        ): bool {
+            $courtOrderUids = array_map($uidPopulateCallback, $clientReport->getActiveCourtOrders());
```
townxelliot: I think this is the same as `$latestCourtOrderUids`, which you already derived above.

api/app/tests/Behat/features-v2/report-checklist/checklist.lay.feature
townxelliot: The step is sort of defined but incorrectly. I have fixed it in my typing branch. I would leave it in for now.

api/app/tests/Unit/Entity/Report/ReportTest.php
```diff
@@ -405,24 +405,24 @@ public function testGetPreviousReportData(): void
         );
     }
 
-    public function testNoPreviousReportDataReturnedOutside15Months(): void
+    public function testPreviousReportDataLogic(): void
```
townxelliot: I'm not sure why you changed the names of the variables as I found them more intuitive with the previous names.

api/app/src/Entity/Report/Report.php
```diff
@@ -969,25 +969,49 @@ public function setReviewChecklist(ReviewChecklist $reviewChecklist): void
...
-
-            $report = $filteredReports->first() ?: null;
-            if ($report !== null) {
-                return [
-                    'report-summary' => $report->getReportSummary(),
-                    'financial-summary' => $report->getFinancialSummary(),
-                    'assets-summary' => $report->getAssetsSummary(),
-                    'debts-summary' => $report->getDebtsSummary(),
-                ];
+            return count(array_diff($latestCourtOrderUids, $courtOrderUids)) === 0 &&
```
townxelliot: Isn't this just returning the report with the latest start date, whose end date is one day after its start date's day? 

Why are we filtering out the HW reports? We might only have HW reports for a court order.

This also treats all previous reports from all court orders as the potential "previous report". I think this might be faulty logic, as it could potentially find a PFA report as the previous report for an HW report.

olivier-nicodemi (dismissed):

api/app/src/Entity/Report/Report.php
```diff
@@ -969,25 +969,49 @@ public function setReviewChecklist(ReviewChecklist $reviewChecklist): void
...
+                $endDate->diff($latestStartDate)->days === 1 &&
+                $clientReport->isPfa();
```
townxelliot, line 994: What if we only have HW reports for a client?

api/app/src/Entity/Report/Report.php
```diff
@@ -969,25 +969,49 @@ public function setReviewChecklist(ReviewChecklist $reviewChecklist): void
...
         $orderedSubmittedClientReports = $this->getClient()->getSubmittedReports();
+        $latestStartDate = $this->getStartDate();
+
+        $filteredReports = $orderedSubmittedClientReports->filter(function (Report $clientReport) use (
+            $latestStartDate,
+            $latestCourtOrderUids,
+            $uidPopulateCallback,
+        ): bool {
+            $courtOrderUids = array_map($uidPopulateCallback, $clientReport->getActiveCourtOrders());
+            sort($courtOrderUids);
```
townxelliot, line 988: Does the order of the UIDs matter?

api/app/src/Entity/Report/Report.php
townxelliot, line 992: Maybe this is why they need to be sorted (for the array_diff)?

api/app/tests/Unit/Entity/Report/ReportTest.php
Raffers: I'll partially revert this, as even though I've left the dates as is. We do not care about the 15month part in the logic anymore.

api/app/src/Entity/Report/Report.php
Raffers: Good question, as we had this in place previously. I had assumed the data we cared about at this point was only relating to finance. I'll double check with Alex that's the case.

Raffers: From what I've just been told. He does not care if it's only pfa or both, but their is currently no information that they would need to compare in a previous hw report. This is also very unlikely to change in the future.

olivier-nicodemi (approved):

PR #2934, opened 2026-09-22
DDLS-1696 - OPG106 - "Deputy fees and expenses" won't save changes

Raffers:
## Purpose
A PA user has reported that they are unable to amend the “Have you charged the client any fees for your services during the reporting period?” question in their report once they have answered it “No”. 

Fixes DDLS-1696

## Approach

## Learning

## Checklist

### Frontend


api/app/src/Controller/Report/ReportController.php
```diff
@@ -284,12 +284,12 @@ public function update(Request $request, int $id): array
             ]);
         }
 
-        /** @var ?string $reasonForNoFees */
-        $reasonForNoFees = $data['reason_for_no_fees'] ?? null;
+        if (array_key_exists('reason_for_no_fees', $data)) {
```
townxelliot, line 287: If the `reason_for_no_fees` key is missing, this will throw an exception, which is why we have the `?? null` in the original code. So we can keep the original and remove the `array_key_exists()` call.

townxelliot (approved):

api/app/src/Controller/Report/ReportController.php
Raffers: But if we keep `?? null` & remove the `if` to check if it's either a null or the entry exists, then wouldn't this affect other updates & clear this field on the report?

townxelliot: You've got a checked for whether it's null on line 292, so that should prevent that from happening, shouldn't it?

Raffers: That check is for the fee's array. We need a check in place prior to this on the `$data` array to make sure we don't accidentally remove the reason for no fees. As we don't actually store a yes/no for the question `Have you charged the client any fees for your services during the reporting period?`. 

We set this question based on the value of `'reason_for_no_fees'` so to correctly display we can't remove this check, Alex has already raised this though, & will make another ticket if users raise we aren't persisting this field when switching between yes/no.

